PR #8, opened 2021-10-06
Simplify Makefile

sthaha:
* Removes need for modifying env vars GOBIN and PATH for all targets.
* Adds `tools` target that install all tools only if they aren't
  installed (faster installs)
* Adds `operator` target that builds operator
* Running `make` builds the operator by default

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,41 +1,61 @@
 SHELL=/usr/bin/env bash -o pipefail
 
 IMAGE = monitoring-stack-operator
-VERSION?=$(shell cat VERSION)
+VERSION ?= $(shell cat VERSION)
 
-TOOLS_BIN_DIR=$(shell pwd)/tmp/bin
-export PATH := $(TOOLS_BIN_DIR):$(PATH)
-export GOBIN := $(TOOLS_BIN_DIR)
+# running make builds the operator (default target)
+all: operator
 
-TOOLS=\
-	$(TOOLS_BIN_DIR)/controller-gen \
-	$(TOOLS_BIN_DIR)/golangci-lint
 
-$(TOOLS_BIN_DIR):
-	mkdir -p $(TOOLS_BIN_DIR)
+## Tools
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,42 +1,62 @@
...
-generate-deepcopy: $(TOOLS)
-	controller-gen object paths=./pkg/apis/...
+generate-deepcopy: $(CONTROLLER_GEN)
+	$(CONTROLLER_GEN) object paths=./pkg/apis/...
 
 .PHONY: generate
 generate: generate-crds generate-deepcopy
 
+operator: generate
+	go build -o operator ./cmd/operator/...
```
fpetkovski, line 58: Could we build this in either the temporary bin directory or add the file to gitignore?

Makefile
```diff
@@ -1,42 +1,62 @@
...
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1
+TOOLS_DIR = $(shell pwd)/tmp/bin
+CONTROLLER_GEN=$(TOOLS_DIR)/controller-gen
+GOLANGCI_LINT=$(TOOLS_DIR)/golangci-lint
+
+$(TOOLS_DIR):
+	@mkdir -p $(TOOLS_DIR)
+
+$(CONTROLLER_GEN): $(TOOLS_DIR)
+	@{ \
```
fpetkovski, line 20: What does this syntax do?

fpetkovski (commented):
Looks good, just have a small suggestion

Makefile
sthaha: lets use the `tmp/` dir :)

Makefile
sthaha: executes all commands in a single shell "process" so that
```
{
  FOO=bar
  echo $FOO
}
```
prints `bar` instead of "" if it wasn't in the same "shell"

PR #10, opened 2021-10-06
Deploy a prometheus-operator instance

fpetkovski:
This commit adds a new operator controller responsible for deploying an
instance of the prometheus-operator. Deploying the CRDs for the operator is optional
and is controlled by a command line flag. This allows the operator to
run seamlessly on OCP without interfering with the cluster monitoring
operator.
    
The controller does not create any watches on prometheus-operator
resources. It only runs periodic reconciliation and revert changes made
to these resources.


fpetkovski:
@simonpasquier @dgrisonnet ptal how we can use controller-runtime with server-side apply to simplify the management of static resources.

sthaha (approved):
/lgtm

PR #14, opened 2021-10-08
Support OLM based release flow

sthaha:
This add all necessary makefile targets to ship the operator through
OLM. To build and publish the following targets are added

- image: builds operator image
- generate: runs controller-gen to generate CRD, DeepCopy()
- image-push: publishes operator IMAGE to registry
- bundle: creates an OLM bundle that is installable through OLM
- bundle-image: creates an image of the bundle which can then be added
  to an OLM Catalog Image
- bundle-push: pushed the bundle image to registry
- catalog-image: creates an OLM Catalog image which serves the bundles
- catalog-push: publishes catalog image to registry


.github/workflows/olm.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Build Bundle Image
+        run: make bundle-image
+
+      - name: Publish Bundle
+        run: make bundle-push
+
+
+      - name: Build Catalog Image
+        run: |
+          env | grep CAT
```
fpetkovski, line 83: Debug statement?

Makefile
```diff
@@ -1,17 +1,23 @@
 SHELL=/usr/bin/env bash -o pipefail
 
-IMAGE = monitoring-stack-operator
+# IMAGE_TAG_BASE defines the registry/namespace and part of the image name
+# This variable is used to construct full image tags for bundle and catalog images.
+#
+IMAGE_TAG_BASE ?= quay.io/sthaha/monitoring-stack-operator
```
fpetkovski, line 6:
```suggestion
IMAGE_NAME ?= quay.io/sthaha/monitoring-stack-operator
```

Makefile
```diff
@@ -28,10 +34,33 @@ $(GOLANGCI_LINT): $(TOOLS_DIR)
 		GOBIN=$(TOOLS_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1 ;\
 	}
 
+$(KUSTOMIZE): $(TOOLS_DIR)
+	@{ \
+		set -e ;\
+		GOBIN=$(TOOLS_DIR) go get sigs.k8s.io/kustomize/kustomize/v3 ;\
```
fpetkovski, line 40: We should be using `go install`, `go get` will be deprecated for compiling binaries from source [1]. Can kustomize be installed with `go install`?

[1] https://golang.org/doc/go-get-install-deprecation

deploy/operator/rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: monitoring-stack-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: reconcile-monitoring-stacks
```
fpetkovski, line 9: Is this the correct name?

deploy/operator/serviceaccount.yaml
```diff
@@ -1,5 +1,4 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: monitoring-stack-operator
-
+  name: monitoring-stack-sa
```
fpetkovski, line 4: Is this SA needed only for deploying without OLM?

.github/workflows/olm.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Registry Login
+        # You may pin to the exact commit or the version. uses: docker/login-action@v1.10.0
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_LOGIN }}
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Build Operator Image
+        run: make image
```
fpetkovski, line 66: Should `make image` depend on `make operator`?

sthaha: `make image` should build operator in the context of a container, and shouldn't require operator to be built locally but ... we should definitely add `generate`!

.github/workflows/olm.yaml
sthaha: indeed, and I would like to leave that on until we have a stable release. Not knowing what these vars are did make the debugging of an CI failure hard

.github/workflows/olm.yaml
fpetkovski: Hm, why do we run `make operator` above then?

PR #15, opened 2021-10-08
Fix olm workflow

sthaha:


sthaha:
@fpetkovski , I noticed that the SOURCE_REF is not the name of the branch (`main`) instead it is `<pr>/merge` so the logic in workflow need to be fixed. Please note this is independent of the registry login failure

PR #16, opened 2021-10-08
Improve the local deployment experience

fpetkovski:
This commit allows the operator to be deployed to a local kubernetes cluster
using kubectl.

The operator can now be deployed with
```
kubectl apply -k deploy/crds -k deploy/operator
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,3 +179,6 @@ catalog-image: $(OPM) ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	docker push $(CATALOG_IMG)
+
+.PHONY: release
+release: operator-image operator-push bundle-image bundle-push catalog-image catalog-push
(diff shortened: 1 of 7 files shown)


deploy/operator/clusterrolebindings.yaml
```diff
@@ -10,16 +10,17 @@ roleRef:
...
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: monitoring-stack-operator-prometheus-operator
+  name: monitoring-stack-operator-prometheus-operator-delegated
 subjects:
   - kind: ServiceAccount
     name: monitoring-stack-operator-sa
-
+    namespace: default
```
sthaha, line 26: IIUC, this may cause issues when bundling.. have you tried installing the bundle?

fpetkovski: The change does not seem to interfere with OLM, the bundle is still installable. I think that the rolebinding is actually ignored when the bundle is compiled.

Makefile
```diff
@@ -179,3 +179,6 @@ catalog-image: $(OPM) ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	docker push $(CATALOG_IMG)
+
+.PHONY: release
```
sthaha, line 183: ❤️

sthaha (approved):
/lgtm

PR #18, opened 2021-10-11
Add an e2e suite

fpetkovski:
This commit adds an e2e test suite which can be run from the local
machine. It also adds a test which verifies that the prometheus-operator
instance which is deployed only reconciles resources with the
app.kubernetes.io/managed-by=monitoring-stack-operator label.


.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,44 @@
...
+    name: E2E tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.golang-version }}
+      - name: Build images
+        run: |
+          export SHELL=/bin/bash
```
sthaha, line 26: why do we need this? doesn't the `makefile` already sets the shell?

go.mod
```diff
@@ -3,11 +3,17 @@ module rhobs/monitoring-stack-operator
...
+	k8s.io/api v0.21.2
+	k8s.io/apiextensions-apiserver v0.21.2
 	k8s.io/apimachinery v0.21.2
-	k8s.io/client-go v0.21.2
+	k8s.io/client-go v12.0.0+incompatible
 	sigs.k8s.io/controller-runtime v0.9.2
 	sigs.k8s.io/kustomize/kustomize/v3 v3.10.0 // indirect
 )
+
+// A replace directive is needed for k8s.io/client-go because Cortex (which
```
sthaha, line 17: Thank you for writing this!

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.golang-version }}
+      - name: Build images
+        run: |
+          export IMAGE_TAG_BASE=monitoring-stack-operator
+          make operator-image
+      - name: Start KinD
```
sthaha, line 28: May in a separate pr. I would be nice to have this as a `hack/kind-setup.sh` script.

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,42 @@
+name: e2e
+on:
+  pull_request:
```
sthaha, line 3: shouldn't we run this before the PR gets merged? 
It may be worth taking advantage of the `build` workflow to run the same in parallel

sthaha (commented):
I have tested that this works fine locally.

.github/workflows/e2e-tests.yaml
sthaha: I `kind` of agree with you but to have a default imho is fine, otherwise we could document it but why document code :D

PR #19, opened 2021-10-11
ci: lint commits according to the Angular convention

fpetkovski:
This commit adds a commit message linter in the CI and documents the
convention that should be used when writing commit messages.

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -3,14 +3,18 @@ on:
   pull_request:
   push:
     branches:
-      - 'release-*'
-      - 'master'
-      - 'main'
-    tags:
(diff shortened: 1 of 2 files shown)


.github/workflows/static-checks.yaml
```diff
@@ -3,14 +3,18 @@ on:
   pull_request:
```
sthaha, line 3: shouldn't we enable it for the `pull_request` instead of `push` ?

fpetkovski: Do you mean only for pull_request? Because now it's enabled for both pull_request and push. 

Also this workflow is for all static checks, including go and manifest linting, so whatever we add here will apply to all static checks.

PR #21, opened 2021-10-11
ci: implement an automated Github release process

fpetkovski:
This commit adds a github workflow which is responsible for creating a
new Github release. The workflow creates:
* a new changelog entry for the new release
* a new git tag
* a new github release marked as a pre-release

The workflow can be triggered with a commit message that has a `chore(release):` prefix.

### DIFF
--- a/.versionrc
+++ b/.versionrc
@@ -1,7 +1,7 @@
 {
-  "commitUrlFormat": "{{host}}/rhobs/monitoring-stack-operator/commit/{{hash}}",
-  "compareUrlFormat": "{{host}}/rhobs/monitoring-stack-operator/commit/{{hash}}",
-  "issueUrlFormat": "{{host}}/rhobs/monitoring-stack-operator/issues/{{id}}",
+  "commitUrlFormat": "https://github.com/rhobs/monitoring-stack-operator/commit/{{hash}}",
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,54 @@
...
+    if: "contains(github.event.head_commit.message, 'chore(release):')"
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: version
+        name: Determine release version
+        run: |
+          version=$(npx standard-version --dry-run | grep 'tagging release' | rev | cut -d ' ' -f 1 | rev | cut -c2-)
```
sthaha, line 21: what installs this package? `npx standard-version` wouldn't it wait for user input if it is not installed already?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,54 @@
+---
+name: Github Release
```
sthaha, line 2: How do we make a release with breaking change? Do we manually update the VERSION 's major version?

fpetkovski: We can either use `fix!` or `feature!` to indicate a breaking change, or we can force the version with `make initiate-release-as RELEASE_VERSION=<version>`. The former is documented in conventional commits, and the latter is added to the release documentation in the readme.

PR #22, opened 2021-10-12
olm: make use of official rhobs org

sthaha:
This patch modifies the Makefile and the olm deployment workflow to make
use of official quay.io/rhobs org for publishing the operator, bundle
and the catalog images.

The PR also fixes an issue introduced in commit 96f52217928 which
installed an old version of `kustomize` that did not have the
`set label` command.

The Makefile is modified to use local image, i.e. no `quay.io/<org>` so
that local development does not require you to override the
`IMAGE_TAG_BASE` variable


sthaha:
@fpetkovski we need to update the secrets before this patch gets merged.

Makefile
```diff
@@ -2,8 +2,7 @@ SHELL=/usr/bin/env bash -o pipefail
 
 # IMAGE_TAG_BASE defines the registry/namespace and part of the image name
 # This variable is used to construct full image tags for bundle and catalog images.
-#
-IMAGE_TAG_BASE ?= quay.io/sthaha/monitoring-stack-operator
+IMAGE_TAG_BASE ?= monitoring-stack-operator
```
fpetkovski, line 5: Isn't this `IMAGE_BASE` since there are no tags in the value?

sthaha: I am happy to change it :-)

PR #24, opened 2021-10-12
DRAFT: ci: use image variables from the repository environment variables

fpetkovski:


.github/workflows/olm.yaml
```diff
@@ -32,8 +32,6 @@ jobs:
             CHANNELS=development,stable
           fi
 
-          IMAGE_BASE=quay.io/rhobs/monitoring-stack-operator
```
sthaha, line 35: nice! I was thinking the same :D

PR #30, opened 2021-10-13
fix: parametrize the namespace of the prometheus operator

fpetkovski:
This commit allows running the prometheus operator in a namespace other
than default. While this was possible before, it was broken because the
service account namespace in the cluster role binding was always set as
default.

When installed through OLM, the namespace argument will be set as the current
namespace. This means that the prometheus operator will be deployed in
the same namespace as the monitoring stack operator.


test/e2e/main_test.go
```diff
@@ -9,25 +9,34 @@ import (
...
 var (
 	f *framework.Framework
 
-	// TODO(fpetkovski): change once we are able to deploy the operator to a different namespace
-	e2eTestNamespace = "default"
+	e2eTestNamespace = "e2e-tests"
 )
 
 func TestMain(m *testing.M) {
+	main(m)
```
fpetkovski, line 26: Deferred calls are not executed on `os.Exit` from TestMain. As a workaround, we call another function in which we can add deferred calls.

pkg/controllers/prometheus-operator/controller.go
```diff
@@ -80,8 +82,8 @@ func (r *reconciler) Reconcile(ctx context.Context, request ctrlruntime.Request)
 	return ctrlruntime.Result{}, nil
 }
 
-func (r *reconciler) loadStaticResources() ([]client.Object, error) {
-	resources := []assets.Asset{
+func (r *reconciler) loadStaticResources(namespace string) ([]client.Object, error) {
```
sthaha, line 85: nit: isn't the namespace already accessible through `r.opts` ?

test/e2e/main_test.go
```diff
@@ -9,25 +9,34 @@ import (
...
+	os.Exit(0)
+}
+
+func main(m *testing.M) {
 	setLogger()
 	op := createOperator()
 	setupFramework(op)
 
+	ns := createTestNamespace()
```
sthaha, line 35: *nit*: may be worth considering something like below so that the setup doesn't leak details. 

```
cleanup := setupTestNamespce()
defer cleanup()
```

test/e2e/main_test.go
```diff
@@ -61,3 +72,27 @@ func createOperator() *operator.Operator {
...
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: e2eTestNamespace,
+		},
+	}
+	if err := f.K8sClient.Create(context.Background(), ns); err != nil {
+		log.Println(err)
+		os.Exit(1)
```
sthaha, line 88: imho, it may be useful to just return error if and have a single place to exit - i.e. main so that TestMain reads soething like ... 

```go
func TestMain() {
  if err := main(t); err != nil {
    os.Exit(1)
  }
  os.Exit(0)
}
```

pkg/controllers/prometheus-operator/controller.go
```diff
@@ -111,8 +109,36 @@ func (r *reconciler) loadStaticResources() ([]client.Object, error) {
...
-		resources = append(crds, resources...)
+		staticAssets = append(crds, staticAssets...)
+	}
+
+	resources, err := r.assetLoader.Load(staticAssets)
+	if err != nil {
+		return nil, err
+	}
+
+	crb := &authorizationv1.ClusterRoleBinding{
```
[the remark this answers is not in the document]
simonpasquier: my 2 cents: static assets are fine as long as they are really static. Once you need to customize, I find it more natural to manage everything in code. CMO uses the approach of loading static manifests and then modify some of them but it's kinda confusing to me.

test/e2e/main_test.go
fpetkovski: Good idea

pkg/controllers/prometheus-operator/controller.go
fpetkovski: Correct, removed.

test/e2e/main_test.go
fpetkovski: Good idea, done

PR #33, opened 2021-10-18
docs: improve documentation for running the operator locally

fpetkovski:


README.md
```diff
@@ -33,8 +33,6 @@ For detailed information on the available code generation markers, please refer
 ### Running the operator in kind
 * Install [kind](https://github.com/kubernetes-sigs/kind)
 * Create a local kubernetes cluster with `kind create cluster`.
-* Build the operator image with `make operator-image`.
-* Load the image into your cluster with `kind load docker-image monitoring-stack-operator:$(cat VERSION)`
 * Apply the CRDs by running `kubectl apply -k deploy/crds/`
```
sthaha, line 36: I think this fails "partially" due to kustomization.yaml in that dir. `kustomize build deploy/crd | kubectl apply -f -` or we would have to remove that file like it was removed from `deploy/operator`

fpetkovski: We are using kubectl `-k` which can understand kustomize code. Does the command fail for you locally?

sthaha (approved):

PR #36, opened 2021-10-18
ci: fix generating development channel versions

fpetkovski:
The release pipeline for the main branch appends the current date to the
value present in VERSION to determine the new bundle version. Such
versioning prevents subscriptions to the development channel to upgrade
to development versions once they reach an RC or a stable version.

This commit bumps the version in the CI before a new development version
is released. This allows upgrades to progress to an RC/stable version to
a development version. It also serves as an indication that the code in
the main branch is going to end up in the next (and not the current)
version.


sthaha (approved):
great find 🙇

PR #42, opened 2021-10-20
test: allow running e2e tests without the operator

fpetkovski:
We currently spin up the operator in a goroutine when running e2e tests.
This works well for the local environment but is problematic in the CI
where we deploy the operator in a cluster before running the tests.

This commit adds a flag to exclude running the operator in a goroutine
during e2e tests. This way we can still maintain the same workflow
locally while disabling the goroutine in the CI.


sthaha (approved):
lgtm

simonpasquier (approved):
lgtm

PR #47, opened 2021-10-21
test: run some prometheus-operator tests in parallel

fpetkovski:
The prometheus operator tests sequentially run long-lasting assertions
that verify whether some resources are created or not. These assertions
are independent and can be checked concurrently.

This commit parallelizes those assertions using t.Parallel().


sthaha (approved):
lgtm

PR #48, opened 2021-10-22
fix: apply base CSV during bundle generation

fpetkovski:
The base CSV needs to have the package prefix in order to be taken
into account during bundle generation.


sthaha (approved):
This was a rename typo ..

PR #55, opened 2021-10-28
test: add additional grafana-operator tests

fpetkovski:
MSO maintains an OLM OperatorGroup and Subscription through which it
installs a Grafana Operator.

This commit adds a test to ensure that these resources are created.
It also adds assertions that the resources are restored if someone
deletes them.


sthaha (approved):
lgtm

PR #56, opened 2021-10-29
feat: deploy a default grafana instance

fpetkovski:
This commit deploys a default Grafana instance in the
monitoring-stack-operator namespace. The instance is configured to
collect all dashboards with the label: `app.kubernetes.io/part-of: monitoring-stack-operator`.

### DIFF
--- a/deploy/operator/monitoring-stack-operator-cluster-role.yaml
+++ b/deploy/operator/monitoring-stack-operator-cluster-role.yaml
@@ -119,6 +119,16 @@ rules:
   - update
   - watch
 - apiGroups:
+  - integreatly.org
+  resources:
+  - grafanas
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
   - operators.coreos.com
   resources:
   - operatorgroups
--- a/pkg/eventsource/ticker.go
+++ b/pkg/eventsource/ticker.go
@@ -21,13 +21,13 @@ type TickerSource struct {
 }
 
 // NewTickerSource creates a new TickerSource
-func NewTickerSource() *TickerSource {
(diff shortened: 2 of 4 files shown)


pkg/eventsource/ticker.go
```diff
@@ -27,7 +27,7 @@ func NewTickerSource() *TickerSource {
 		Channel: source.Channel{
 			Source: channel,
 		},
-		ticker:  time.NewTicker(30 * time.Second),
+		ticker:  time.NewTicker(30 * time.Minute),
```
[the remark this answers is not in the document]
sthaha: nit: may be worth passing in `time.Duration` as an argument so that each controller can set its reconciliation interval.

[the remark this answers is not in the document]
fpetkovski: done :+1:

deploy/operator/monitoring-stack-operator-cluster-role.yaml
```diff
@@ -118,6 +118,16 @@ rules:
   - list
   - update
   - watch
+- apiGroups:
+  - integreatly.org
+  resources:
+  - grafanas
+  verbs:
+  - create
+  - get
```
sthaha, line 127: wondering if we should remove `get` ?

sthaha (approved):

deploy/operator/monitoring-stack-operator-cluster-role.yaml
fpetkovski: Yes, good idea. I'll try it out.

fpetkovski: Let me merge this and I'll follow up with a PR to remove get

PR #58, opened 2021-10-29
feat: implement self-scraping for monitoring stacks

fpetkovski:
Currently the Prometheus instances provisioned by MonitoringStacks are
empty and have no metrics.

This commit creates the necessary ServiceMonitor and RBAC rules for
Prometheus instances to scrape their own metrics.


pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,343 @@
...
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     rbacResourceName,
+		},
+	}
+	return roleBinding
+}
+
+func newServiceMonitor(ms *stack.MonitoringStack) *monv1.ServiceMonitor {
```
fpetkovski, line 311: This does not work when a monitoring stack has label selectors because the service monitor will not get picked up. Should we use `additionalScrapeConfigs` in the `PrometheusSpec` to make sure that setting selectors will not interfere with self-scraping?

https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheusspec

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		StringData: map[string]string{
+			AdditionalScrapeConfigsSelfScrapeKey: `- job_name: prometheus-self
+  honor_labels: true
```
fpetkovski, line 335: I think it's better to define the self-scraping configuration like this so that we don't have to worry about the resource selectors defined for the monitoring stack. If we decide to go with a ServiceMonitor, the labels on that ServiceMonitor will need to be synthesised in a way to match what the user defines in the resourceSelector for the MonitorigStack.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	err := r.k8sClient.Get(ctx, key, existing)
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	createNew := errors.IsNotFound(err)
+
+	desired, err := patcher.patch(existing)
+	if err != nil {
+		return err
```
sthaha, line 163: how about we logger.Error(..) ?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	if ms.Namespace == desired.GetNamespace() {
```
sthaha, line 166: why is this check needed?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -0,0 +1,179 @@
...
+			return nil, nil
+		}
+		logger.Error(err, "failed to get monitoring stack")
+		return nil, err
+	}
+
+	return &ms, nil
+}
+
+func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringStack, patcher objectPatcher, logger logr.Logger) error {
```
sthaha, line 145: nit: doesn't `r.logger` suffice?

test/e2e/framework/framework.go
```diff
@@ -1,7 +1,50 @@
...
+	"k8s.io/client-go/transport/spdy"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
 
 type Framework struct {
+	Config    *rest.Config
 	K8sClient client.Client
 }
+
+// StartPortForward initiates a port forwarding connection to a pod on the localhost interface.
```
sthaha, line 21: nit .. edit error?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		// no such monitoring stack, so stop here
+		return ctrl.Result{}, nil
+	}
+
+	patchers, err := r.stackComponentPatchers(ms, r.instanceSelectorKey, r.instanceSelectorValue)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	for _, mapper := range patchers {
```
sthaha, line 119: nit: mapper -> `patcher`

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		if err := r.reconcileObject(ctx, ms, mapper, logger); err != nil {
```
sthaha, line 120: Just a thought ... will this do a partial reconciliation and leave system from a stable to unstable state?

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+			APIVersion: "monitoring.coreos.com/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ms.Name,
+			Namespace: ms.Namespace,
+			Labels:    prometheusLabels(ms.Name, instanceSelectorKey, instanceSelectorValue),
+		},
+
+		Spec: monv1.PrometheusSpec{
+			Image: nil,
```
sthaha, line 271: typo ?

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+				}
+			},
+			patch: func(object client.Object) (client.Object, error) {
+				role := newRole(ms, rbacResourceName, rbacVerbs)
+
+				if object == nil {
+					return role, nil
+				}
+
+				existing, ok := object.(*rbacv1.Role)
```
sthaha, line 74: because `existing` is an object from cache, is it okay to change it without a `DeepCopy()` ?

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+			patch: func(object client.Object) (client.Object, error) {
+				roleBinding := newRoleBinding(ms, rbacResourceName)
+
+				if object == nil {
+					return roleBinding, nil
+				}
+
+				existing, ok := object.(*rbacv1.RoleBinding)
+				if !ok {
+					return nil, NewErrInvalidObjectType("RoleBinding.rbac.authorization.k8s.io/v1")
```
sthaha, line 100: this can be simplified to avoid typo by passing the `empty` obj ... e.g
```
func NewErrInvalidObjectType(obj client.Object) error {
	gvk := obj.GetObjectKind().GroupVersionKind()
	gvkString := gvk.Group + "/" + gvk.Version + " " + gvk.Kind

	return ErrInvalidObjectType{expectedType: gvkString}
}
...

return nil, NewErrInvalidObjectType(roleBinding)

```
while at it, how about adding information about what object was actually passed and not just what was expected?

sthaha (dismissed):

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+type emptyObjectFunc func() client.Object
+
+type patchObjectFunc func(object client.Object) (client.Object, error)
+
+type objectPatcher struct {
+	empty emptyObjectFunc
+	patch patchObjectFunc
+}
+
+type ErrInvalidObjectType struct {
```
sthaha, line 29: how about something like ... 

```go
type TypeError struct {
	expectedType string
	actualType   string
}

func (e TypeError) Error() string {
	return fmt.Sprintf("object type %q is not %q", e.actualType, e.expectedType)
}

func gvk(o client.Object) string {
	gvk := o.GetObjectKind().GroupVersionKind()
	return gvk.Group + "/" + gvk.Version + " " + gvk.Kind
}

func TypeMismatchError(expected, actual client.Object) error {
	return TypeError{expectedType: gvk(expected), actualType: gvk(actual)}
}

```
This makes the return error easier like 

```
return nil, TypeMismatchError(role, obj)
return nil, TypeMismatchError(sa, obj)
return nil, TypeMismatchError(rb, obj)

```

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+
+func NewErrInvalidObjectType(expectedType string) error {
+	return ErrInvalidObjectType{expectedType: expectedType}
+}
+
+func (e ErrInvalidObjectType) Error() string {
+	return fmt.Sprintf("object type is not %s", e.expectedType)
+}
+
+func (r *reconciler) stackComponentPatchers(ms *stack.MonitoringStack, instanceSelectorKey string, instanceSelectorValue string) ([]objectPatcher, error) {
```
sthaha, line 41: this need not be a  method of `reconciler` and can be a pure function.

pkg/controllers/monitoring-stack/components.go
```diff
@@ -0,0 +1,403 @@
...
+				existing.Spec = dataSource.Spec
+				return existing, nil
+			},
+		},
+	}, nil
+}
+
+func newGrafanaDataSource(ms *stack.MonitoringStack) *grafanav1alpha1.GrafanaDataSource {
+	datasourceName := fmt.Sprintf("ms-%s-%s", ms.GetNamespace(), ms.GetName())
+	prometheusUrl := fmt.Sprintf("prometheus-operated.%s:9090", ms.GetNamespace())
```
sthaha, line 185: nit: `Url` -> `URL`

pkg/controllers/monitoring-stack/controller.go
fpetkovski: Cross-namespace owner references are disallowed in Kubernetes

pkg/controllers/monitoring-stack/controller.go
fpetkovski: Yes, I think it could. I am just not sure how to solve it.

sthaha (approved):

PR #61, opened 2021-11-04
perf: avoid calling deepcopy of existing object

sthaha:
DeepCopyObject() was used to avoid modifying the object in informer's
cache but this isn't the case as `client.Get(ctx, gvk, &obj)` ensures
that a new/empty object is passed which then gets initialised by the
cache-reader.

See also: https://github.com/kubernetes-sigs/controller-runtime/blob/1638a6a9b82dc1e0046c7a1006f12dacd9475f54/pkg/cache/internal/cache_reader.go#L81

### DIFF
--- a/pkg/controllers/monitoring-stack/components.go
+++ b/pkg/controllers/monitoring-stack/components.go
@@ -21,7 +21,7 @@ const AdditionalScrapeConfigsSelfScrapeKey = "self-scrape-config"
 
 type emptyObjectFunc func() client.Object
 
-type patchObjectFunc func(object client.Object) (client.Object, error)
+type patchObjectFunc func(existing client.Object) (client.Object, error)
 
 type objectPatcher struct {
 	empty emptyObjectFunc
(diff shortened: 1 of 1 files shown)


pkg/controllers/monitoring-stack/components.go
```diff
@@ -21,7 +21,7 @@ const AdditionalScrapeConfigsSelfScrapeKey = "self-scrape-config"
 
 type emptyObjectFunc func() client.Object
 
-type patchObjectFunc func(object client.Object) (client.Object, error)
+type patchObjectFunc func(existing client.Object) (client.Object, error)
```
sthaha, line 24: I was also thinking patch should just mutate the `existing` and return error if any .. so that the caller only needs to 
```
if err := pather.patch(existing); err != nil {
  return err
}
```

but it will require modification of `serviceAccount` patcher. 
Additionally, the patch will change to 
```go
patch: func (existing client.Object) error {
  if existing == nil {
    return errNilObject
  }
  actual, ok := existing.(v1.ProperType)
  if !ok {
    return errTypeMismatch
  }
 
 desired := newObjectForMs(ms)
  current.Spec = desired.Spec
  return nil
}
```

fpetkovski (dismissed):
This is more like a `perf: ` but we can tweak the changelog :)

fpetkovski (approved):

PR #69, opened 2021-11-11
feat: Deploy an Alertmanager instance for each stack

fpetkovski:
This commit adds an Alertmanager to each MonitoringStack and
configures the Prometheus instances to send alerts to that Alertmanager
instance.

### DIFF
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -2,12 +2,15 @@ package e2e
 
 import (
 	"context"
-	monitoringstack "rhobs/monitoring-stack-operator/pkg/controllers/monitoring-stack"
+	"encoding/json"
+	"fmt"
+	"net/http"
 	"rhobs/monitoring-stack-operator/test/e2e/framework"
 	"testing"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	stack "rhobs/monitoring-stack-operator/pkg/apis/v1alpha1"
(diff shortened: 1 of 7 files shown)


test/e2e/monitoring_stack_controller_test.go
```diff
@@ -245,6 +242,95 @@ func assertPrometheusScrapesItself(t *testing.T) {
...
+	}
+}
+
+func getAlertmanagerAlerts() ([]alert, error) {
+	client := http.Client{}
+	resp, err := client.Get("http://localhost:9093/api/v2/alerts")
+	if err != nil {
+		return nil, err
+	}
+
```
sthaha, line 292: defer resp.Body.Close()

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -60,35 +71,20 @@ func TestMonitoringStackController(t *testing.T) {
 
 func emptyStackCreatesPrometheus(t *testing.T) {
 	ms := newMonitoringStack("empty-stack")
+	defer f.K8sClient.Delete(context.Background(), ms)
```
sthaha, line 74: good find! 🙇 
How about `newFoobar(t, ..)`and each `newFoobar` calls `t.Cleanup(...)` ?

fpetkovski: Good question, I am not sure what's the best way to handle this. It's really a question of who should own the lifecycle management of an object: the `newX` function or the caller itself.

sthaha: For the tests at least, the default would be to cleanup resource it creates, wouldn't it? If we follow this pattern all resources created will be automatically deleted. What I haven't tested is if it would yield the same behaviour as it does now. I.E .. currently `empty-stack` will be deleted at the end of `emptyStackCreatesPrometheus` function whereas t.Cleanup may be deleting the resource at the end of the a test unless these methods are called within t.Run(...).

sthaha: btw, this doesn't block the patch from merging ... it was just a thought.

fpetkovski: Let's try out the `t.Cleanup` approach. It might help us keep tests independent by default.

sthaha:
@fpetkovski is this still a draft?

fpetkovski:
Should be ready for review now

sthaha (approved):

fpetkovski:
@sthaha looks like tests are passing now, could you ptal again?

sthaha (approved):

PR #71, opened 2021-11-16
ci: ensure operator has no reconcile errors

sthaha:
This patch makes use of promq tool to scape operator metrics to ensure
that it does not generate any reconcile errors.

### DIFF
--- a/.github/e2e-tests-olm/action.yaml
+++ b/.github/e2e-tests-olm/action.yaml
@@ -33,11 +33,13 @@ runs:
         wait: 300s
         config: ./hack/kind/config.yaml
 
+    - name: Install required tools
+      shell: bash
+      run: make operator-sdk promq
+
     - name: Install OLM
       shell: bash
-      run: |
-        make operator-sdk
-        ./tmp/bin/operator-sdk olm install
+      run: ./tmp/bin/operator-sdk olm install
 
     - name: Create a local docker registry in the Kind cluster
       shell: bash
@@ -48,8 +50,8 @@ runs:
         echo "127.0.0.1 local-registry" | sudo tee -a /etc/hosts
 
     - name: Build images
-      run: make operator-image bundle bundle-image IMAGE_BASE="local-registry:30000/monitoring-stack-operator" VERSION=0.0.0-ci
       shell: bash
+      run: make operator-image bundle bundle-image IMAGE_BASE="local-registry:30000/monitoring-stack-operator" VERSION=0.0.0-ci
 
     - name: Wait for cluster to finish bootstraping
       shell: bash
@@ -59,8 +61,8 @@ runs:
         kubectl get pods -A
 
     - name: Publish images
-      run: make operator-push bundle-push IMAGE_BASE="local-registry:30000/monitoring-stack-operator" VERSION=0.0.0-ci
       shell: bash
+      run: make operator-push bundle-push IMAGE_BASE="local-registry:30000/monitoring-stack-operator" VERSION=0.0.0-ci
 
     - name: Deploy the operator
       shell: bash
@@ -75,6 +77,73 @@ runs:
           --namespace operators \
           --skip-tls
 
+        kubectl rollout status deployment monitoring-stack-operator -n operators
+
+    - name: Expose operator metrics endpoint
+      shell: bash
+      run: |
+        kubectl apply -n operators -f hack/kind/operator-metrics-service.yaml
+        sleep 2
+
+        # wait for the service to create an endpoint
+        while ! kubectl get -n operators ep  operator-metrics-service && [[ "$i" -le 5 ]]; do
+            (( i++ ))
+            echo " - $i ... repeat "
+            sleep 2
+        done
+
+    - name: Assert no reconciliation errors before running the test
+      shell: bash
+      run: |
+        # assert there are no reconciliation errors before running test
+        # give operator some time to start before collecting the metrics
+        sleep 5
+        ./tmp/bin/promq -t http://localhost:30001/metrics  \
+          -q 'controller_runtime_reconcile_errors_total' -o yaml
+
+        reconcile_errors=$( ./tmp/bin/promq \
+          -t http://localhost:30001/metrics  -o yaml \
+          -q 'sum(controller_runtime_reconcile_errors_total)' \
+          | tr -d ' ' | grep ^v: | cut -f2  -d: )
+
+        echo "Reconcile errors before running tests: $reconcile_errors"
+
+        if [[ "$reconcile_errors" -ne 0 ]]; then
+          kubectl logs -n operators deploy/monitoring-stack-operator
+          echo ================================================================
+          echo "Expecting 0 reconcile errors but found $reconcile_errors"
+          exit 1
+        fi
+
     - name: Run tests
       shell: bash
-      run: go test -v ./test/e2e/...
+      run: |
+        go test -v ./test/e2e/... || {
+          echo ================================================================
+          # print operator logs on failure
+          kubectl logs -n operators deploy/monitoring-stack-operator
+          exit 1
+        }
+
+    - name: Print operator logs and reconcile stats
+      shell: bash
+      run: |
+        kubectl logs -n operators deploy/monitoring-stack-operator
+        echo ================================================================
+
+        ./tmp/bin/promq -t http://localhost:30001/metrics  \
+          -q 'controller_runtime_reconcile_errors_total'
+
+    - name: Assert that there are no reconciliation errors
+      shell: bash
+      run: |
+        reconcile_errors=$( ./tmp/bin/promq \
+          -t http://localhost:30001/metrics  -o yaml \
+          -q 'sum(controller_runtime_reconcile_errors_total)' \
+          | tr -d ' ' | grep ^v: | cut -f2  -d: )
+
+        echo "Reconcile errors after test run: $reconcile_errors"
+        if [[ "$reconcile_errors" -ne 0 ]]; then
+          echo "Found $reconcile_errors reconciliation errors"
+          exit 1
+        fi
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -44,9 +44,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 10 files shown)


.github/e2e-tests-olm/action.yaml
```diff
@@ -33,10 +33,14 @@ runs:
         wait: 300s
         config: ./hack/kind/config.yaml
 
+    - name: Install required tools
+      shell: bash
+      run: |
+        make operator-sdk promq
```
fpetkovski, line 39:
```suggestion
      run: make operator-sdk promq
```

.github/e2e-tests-olm/action.yaml
```diff
@@ -33,10 +33,14 @@ runs:
...
+
     - name: Install OLM
       shell: bash
       run: |
-        make operator-sdk
         ./tmp/bin/operator-sdk olm install
```
fpetkovski, line 44: We can consider inlining this command now that there's only one thing to execute

.github/e2e-tests-olm/action.yaml
```diff
@@ -75,6 +79,46 @@ runs:
           --namespace operators \
           --skip-tls
 
+        # NOTE: operator will deploy other pods as well which we don't want to wait for
+        kubectl wait --for=condition=Ready pods \
```
fpetkovski, line 83: This looks a bit fragile tbh. If we change the labels, it might be hard to find why e2e tests suddenly start to fail. I would suggest that we simply wait for all pods in the `operators` namespace

.github/e2e-tests-olm/action.yaml
```diff
@@ -75,6 +79,46 @@ runs:
...
       shell: bash
       run: go test -v ./test/e2e/...
+
+    - name: Print operator reconcile errors
+      shell: bash
+      run: |
+        ./tmp/bin/promq -t http://localhost:30001/metrics  \
+          -q 'controller_runtime_reconcile_errors_total'
+
+    - name: Ensure there are reconcile errors
```
fpetkovski, line 116: Should we make this assertion part of the query? E.g.
```
promq -t http://localhost:30001/metrics  -q 'controller_runtime_reconcile_errors_total == 0'
```
and then we just need to check that we get some results back.

.github/e2e-tests-olm/action.yaml
sthaha: I did think of all pods but then it will also wait for pods that operator (controllers in operator) starts.

sthaha: This should be solved in theory when your sig-instrumentation patch merges upstream and we should be able to use promq as dependency.

.github/e2e-tests-olm/action.yaml
sthaha: Good idea .. let me try :)

.github/e2e-tests-olm/action.yaml
fpetkovski: Can we not wait only for the operator to start? E.g.
```
kubectl rollout status -n operators deploy monitoring-stack-operator
```

sthaha: Yes, that is a way better idea!

.github/e2e-tests-olm/action.yaml
sthaha: I tried this but ... it gives us false negatives 

Given: 

```yaml
❯ ./tmp/bin/promq -t http://localhost:30001/metrics -q 'controller_runtime_reconcile_errors_total' -oyaml
- point:
    t: 1637113651012
    v: 6
  metric:
    __name__: controller_runtime_reconcile_errors_total
    controller: grafana-operator
    instance: http://localhost:30001/metrics
- point:
    t: 1637113651012
    v: 0
  metric:
    __name__: controller_runtime_reconcile_errors_total
    controller: monitoringstack
    instance: http://localhost:30001/metrics
```

```yaml
❯  ./tmp/bin/promq -t http://localhost:30001/metrics -oyaml -q 'controller_runtime_reconcile_errors_total == 0'
- point:
    t: 1637113544080
    v: 0
  metric:
    __name__: controller_runtime_reconcile_errors_total
    controller: monitoringstack
    instance: http://localhost:30001/metrics
```
Here monitoring stack controller hasn't produced any reconciliation errors, so checking if we have a result isn't sufficient. I will use use `count(controller_runtime_reconcile_errors_total != 0)` to assert that it is `0`

sthaha: I will revert to using `sum` since the number of reconciliation errors make more sense IMHO

pkg/controllers/grafana-operator/controller.go
```diff
@@ -46,7 +47,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	controllerruntime "sigs.k8s.io/controller-runtime"
```
fpetkovski, line 49: Personally, I would prefer to have this as it was before. `ctrl` is a bit ambiguous since it could also mean controller.

Also renaming it in this PR introduces a lot of unrelated changes and makes it a bit harder to review it.

pkg/controllers/grafana-operator/controller.go
```diff
@@ -400,3 +480,60 @@ func singleResourceInformer(name string, namespace string, resource string, obje
 		cache.Indexers{},
 	)
 }
+
+func creationResult(d time.Duration, err error) reconcileResult {
```
fpetkovski, line 484: I understand what this is trying to do, but is it really in the scope of reducing the error rate of the operator? I agree that we don't want extra reconcile runs, but to me this looks more like an optimization. I am also not sure whether this is the correct way to solve the problem, or if we should look into using predicates.

pkg/controllers/grafana-operator/controller.go
```diff
@@ -151,108 +159,180 @@ func RegisterWithManager(mgr controllerruntime.Manager) error {
...
 
-	return err
+	// update
+	r.logger.Info("Updating Grafana instance")
+	grafana.Spec = newGrafana().Spec
+	err = r.k8sClient.Update(ctx, &grafana)
+	return updationResult(err)
+}
+
+func (r *reconciler) reconcileGrafanaCRD(ctx context.Context) reconcileResult {
```
fpetkovski, line 310: Instead of adding additional RBAC on CRDs and adding this logic, could we simply requeue the request if we get an error that a resource does not exist?

pkg/controllers/grafana-operator/controller.go
sthaha: I found it confusing the other way .. since monitoring-stack/controller.go calls it `ctrl` and also align with the usage I have seen elsewhere e.g. - https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/scratch-env/main.go#L26

pkg/controllers/grafana-operator/controller.go
sthaha: certainly worth a try. I will see if we can detect the `CRD not present` from a creation error.

pkg/controllers/grafana-operator/controller.go
sthaha: > is it really in the scope of reducing the error rate of the operator?
yes, otherwise what happens is that the `Get` call says object not found (since the cache is invalid) but the object actually exists.

fpetkovski: Which `get` call exactly? Can we handle just that case?

pkg/controllers/grafana-operator/controller.go
fpetkovski: Ok, let's stick to this convention then.

pkg/controllers/grafana-operator/controller.go
sthaha: Without waiting for `CRD` to be `Established` we can get the following errors if `grafana` crd is deleted 
```
"error": "create not allowed while custom resource definition is terminating"}
```
The other errors like the `No matching kind` and `Not found` can be detected

fpetkovski: Did you mean can or _cannot_ be detected? If no matching kind is a detectable error, maybe that could be used to requeue the request?

sthaha: ` errors.IsNotFound(err) || errors.IsMethodNotSupported(err) ` seems sufficient to wait for CRD creation.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -118,6 +119,10 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	for _, patcher := range patchers {
 		if err := r.reconcileObject(ctx, ms, patcher); err != nil {
+			if errors.IsAlreadyExists(err) || errors.IsConflict(err) {
```
fpetkovski, line 122: Can you pull this out of the if statement in order to avoid double nesting?
```
err := r.reconcileObject(ctx, ms, patcher)
...
}

if err != nil {
...
}

File: pkg/controllers/grafana-operator/controller.go [hunk 9]
```diff
@@ -400,3 +460,60 @@ func singleResourceInformer(name string, namespace string, resource string, obje
...
+	// do not requeue if object exists
+	if errors.IsAlreadyExists(err) {
+		return next()
+	}
+
+	return reconcileError(err)
+}
+
+// returns whether to requeue
+func updationResult(err error) reconcileResult {
```
fpetkovski, line 480: Should this be in its own package?

pkg/controllers/grafana-operator/controller.go
```diff
@@ -68,13 +69,19 @@ type reconciler struct {
 	grafanaClientset rest.Interface
 }
 
+type reconcileResult struct {
```
fpetkovski, line 72: Should this struct and all of the related constructors be in its own package? I don't think this is specific to one controller

pkg/controllers/grafana-operator/controller.go
```diff
@@ -151,108 +158,161 @@ func RegisterWithManager(mgr controllerruntime.Manager) error {
...
+		return reconcileError(err)
 	}
+
+	// create
 	if errors.IsNotFound(err) {
-		r.logger.Info("Creating Grafana Operator OperatorGroup")
-		return r.k8sClient.Create(ctx, NewOperatorGroup())
+		r.logger.Info("Creating Grafana OperatorGroup")
+		err := r.k8sClient.Create(ctx, NewOperatorGroup())
+		return creationResult(5*time.Second, err)
```
fpetkovski, line 241: Do we need an explicit requeue here? Since there's a watcher on the resource, it will trigger a new reconcile again after creation.

pkg/controllers/grafana-operator/controller.go
```diff
@@ -151,108 +158,161 @@ func RegisterWithManager(mgr controllerruntime.Manager) error {
...
 	}
 
+	// create
 	if errors.IsNotFound(err) {
 		r.logger.Info("Creating Grafana instance")
-		return r.k8sClient.Create(ctx, newGrafana())
+		err := r.k8sClient.Create(ctx, newGrafana())
+
+		// can fail because grafana operator hasn't created the CRD yet
+		if errors.IsNotFound(err) || errors.IsMethodNotSupported(err) || meta.IsNoMatchError(err) {
```
fpetkovski, line 303: I am still not too comfortable with this approach, I find it hard track what is going on because the logic of what the loop should do is scattered around. Why can't we make this check in the parent function and decide whether to requeue or not based on the error that is returned by the individual reconcilers (`reconcileGrafana`, `reconcileOperatorGroup`, etc)?

sthaha: I am not sure how it is going to change ... from the outer loop is is going to be like this 

```
1. reconcile
2. stop ? then return the result
3. else  next reconcile method
```
 
The inner loop / `reconcileFoobar` knows what errors can arise and how to handle then and if it needs to requeue or not - right?

pkg/controllers/monitoring-stack/controller.go
sthaha: yes, we definitely can and reads better

pkg/controllers/grafana-operator/controller.go
fpetkovski: Alright, let's optimize this later if we have to. I have a feeling that we're missing something, but It doesn't have to be perfect from the get go.

pkg/controllers/grafana-operator/controller.go
sthaha: I agree and I felt the same but  thought we can refactor this when/if another controller takes the same approach. But I am happy to do that in this patch if you strongly feel so.

fpetkovski: Ok, maybe we can just move it to a dedicated file for now

pkg/controllers/grafana-operator/controller.go
sthaha: You are right, I don't think we should need to requeue! Good catch!

sthaha: I have updated the `creationResult` logic to terminate the loop if there are not creation errors

fpetkovski (approved):

PR #72, opened 2021-11-18
feat: query Prometheus through a dedicated service

fpetkovski:
The prometheus-operated service is a statefulset governing service. It
is intended to be used by the statefulset controller for assigning a
network identity to individual pods, and is not supposed to be used for
http communication.

This commit creates a dedicated service for each Prometheus statefulset
and reconfigures the GrafanaDataSource to query it instead.

### DIFF
--- a/pkg/controllers/monitoring-stack/components.go
+++ b/pkg/controllers/monitoring-stack/components.go
@@ -183,14 +183,14 @@ func stackComponentPatchers(ms *stack.MonitoringStack, instanceSelectorKey strin
 		},
 		{
 			empty: func() client.Object {
-				service := newAlertmanagerService(ms)
+				service := newAlertmanagerService(ms, instanceSelectorKey, instanceSelectorValue)
 				return &corev1.Service{
 					TypeMeta:   service.TypeMeta,
 					ObjectMeta: service.ObjectMeta,
 				}
 			},
 			patch: func(existing client.Object) (client.Object, error) {
-				service := newAlertmanagerService(ms)
+				service := newAlertmanagerService(ms, instanceSelectorKey, instanceSelectorValue)
 
 				if existing == nil {
 					return service, nil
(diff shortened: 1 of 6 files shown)


pkg/controllers/monitoring-stack/components.go
```diff
@@ -416,6 +447,30 @@ func newRoleBinding(ms *stack.MonitoringStack, rbacResourceName string) *rbacv1.
 	return roleBinding
 }
 
+func newService(name string, ms *stack.MonitoringStack, instanceSelectorKey string, instanceSelectorValue string) *corev1.Service {
```
sthaha, line 450: shouldn't the creation of these resources be validated in a test?

pkg/controllers/monitoring-stack/components.go
```diff
@@ -262,7 +293,7 @@ func stackComponentPatchers(ms *stack.MonitoringStack, instanceSelectorKey strin
 
 func newGrafanaDataSource(ms *stack.MonitoringStack) *grafanav1alpha1.GrafanaDataSource {
 	datasourceName := fmt.Sprintf("ms-%s-%s", ms.GetNamespace(), ms.GetName())
-	prometheusURL := fmt.Sprintf("prometheus-operated.%s:9090", ms.GetNamespace())
+	prometheusURL := fmt.Sprintf("%s-prometheus.%s:9090", ms.GetName(), ms.GetNamespace())
```
sthaha, line 296: isn't `ms.Name` and `ms.GetName()` the same?
Same with `GetNamespace()` ?

pkg/controllers/monitoring-stack/components.go
fpetkovski: Yes, this is a good point. I would just try to validate correctness by issuing queries against the service instead of running assertions on the service object fields. The reason for this is because the former is a proper behavior test, and the latter feels a bit like testing an implementation detail.

pkg/controllers/monitoring-stack/components.go
fpetkovski: Yes, they're the same. Let's stick with the property for consistency.

sthaha (approved):

PR #73, opened 2021-11-19
fix: use api-version from scheme information

sthaha:
Replaces all hard-coded APIVersion: "group/version" with
`package.SchemeGroupVersion.String()` to prevent any typos and to use the
computed value.


fpetkovski (approved):
nice!

fpetkovski:
btw can you change the commit type to refactor when you merge the PR?

PR #76, opened 2021-11-22
ci: capture cluster state after e2e run

fpetkovski:
Flaky tests can be hard to troubleshoot in the CI. This commit adds a
step in the pipeline which captures the cluster-state after each e2e
test run. It also adds a --retain-namespace flag which is toggled in the
pipeline to prevent deletion of the e2e-tests namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ KUSTOMIZE=$(TOOLS_DIR)/kustomize
 OPERATOR_SDK = $(TOOLS_DIR)/operator-sdk
 OPM = $(TOOLS_DIR)/opm
 PROMQ = $(TOOLS_DIR)/promq
+OC = $(TOOLS_DIR)/oc
 
 $(TOOLS_DIR):
 	@mkdir -p $(TOOLS_DIR)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -87,6 +88,15 @@ $(PROMQ) promq: $(TOOLS_DIR)
 		rm -rf $$TMP_DIR ;\
 	}
 
+.PHONY: oc
+$(OC) oc: $(TOOLS_DIR)
+	@{ \
+		set -ex ;\
```
sthaha, line 94: could you please add `[[ -f $(OC) ]] && exit 0 ; ` so avoid redownloading if it already exists ?

fpetkovski: Yes, good idea. Done

test/e2e/main_test.go
```diff
@@ -21,7 +22,10 @@ var (
 
 const e2eTestNamespace = "e2e-tests"
 
+var Retain = flag.Bool("retain", false, "When set, the namespace in which tests are run will not be cleaned up")
```
sthaha, line 25: nit: do we want to export this?

fpetkovski: good catch, changed to lowercase

sthaha (approved):

PR #80, opened 2021-11-24
ci: cache development tools

sthaha:
Previously, we have seen that downloading of tools like `oc` can fail due to 
transient errors. These tools are downloaded for every CI run and do not change
often. 

This PR reduces the need to download these tools over and over by making use
of  github's cache to cache all the tools downloaded by `make tools`. The cache
relies on the checksum of `.github/tools` file as key, so  any change to the 
`.github/tools` file will result in a new set of tools being downloaded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,34 @@ all: operator
 
 ## Tools
 TOOLS_DIR = $(shell pwd)/tmp/bin
+
+## NOTE: each tool must have a version that will be recorded in .github/tools
+# The .github/tools file's hash is used to compute the key for cache in github
+# see: .github/tools-cache/action.yaml
+
 CONTROLLER_GEN=$(TOOLS_DIR)/controller-gen
-GOLANGCI_LINT=$(TOOLS_DIR)/golangci-lint
+CONTROLLER_GEN_VERSION= v0.7.0
+
 KUSTOMIZE=$(TOOLS_DIR)/kustomize
+KUSTOMIZE_VERSION= v3.9.4
+
 OPERATOR_SDK = $(TOOLS_DIR)/operator-sdk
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -96,11 +117,31 @@ $(OC) oc: $(TOOLS_DIR)
...
 
 # Install all required tools
 .PHONY: tools
-tools: $(CONTROLLER_GEN) $(KUSTOMIZE) $(OPERATOR_SDK) $(OPM) $(PROMQ)
+tools: $(CONTROLLER_GEN) \
+ 		$(KUSTOMIZE) \
+		$(OC) \
+		$(OPERATOR_SDK) \
+		$(OPM) \
+	 	$(PROMQ)
```
sthaha, line 133: should we add golangci-lint  as well to the list?

fpetkovski: If it is used in the CI, then it makes sense to have it here as well

sthaha: no we don't use it in CI instead we make use of an action 

```yaml
- name: golangci-lint
  uses: golangci/golangci-lint-action@v2
```

fpetkovski: Ok in that case we can leave it out. I also think that the benefits from caching binaries will be most visible in e2e tests since they're the bottleneck.

sthaha: I believe all jobs that need to download the binaries are running faster .. 
e.g. validate  `make generate`, which needs operator-sdk and controller-gen is faster by around a minute

fpetkovski: That's true, but e2e tests take 15 minutes and are currently the bottleneck. 
I don't think speeding up linters makes difference for us.

fpetkovski (approved):

PR #84, opened 2021-11-26
feat: make the module go-gettable

fpetkovski:
This commit changes the module name so that it can be fetched with the
go get command.


sthaha (approved):
long pending :) Thanks for this patch

PR #87, opened 2021-11-29
fix: fix self-scrape prometheus configuration for stacks

fpetkovski:
The prometheus instances are incorrectly configured for self-scraping
and scrape themselves twice, once through the prometheus job and once
more through the alertmanager job. The reason for this is because they
were using the app.kubernetes.io/part-of label instead of the
app.kubernetes.io/name label to identify their own service.

### DIFF
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -210,18 +210,35 @@ func assertPrometheusScrapesItself(t *testing.T) {
 	}
 
 	promClient := framework.NewPrometheusClient("http://localhost:9090")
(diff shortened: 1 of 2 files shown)


test/e2e/monitoring_stack_controller_test.go
```diff
@@ -210,18 +210,35 @@ func assertPrometheusScrapesItself(t *testing.T) {
 	}
 
 	promClient := framework.NewPrometheusClient("http://localhost:9090")
+	expectedResults := map[string]int{
```
sthaha, line 213: nice!

sthaha (approved):

PR #88, opened 2021-12-01
fix: prevent automatic upgrades of Grafana Operator

fpetkovski:
The Grafana Operator is a third party bundle and its releases are not in
our control. The 4.0.2 release was buggy and should not have been used at all.

This commit makes sure that new versions are not installed
automatically.

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -18,11 +18,11 @@ import (
 	"fmt"
 	"time"
 
+	networkingv1 "k8s.io/api/networking/v1"
+
 	"github.com/rhobs/monitoring-stack-operator/pkg/eventsource"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	networkingv1 "k8s.io/api/networking/v1"
 
 	integreatlyv1alpha1 "github.com/grafana-operator/grafana-operator/v4/api/integreatly/v1alpha1"
 	v1 "k8s.io/api/apps/v1"
@@ -59,6 +59,7 @@ const (
 	subscriptionName  = "monitoring-stack-operator-grafana-operator"
 	operatorGroupName = "monitoring-stack-operator-grafana-operator"
 	grafanaName       = "monitoring-stack-operator-grafana"
+	grafanaCSV        = "grafana-operator.v4.0.1"
 )
 
 type reconciler struct {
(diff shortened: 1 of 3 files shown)


pkg/controllers/grafana-operator/controller.go
```diff
@@ -276,6 +291,42 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult
...
 }
 
+func (r *reconciler) approveInstallPlan(ctx context.Context) reconcileResult {
+	subscriptionKey := types.NamespacedName{
+		Name:      subscriptionName,
+		Namespace: Namespace,
+	}
+	var subscription v1alpha1.Subscription
+	if err := r.k8sClient.Get(ctx, subscriptionKey, &subscription); err != nil {
+		return updationResult(err)
```
sthaha, line 301: shouldn't this be a `reconcileError(err)` ?

pkg/controllers/grafana-operator/controller.go
```diff
@@ -276,6 +291,42 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult
...
+	}
+
+	installPlanRef := subscription.Status.InstallPlanRef
+	if installPlanRef == nil {
+		return reconcileResult{
```
sthaha, line 306: return end()

pkg/controllers/grafana-operator/controller.go
```diff
@@ -276,6 +291,42 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult
...
+	}
+	var installPlan v1alpha1.InstallPlan
+	err := r.k8sClient.Get(ctx, installPlanKey, &installPlan)
+	if err != nil {
+		return reconcileError(err)
+	}
+
+	// Only approve install plans for the specified version
+	if installPlan.Spec.ClusterServiceVersionNames[0] != grafanaCSV {
+		return updationResult(nil)
```
sthaha, line 323: did you want to `return next()` ?

fpetkovski: I don't know what these terms mean :)

fpetkovski: I just want to return a no-op basically

sthaha: 😂
`next` moves to the next step in the reconcile loop
`end` terminates the reconcile loop

fpetkovski: Ok, updated

sthaha (approved):

PR #90, opened 2021-12-02
docs: auto generate CRD docs

fpetkovski:
Here is how the docs generated by this tool look like: https://github.com/fpetkovski/monitoring-stack-operator/blob/crd-docs/docs/api.md

The only downside of the tool is that it's not able to link to external resources because it relies to yaml files instead of go structs.

We can look into an alternative solution if this one's too verbose.

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -1,0 +1,211 @@
+SHELL=/usr/bin/env bash -o pipefail
+
+## Tools
+TOOLS_DIR = $(shell pwd)/tmp/bin
+
+## NOTE: each tool must have a version that will be recorded in .github/tools
+# The .github/tools file's hash is used to compute the key for cache in github
+# see: .github/tools-cache/action.yaml
+
+CONTROLLER_GEN=$(TOOLS_DIR)/controller-gen
+CONTROLLER_GEN_VERSION= v0.7.0
+
+KUSTOMIZE=$(TOOLS_DIR)/kustomize
+KUSTOMIZE_VERSION= v3.9.4
+
+OPERATOR_SDK = $(TOOLS_DIR)/operator-sdk
+OPERATOR_SDK_VERSION = v1.13.0
+
+OPM=$(TOOLS_DIR)/opm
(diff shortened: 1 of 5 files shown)


Makefile.tools
```diff
@@ -0,0 +1,146 @@
...
+## NOTE: promq does not have any releases, so we use a fake version starting with v0.0.1
+# thus to upgrade/invalidate the github cache, increment the value
+PROMQ = $(TOOLS_DIR)/promq
+PROMQ_VERSION = v0.0.1
+
+# NOTE: oc is NOT downloadable using the OC_VERSION in its URL, so this has to be manually updated
+OC = $(TOOLS_DIR)/oc
+OC_VERSION = v4.9.8
+
+CRDOC = $(TOOLS_DIR)/crdoc
```
sthaha, line 34: could you please modify the tools/cache action to validate / print the `crdoc` version

fpetkovski: Can you ptal again? I moved this part to the Makefile so that it's more concise and we know that we need to update it when we add/remove tools

sthaha: Sorry, I missed this update. Yes, it looks good to me but needs rebasing :(

sthaha (approved):

Makefile.tools
```diff
@@ -0,0 +1,210 @@
...
+		$(OC) \
+		$(OPERATOR_SDK) \
+		$(OPM) \
+	 	$(PROMQ) \
+	 	$(CRDOC) \
+	 	jsonnet-tools
+	@{ \
+		set -ex ;\
+		tools_file=.github/tools ;\
+		>$$tools_file ;\
```
sthaha, line 184: This misses the `### DO NOT EDIT ###` warning that was added recently in master .. is that on purpose?

sthaha (dismissed):

Makefile.tools
fpetkovski: I think I removed it by accident, reverted.

sthaha (approved):

PR #91, opened 2021-12-02
feat: deploy alertmanagers on different nodes

fpetkovski:
This commit ensures that Alertmanager instances are deployed on
different nodes. It also reduces the number of instances to 2 to make
sure we don't have unnecessary resource usage.

### DIFF
--- a/pkg/controllers/monitoring-stack/alertmanager.go
+++ b/pkg/controllers/monitoring-stack/alertmanager.go
@@ -20,7 +20,7 @@ func newAlertmanager(
 	if resourceSelector == nil {
 		resourceSelector = &metav1.LabelSelector{}
 	}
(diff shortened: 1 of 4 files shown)


pkg/controllers/monitoring-stack/alertmanager.go
```diff
@@ -20,7 +20,7 @@ func newAlertmanager(
 	if resourceSelector == nil {
 		resourceSelector = &metav1.LabelSelector{}
 	}
-	replicas := int32(3)
+	replicas := int32(2)
```
fpetkovski, line 23: We are making the same change for CMO: https://github.com/openshift/cluster-monitoring-operator/pull/1489

sthaha (approved):

PR #93, opened 2021-12-06
fix: update operator image name

slashpai:
since operator image is in quay.io, without speciying full registry unable to deploy operator

Signed-off-by: Jayapriya Pai <janantha@redhat.com>

### DIFF
--- a/deploy/operator/kustomization.yaml
+++ b/deploy/operator/kustomization.yaml
@@ -8,5 +8,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: monitoring-stack-operator
+  newName: quay.io/rhobs/monitoring-stack-operator
   newTag: 0.0.6
 namespace: operators
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ SHELL=/usr/bin/env bash -o pipefail
 
 # IMAGE_BASE defines the registry/namespace and part of the image name
 # This variable is used to construct full image tags for bundle and catalog images.
-IMAGE_BASE ?= monitoring-stack-operator
+IMAGE_BASE ?= quay.io/rhobs/monitoring-stack-operator
 
 
 VERSION ?= $(shell cat VERSION)


sthaha (commented):
This is on purpose! This prevents us from pushing images by mistake. Please make use of IMAGE_BASE when creating the operator image. 
e.g. 

```
make release IMAGE_BASE=quay.io/sthaha/monitoring-stack-operator
```

deploy/operator/kustomization.yaml
```diff
@@ -8,5 +8,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: monitoring-stack-operator
+  newName: quay.io/rhobs/monitoring-stack-operator
```
slashpai, line 11: @sthaha we can update here so that deployment works? i was trying to deploy existing image

sthaha: you want to use OLM  for deployment instead of `kubectl create -f` . e.g 

```
 make operator-image bundle-image operator-push bundle-push IMAGE_BASE=local-registry:30000/monitoring-stack-operator VERSION=0.0.0-ci

./tmp/bin/operator-sdk run bundle local-registry:30000/monitoring-stack-operator-bundle:0.0.0-ci --install-mode AllNamespaces --skip-tls --namespace operators --verbose | tee tmp/olm.log
```
note: `local-registry` can also be used for pushing operator-images

slashpai: I was not using olm. I just ran kubectl create since image was not in docker hub it was erroring out when I updated the image to quay.io it was working. Was the manifest meant to use for local deployment?

sthaha: You will need to install OLM anyway for the operator to work properly since it creates a grafana subscription and requires OLM. 

For deployment of the operator into the cluster, my recommendation will be to create the bundle which packages all required CRD,  roles and deployments and use `operator-sdk run bundle` to run the operator. 

See: how e2e-tests workflow which installs the operator and then runs test works.

sthaha: If this is all too confusing I think we could create a make target `operator-deploy` which does this. Also update the docs on how to run deploy the operator in a cluster.

slashpai: Thanks Sunil. I will update the docs once I try this out

slashpai:
closing the PR based on above conversation

PR #94, opened 2021-12-07
feat: add alert rules for mso

slashpai:
This commit adds initial rule file for monitoring-stack-operator

Signed-off-by: Jayapriya Pai janantha@redhat.com

**Description:**

Add alert rules for controller, create service and servicemonitor

### DIFF
--- a/deploy/operator/monitoring-stack-operator-rules.yaml
+++ b/deploy/operator/monitoring-stack-operator-rules.yaml
@@ -1,0 +1,49 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    app.kubernetes.io/component: operator
+    app.kubernetes.io/name: monitoring-stack-operator
+    app.kubernetes.io/part-of: monitoring-stack-operator
+    prometheus: k8s
+    role: alert-rules
+  name: monitoring-stack-operator-prometheus-rules
+spec:
+  groups:
+  - name: monitoring-stack-operator.rules
+    rules:
+    - alert: Monitoring Stack Operator controller reconcile errors
+      annotations:
+        description: |
+          Monitoring Stack Operator controller - {{ $labels.controller }} fails to reconcile.
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator controller - {{ $labels.controller }} fails to reconcile
+      expr: |
+        increase(controller_runtime_reconcile_errors_total{job="monitoring-stack-operator"}[15m]) > 0
+      for: 15m
+      labels:
+        severity: warning
+    - alert: Monitoring Stack Operator controller reconcilation takes longer than 10 minutes
+      annotations:
+        description: |
+          Monitoring Stack Operator controller reconcilation takes longer than 10 minutes for the controller - {{ $labels.controller }}.
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator controller - {{ $labels.controller }} reconcilation takes too long to reconcile
+      expr: |
+        rate(controller_runtime_reconcile_time_seconds_sum{job="monitoring-stack-operator"}[5m]) /
+        rate(controller_runtime_reconcile_time_seconds_count{job="monitoring-stack-operator"}[5m])
+        > 600
+      for: 10m
+      labels:
+        severity: warning
+    - alert: Monitoring Stack Operator controller backlog is not being drained
+      annotations:
+        description: |
+          The backlog of Monitoring Stack Operator controller - {{ $labels.name }} is not getting drained; an indication that reconcile loop may be stuck
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator controller - {{ $labels.name }} backlog is not being drained.
+      expr: |
+        rate(workqueue_depth{job="monitoring-stack-operator"}[15m]) > 0
+      for: 15m
+      labels:
+        severity: critical
--- a/deploy/operator/monitoring-stack-operator-service-monitor.yaml
+++ b/deploy/operator/monitoring-stack-operator-service-monitor.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: monitoring-stack-operator
+  labels:
+    app.kubernetes.io/component: operator
+    app.kubernetes.io/name: monitoring-stack-operator
+    app.kubernetes.io/part-of: monitoring-stack-operator
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: monitoring-stack-operator
+      app.kubernetes.io/component: operator
+  endpoints:
+  - port: metrics
--- a/deploy/operator/kustomization.yaml
+++ b/deploy/operator/kustomization.yaml
@@ -3,6 +3,9 @@ resources:
 - monitoring-stack-operator-service-account.yaml
 - monitoring-stack-operator-cluster-role.yaml
 - monitoring-stack-operator-cluster-role-binding.yaml
+- monitoring-stack-operator-service.yaml
+- monitoring-stack-operator-service-monitor.yaml
+- monitoring-stack-operator-rules.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
--- a/deploy/operator/monitoring-stack-operator-service.yaml
+++ b/deploy/operator/monitoring-stack-operator-service.yaml
@@ -1,0 +1,15 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: monitoring-stack-operator
+  labels:
+    app.kubernetes.io/component: operator
+    app.kubernetes.io/name: monitoring-stack-operator
+    app.kubernetes.io/part-of: monitoring-stack-operator
+spec:
+  selector:
+    app.kubernetes.io/name: monitoring-stack-operator
+    app.kubernetes.io/component: operator
+  ports:
+  - name: metrics
+    port: 8080


deploy/samples/monitoring-stack.yaml
```diff
@@ -7,4 +7,4 @@ spec:
   retention: 1d
   resourceSelector:
     matchLabels:
-      system: foo
+      app.kubernetes.io/part-of: monitoring-stack-operator
```
slashpai, line 10: I updated this to `app.kubernetes.io/part-of: monitoring-stack-operator` since resourceSelector label need to match the object label while creating prometheus cr's, I either needed to add `system:foo` to cr's or just update the label here. Looking for better way to handle if this is not the right way

slashpai:
@fpetkovski @sthaha PTAL

deploy/samples/monitoring-stack.yaml
fpetkovski: Hi JP, I wonder if we the stacks should collect these service monitors. I would expect them to end up in some other prometheus since they're related to the operator itself. So I don't think we should worry too much about matching the selectors of the stack resource and the service monitor(s) for the operator.

slashpai: sure I added that since I was trying to test alert I will remove that. We want to keep only rules file?

fpetkovski: I think it's fine to keep the ServiceMonitor as well. Let's just make sure it selects the component label from the pods, not just the `app.kubernetes.io/name`:
https://github.com/rhobs/monitoring-stack-operator/blob/main/deploy/operator/monitoring-stack-operator-deployment.yaml#L18-L20

slashpai: updated, I generated bundle locally it works fine

I looked at https://docs.openshift.com/container-platform/4.9/operators/understanding/olm-packaging-format.html#olm-bundle-format_olm-packaging-format servicemonitor and prometheusrule are supported for bundle

So a question is since the service, servicemonitor and prometheusrule will be only used when prometheus resource is created deploying them by default is ok?

fpetkovski: I think deploying them by default is okay, we just have to make sure that they don't get picked up by some prometheus deployed by a monitoring stack. 

We should probably document how to collect these monitoring resources in a Prometheus instance, I think SRE will definitely be interested in that.

slashpai: Without a matching resourceSelector label it doesn't get picked up by default, cant we use that logic? I meant while creating monitoringstack cr we can specify the label we should use as in this sample-monitoring stack file?

deploy/olm/kustomization.yaml
```diff
@@ -10,4 +10,4 @@ kind: Kustomization
 images:
 - name: monitoring-stack-operator
   newName: monitoring-stack-operator
-  newTag: 0.0.2
+  newTag: 0.0.6
```
sthaha, line 13: Unrelated change?

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,25 @@
...
+    prometheus: k8s
+    role: alert-rules
+  name: monitoring-stack-operator-prometheus-rules
+spec:
+  groups:
+  - name: monitoring-stack-operator.rules
+    rules:
+    - alert: MonitoringStackOperatorReconcileErrors
+      annotations:
+        description: MonitoringStackOperator failing to reconcile controller {{ $labels.controller
```
sthaha, line 17: How about `Monitoring Stack Operator controller - {{$labels.controller}} fails to reconcile`

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,25 @@
...
+              }}
+        summary: MonitoringStack Operator failing to reconcile controller {{ $labels.controller
```
sthaha, line 19: Same - `Monitoring Stack Operator controller - {{$labels.controller}} fails to reconcile`

deploy/operator/monitoring-stack-operator-service-monitor.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: monitoring-stack-operator
+  labels:
```
sthaha, line 5: I wonder if we can make use of `commonLabels` in `kustomize.yaml` ?

deploy/samples/monitoring-stack.yaml
sthaha: Since the goal of this PR is to generate alerts if **MSO** fails to reconcile, should we worry about the stacks we deploy? Not only that, these rules don't apply to stack we deploy. Hence ... we shouldn't need to modify the sample stack - cr

deploy/olm/kustomization.yaml
slashpai: ya i can remove that, got update when I ran `make bundle`

deploy/operator/monitoring-stack-operator-service-monitor.yaml
slashpai: ya we can, I thought to have it in a separate PR since I need to update other files also  unrelated to this  change

deploy/samples/monitoring-stack.yaml
slashpai: The prometheus resources servicemonitor, prometheusrule were not picking by operator unless the resourceSelector matching label is there. I had updated to test this change but I can remove.

slashpai: I reverted it to same as before but to confirm we can use this logic to make sure it doesn't picked by monitoring-stack-operator by default unless a matchingLabel found?

deploy/operator/monitoring-stack-operator-service-monitor.yaml
fpetkovski: The problem with commonLabels is that it sets selectors as well. So we have to make sure we don't have any values that change (e.g. the version) in those labels.

slashpai: ya we can keep component, name and part-of?

fpetkovski: I think we can only have `part-of`, no? Otherwise every component will have the same name and same component.

deploy/samples/monitoring-stack.yaml
slashpai: @fpetkovski wdyt?

fpetkovski:
This looks good so far. Do we want to monitor additional metrics, e.g. queue depth?

slashpai:
> This looks good so far. Do we want to monitor additional metrics, e.g. queue depth?

Ya I will add other rules as well, created initial PR with only one just to get an idea. Also since we are not using jsonnet how do we want to include other alerts like monitoring prometheus-operator?

fpetkovski:
Yes, that's a good point. We should include PO alerts as well.

slashpai:
@fpetkovski I added couple of alerts as well, PTAL. I can create separate PR for Prometheus Operator

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  groups:
+  - name: monitoring-stack-operator.rules
+    rules:
+    - alert: MonitoringStackOperatorReconcileErrors
+      annotations:
+        description: Monitoring Stack Operator controller - {{ $labels.controller }} fails to reconcile. Inspect the
+          monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator controller - {{ $labels.controller }} fails to reconcile
+      expr: increase(controller_runtime_reconcile_errors_total{job="monitoring-stack-operator"}[15m])
```
sthaha, line 20: nit: could you consider making use of `|` or `>` to make this more readable 

e.g.
```yaml
description: >- # or even |-
  Monitoring Stack Operator controller - {{ $labels.controller }} fails to reconcile. 
  Inspect the monitoring-stack-operator log for potential root causes.
expr: |
  increase(controller_runtime_reconcile_errors_total{job="monitoring-stack-operator"[15m]} > 0

 
```

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,44 @@
...
+        > 0
+      for: 15m
+      labels:
+        severity: warning
+    - alert: MonitoringStackOperatorReconcilation takes longer than 10 minutes
```
sthaha, line 25: `Monitoring Stack Operator reconciliation`  to be consistent with the rest of the alerts

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,44 @@
...
+      annotations:
+        description: MonitoringStackOperator reconcilation takes longer than 10 minutes for the controller - {{ $labels.controller }}.
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: MonitoringStackOperator reconcilation takes longer for controller - {{ $labels.controller }}
+      expr: rate(controller_runtime_reconcile_time_seconds_sum{job="monitoring-stack-operator"}[5m]) /
+            rate(controller_runtime_reconcile_time_seconds_count{job="monitoring-stack-operator"}[5m]) > 600
+      for: 10m
+      labels:
+        severity: warning
+    - alert: MonitoringStackController backlog is not being drained
```
sthaha, line 35: Monitoring Stack Operator

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,44 @@
...
+      annotations:
+        description: The backlog for {{ $labels.name }} is not getting drained. It is an indication of reconciliation loop of controller got stuck.
```
sthaha, line 37: > It is an indication of reconciliation loop of controller got stuck.

The backlog of Monitoring Stack Operator - controller {{ $labels.name }} is not getting drained; an indication that reconcile loop may be stuck.

sthaha (commented):
Nice patch @slashpai ! 
Looks good overall. Some minor corrections

slashpai:
> Nice patch @slashpai ! Looks good overall. Some minor corrections

Thanks Sunil. I updated with changes, PTAL

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,49 @@
...
+        increase(controller_runtime_reconcile_errors_total{job="monitoring-stack-operator"}[15m]) > 0
+      for: 15m
+      labels:
+        severity: warning
+    - alert: Monitoring Stack Operator reconcilation takes longer than 10 minutes
+      annotations:
+        description: |
+          Monitoring Stack Operator reconcilation takes longer than 10 minutes for the controller - {{ $labels.controller }}.
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator reconcilation takes longer for controller - {{ $labels.controller }}
```
sthaha, line 31: > takes a long time to reconcile 
or to be consisten ... 

`Monitoring Stack Operator controller - {{ $labels.controller }} takes too long to reconcile`

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,49 @@
...
+        > 600
+      for: 10m
+      labels:
+        severity: warning
+    - alert: Monitoring Stack Operator backlog is not being drained
+      annotations:
+        description: |
+          The backlog of Monitoring Stack Operator - {{ $labels.name }} is not getting drained; an indication that reconcile loop may be stuck
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator - {{ $labels.name }} backlog is not being drained.
```
sthaha, line 44: `Monitoring Stack Operator controller - {{ $labels.name }}` - right? Isn't `labels.name` the controller name?

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,49 @@
...
+        rate(controller_runtime_reconcile_time_seconds_sum{job="monitoring-stack-operator"}[5m]) /
+        rate(controller_runtime_reconcile_time_seconds_count{job="monitoring-stack-operator"}[5m])
+        > 600
+      for: 10m
+      labels:
+        severity: warning
+    - alert: Monitoring Stack Operator backlog is not being drained
+      annotations:
+        description: |
+          The backlog of Monitoring Stack Operator - {{ $labels.name }} is not getting drained; an indication that reconcile loop may be stuck
```
sthaha, line 42: nit: We can be consistent in our alert messages .. e.g. 

Monitoring Stack Operator controller - {{ $labels.name }}  backlog  is no getting drained; an indication that reconcile loop may be stuck.

deploy/operator/monitoring-stack-operator-rules.yaml
slashpai: yes its controller name

sthaha (approved):
nice!

slashpai:
@fpetkovski are we good to merge? or anymore change is required :)

deploy/operator/monitoring-stack-operator-rules.yaml
```diff
@@ -0,0 +1,49 @@
...
+        severity: warning
+    - alert: Monitoring Stack Operator controller reconcilation takes longer than 10 minutes
+      annotations:
+        description: |
+          Monitoring Stack Operator controller reconcilation takes longer than 10 minutes for the controller - {{ $labels.controller }}.
+          Inspect the monitoring-stack-operator log for potential root causes.
+        summary: Monitoring Stack Operator controller - {{ $labels.controller }} reconcilation takes too long to reconcile
+      expr: |
+        rate(controller_runtime_reconcile_time_seconds_sum{job="monitoring-stack-operator"}[5m]) /
+        rate(controller_runtime_reconcile_time_seconds_count{job="monitoring-stack-operator"}[5m])
```
fpetkovski, line 34: What is a rate of a count in this case?

fpetkovski: Ah ok so this is the average loop time, makes sense :+1:

fpetkovski (approved):

PR #96, opened 2021-12-10
fix: enabling deleted grafana datasources to be recreated

prashbnair:


pkg/controllers/monitoring-stack/controller.go
```diff
@@ -42,6 +46,11 @@ import (
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 )
 
+const (
+	grafanaDatasourceOwnerName      = "grafana.datasource.owner/name"
```
fpetkovski, line 50: Can we use the following annotations here?
```
monitoring-stack-operator/owner-name
monitoring-stack-operator/owner-namespace
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -100,7 +109,13 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
 		Owns(&rbacv1.RoleBinding{}).WithEventFilter(p).
 		Owns(&monv1.ServiceMonitor{}).WithEventFilter(p).
 		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
-		Complete(r)
+		Watches(&source.Kind{Type: &grafanav1alpha1.GrafanaDataSource{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
+			name := object.GetAnnotations()[grafanaDatasourceOwnerName]
+			namespace := object.GetAnnotations()[grafanaDatasourceOwnerNamespace]
+			namespaceName := types.NamespacedName{Name: name, Namespace: namespace}
+			return []reconcile.Request{{namespaceName}}
+
```
fpetkovski, line 117:
```suggestion
```

fpetkovski (commented):
Looks good, just left a few small nits.

fpetkovski:
@sthaha do you think we want an e2e test here?

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/components.go
```diff
@@ -318,15 +318,21 @@ func stackComponentPatchers(ms *stack.MonitoringStack, instanceSelectorKey strin
...
 	return &grafanav1alpha1.GrafanaDataSource{
 		TypeMeta: metav1.TypeMeta{
 			// NOTE: uses a different naming convention for SchemeGroupVersion
 			APIVersion: grafanav1alpha1.GroupVersion.String(),
 			Kind:       "GrafanaDataSource",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      datasourceName,
-			Namespace: grafana_operator.Namespace,
+			Name:        datasourceName,
```
sthaha, line 333: I think should take the opportunity to fix this name collision as well.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -100,7 +110,12 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
 		Owns(&rbacv1.RoleBinding{}).WithEventFilter(p).
 		Owns(&monv1.ServiceMonitor{}).WithEventFilter(p).
 		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
-		Complete(r)
+		Watches(&source.Kind{Type: &grafanav1alpha1.GrafanaDataSource{}}, handler.EnqueueRequestsFromMapFunc(func(object client.Object) []reconcile.Request {
```
sthaha, line 113: nice!

sthaha:
> @sthaha do you think we want an e2e test here?

I agree, we should test if operator restores datasource if it is modified or deleted. I suspect that given then current code that it might crash if the datasource is deleted. 
@prashbnair  have you tested this?

prashbnair:
> I agree, we should test if operator restores datasource if it is modified or deleted. I suspect that given then current code that it might crash if the datasource is deleted.

Yes this has been tested.

prashbnair:
> I think should take the opportunity to fix this name collision as well.

Can we do this in  a separate PR?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -100,7 +110,12 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
...
+			name := object.GetAnnotations()[grafanaDatasourceOwnerName]
+			namespace := object.GetAnnotations()[grafanaDatasourceOwnerNamespace]
+			namespaceName := types.NamespacedName{Name: name, Namespace: namespace}
+			return []reconcile.Request{{NamespacedName: namespaceName}}
```
sthaha, line 117: nit: `namespace*d*Name` to match `types.Namespace*d*Name`

pkg/controllers/monitoring-stack/controller.go
sthaha: This does fail for me on a new cluster ... 
```
2021-12-14T12:31:27.489+1000    INFO    grafana-operator        Creating Grafana OperatorGroup
2021-12-14T12:31:27.489+1000    ERROR   controller-runtime.manager.controller.monitoringstack   Could not wait for Cache to sync        {"reconciler group": "monitoring.rhobs", "reconciler kind": "MonitoringStack", "error": "failed to wait for monitoringstack caches to sync: no matches for kind \"GrafanaDataSource\" in version \"integreatly.org/v1alpha1\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
        /home/sthaha/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:195
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start
        /home/sthaha/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:221
sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).startRunnable.func1
        /home/sthaha/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/manager/internal.go:696
2021-12-14T12:31:27.489+1000    INFO    grafana-operator.controller.grafana-operator    Shutdown signal received, waiting for all workers to finish
2021-12-14T12:31:27.490+1000    ERROR   grafana-operator        End Reconcile - creation or updation error      {"req": "/", "error": "context canceled"}
```

sthaha (changes requested):

sthaha:
> Can we do this in a separate PR?

Sure thing :-)

pkg/controllers/monitoring-stack/controller.go
sthaha: I dug around a bit ... 
https://github.com/rhobs/monitoring-stack-operator/pull/100/files may be a way to solve it. What I wanted to do  was to add a watch dynamically.  

see also: https://github.com/kubernetes-sigs/controller-runtime/issues/246

sthaha:
@prashbnair , Could you please also ensure that deletion a namespace that contains monitoring-stacks also cleans up their associated grafana-datasources ?

sthaha:
As discussed, it would be nice to have [all reconciliation issues](https://github.com/rhobs/monitoring-stack-operator/issues/101) sorted in a single or multiple PRs

pkg/controllers/monitoring-stack/controller.go
prashbnair: isnt namespaceName a better name for a variable?

prashbnair:
@sthaha I will address the clean up of grafana data sources when the monitoring stack or namespace is deleted as a separate PR.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -196,3 +219,34 @@ func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringSt
 	logger.Info("Updating stack component")
 	return r.k8sClient.Update(ctx, desired)
 }
+
+func (r *reconciler) createGrafanaDSWatch(ctx context.Context) error {
+	if r.shouldCreateGrafanaDSWatch {
```
sthaha, line 224: nit: it may be more readable if exited early .. e..g

```go 
if !r.shouldCreateGrafanaDSWatch {
  return nil
}

// rest of the code to setup the watch ...
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -196,3 +219,34 @@ func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringSt
...
+		log := r.logger.WithName("create-grafana-ds-watch")
+		var (
+			dataSources grafanav1alpha1.GrafanaDataSourceList
+		)
```
sthaha, line 228: nit: `var datasources ....`

pkg/controllers/monitoring-stack/controller.go
sthaha: its a `nit` so please feel free to ignore it ... I wouldn't mind if the var was called `nn` :D 
but the point I was trying to make was consistency. 

IMHO  `sheep := Sheep()` is better than `shep := Sheep()` which is better than  ` sheep := Dog()`

prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -196,3 +219,35 @@ func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringSt
...
+		return nil
+	}
+	log := r.logger.WithName("create-grafana-ds-watch")
+	var dataSources grafanav1alpha1.GrafanaDataSourceList
+
+	err := r.k8sClient.List(ctx, &dataSources, client.InNamespace("default"))
+
+	if err != nil {
+		log.V(3).Info("grafana data source CRD is not defined")
+		return err
```
sthaha, line 234: I think we should follow same the implementation in @fpetkovski 's PR ... https://github.com/rhobs/monitoring-stack-operator/pull/103/files#diff-606d75b2cccce3c13550d55e0c1b27bb1c1fad0431b4f16d466e41dd4e0c801dR355 


I.E. function should return `requeue (bool)` and `error`

```
func setWatch() (requeue bool, err error){

  if err = listDataSources(); err != nil 
    return true, nil 
  }
  err = setWatch()
  return false, err
}
```

So that the caller (Reconcile func) can ... 

```go
if requeue, err := r.setGrafanaDatasourceWatch(); requeue || err != nil {
   if err {
     return ctrl.Result{}, err
  }
  return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
}
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -163,6 +167,24 @@ func reconcileRevertsManualChanges(t *testing.T) {
...
+func recreateDeleteGrafanaDS(t *testing.T) {
+	ms := newMonitoringStack(t, "test-grafana-ds")
+	datasourceName := fmt.Sprintf("ms-%s-%s", ms.Namespace, ms.Name)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := grafanav1alpha1.GrafanaDataSource{}
+
+	f.GetResourceWithRetry(t, datasourceName, "monitoring-stack-operator", &grafanaDS)
```
sthaha, line 179: "monitoring-stack-operator" -> grafanaop.Namespace

sthaha: Is it possible to  make use of `AssertResourceEventuallyExists` ? to avoid Get & assert?

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -163,6 +167,24 @@ func reconcileRevertsManualChanges(t *testing.T) {
...
+
+	assert.NilError(t, err, "grafana data source is not created")
+
+	f.K8sClient.Delete(context.Background(), &grafanaDS)
+
+	f.GetResourceWithRetry(t, datasourceName, "monitoring-stack-operator", &grafanaDS)
```
sthaha, line 185: I think we should make use of `AssertResourceEventuallyExists` here as well

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -80,28 +93,35 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
...
 		WithLogger(ctrl.Log).
 		For(&stack.MonitoringStack{}).
 		Owns(&monv1.Prometheus{}).WithEventFilter(p).
 		Owns(&v1.ServiceAccount{}).WithEventFilter(p).
 		Owns(&rbacv1.Role{}).WithEventFilter(p).
 		Owns(&rbacv1.RoleBinding{}).WithEventFilter(p).
 		Owns(&monv1.ServiceMonitor{}).WithEventFilter(p).
-		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
-		Complete(r)
+		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).Build(r)
```
sthaha, line 118: nit: `.Build` in a separate line just like how we used to have `.Complete`

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -42,12 +48,19 @@ import (
...
-	scheme                *runtime.Scheme
-	logger                logr.Logger
-	instanceSelectorKey   string
-	instanceSelectorValue string
+	k8sClient                  client.Client
+	scheme                     *runtime.Scheme
+	logger                     logr.Logger
+	instanceSelectorKey        string
+	instanceSelectorValue      string
+	shouldCreateGrafanaDSWatch bool
```
sthaha, line 62: may want to use the name var name as Filips patch for consistent. [`grafanaWatchEstablished`](https://github.com/rhobs/monitoring-stack-operator/pull/103/files#diff-606d75b2cccce3c13550d55e0c1b27bb1c1fad0431b4f16d466e41dd4e0c801dR69)

test/e2e/monitoring_stack_controller_test.go
prashbnair: `AssertResourceEventuallyExists` is created for statefulset,  there will need to be some refactoring done to make it work.

pkg/controllers/monitoring-stack/controller.go
prashbnair: All the patchers in this `reconcile` function return an `error` , I have followed the same model.

sthaha: This part is different from the rest of the patchers. This isn't a patcher to start with :-) 

More importantly ... 
We need the ability to `requeue`  *without* error if GrafanaDataSource CRD is not created and to log **real** errors when setting up a watch fails.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -196,3 +220,35 @@ func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringSt
...
+func (r *reconciler) createGrafanaDSWatch(ctx context.Context) error {
+	if r.grafanaDSWatchCreated {
+		return nil
+	}
+	log := r.logger.WithName("create-grafana-ds-watch")
+	var dataSources grafanav1alpha1.GrafanaDataSourceList
+
+	err := r.k8sClient.List(ctx, &dataSources, client.InNamespace("default"))
+
+	if err != nil {
```
sthaha, line 233: Failing to list shouldn't return a error since it isn't an actual reconcile error.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -122,6 +143,16 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	requeue, err := r.createGrafanaDSWatch(ctx)
+
+	if requeue {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
sthaha, line 155: *nit:* always `err` check first 
Go also allows of this to be written succinctly while keeping the scope of these variables minimal as follows 

```go 

if requeue, err := r.createGrafanaDSWatch(ctx); err != nil {
  return ctrl.Result{}, err
} else if requeue {
   return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
}
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -163,6 +169,22 @@ func reconcileRevertsManualChanges(t *testing.T) {
...
 
+func recreateDeleteGrafanaDS(t *testing.T) {
+	ms := newMonitoringStack(t, "test-grafana-ds")
+	datasourceName := fmt.Sprintf("ms-%s-%s", ms.Namespace, ms.Name)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := grafanav1alpha1.GrafanaDataSource{}
+
```
sthaha, line 180: nit: do we need a blank line?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -196,3 +226,35 @@ func (r *reconciler) reconcileObject(ctx context.Context, ms *stack.MonitoringSt
...
+	if r.grafanaDSWatchCreated {
+		return false, nil
+	}
+	log := r.logger.WithName("create-grafana-ds-watch")
+	var dataSources grafanav1alpha1.GrafanaDataSourceList
+
+	err := r.k8sClient.List(ctx, &dataSources, client.InNamespace("default"))
+
+	if err != nil {
+		log.V(3).Info("grafana data source CRD is not defined")
```
sthaha, line 240: IIRC, @fpetkovski changed these log levels to `V(6)` so you may want to keep it the same.

sthaha (approved):
Good stuff @prashbnair 

You may want to consider incorporating the nits

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -122,6 +143,16 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if requeue, err := r.createGrafanaDSWatch(ctx); err != nil {
+		return ctrl.Result{}, err
+	} else if requeue {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
sthaha, line 155: this isn't needed now

test/e2e/framework/assertions.go
```diff
@@ -82,7 +82,7 @@ func (f *Framework) AssertResourceEventuallyExists(name, namespace string, resou
 			}
 			return false, nil
 		}); err == wait.ErrWaitTimeout {
-			t.Fatal(fmt.Errorf("statefulset %s/%s was never created", namespace, name))
+			t.Fatal(fmt.Errorf("resource %s/%s was never created", namespace, name))
```
sthaha, line 85: good catch!

sthaha (approved):
🎉

sthaha (approved):

PR #97, opened 2021-12-13
fix: install-plan approval logic to approve the right plan

sthaha:
Previously, install-plan for the desired version of Grafana Operator was
obtained by lookup the status.installPlanRef of the subscription. This
isn't reliable since the installPlanRef can point to a version higher
than desired and the reconcile loop would wait for ever without
installing Grafana.

E.g. say we wanted to install `4.0.1` and the installPlanRef refers to
`4.0.2` because by the time reconcile loop was trigger, OLM already
updated the installPlanRef to the next upgradable version - 4.0.2, the
grafana controller would wait for ever to install the operator.

This is fixed by following the steps below.
  1. Iterate  all install-plans in the monitoring-stack-operator namespace
  2. Filter the plan that matches the desired CSV version
  3. Wait for the plan to be properly initialised (i.e. status is not empty)
  4. Approve the desired install plan

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -279,50 +279,62 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult 
 	}
 
 	// create
+	desired := NewSubscription()
 	if errors.IsNotFound(err) {
 		r.logger.Info("Creating Grafana Operator Subscription")
-		err := r.k8sClient.Create(ctx, NewSubscription())
+		err := r.k8sClient.Create(ctx, desired)
 		return creationResult(err)
 	}
 
 	r.logger.Info("Updating Grafana Operator Subscription")
-	subscription.Spec = NewSubscription().Spec
-	err = r.k8sClient.Update(ctx, &subscription)
-	return updationResult(err)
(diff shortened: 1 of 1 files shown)


pkg/controllers/grafana-operator/controller.go
```diff
@@ -279,50 +279,62 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult
...
+
+		r.logger.V(8).Info("install plan", "name", installPlan.Name, "csv", csv, "approved", installPlan.Spec.Approved)
+
+		// look no further if an install plan for the desired CSV is already approved
+		if installPlan.Spec.Approved {
+			r.logger.V(8).Info("install plan already approved", "name", installPlan.Name, "csv", csv)
+			return next()
+		}
+
+		approvePlan = &installPlan
```
fpetkovski, line 325: Do we want to approve all install plans or just one?

sthaha: There should be only one plan if we wait for the plan to be properly initialized.

fpetkovski: Ok, sounds good :+1:

fpetkovski (approved):

PR #98, opened 2021-12-13
feat: upgrade grafana-operator to 4.1.0

fpetkovski:
This commit upgrades the Grafana Operator to 4.1.0 and configures the
operator to scan dashboards from all namespace. This change was
introduced in 4.0.2 [1].

https://github.com/grafana-operator/grafana-operator/pull/579

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -59,7 +59,7 @@ const (
 	subscriptionName  = "monitoring-stack-operator-grafana-operator"
 	operatorGroupName = "monitoring-stack-operator-grafana-operator"
 	grafanaName       = "monitoring-stack-operator-grafana"
-	grafanaCSV        = "grafana-operator.v4.0.1"
+	grafanaCSV        = "grafana-operator.v4.1.0"
 )
 
 type reconciler struct {
@@ -162,7 +162,7 @@ func RegisterWithManager(mgr ctrl.Manager) error {
 	go installPlanInformer.Run(nil)
 	if err := c.Watch(&source.Informer{
 		Informer: installPlanInformer,
-	}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{}); err != nil {
+	}, &handler.EnqueueRequestForObject{}, installPlanFilter{}); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/grafana-operator/controller.go
```diff
@@ -162,7 +162,7 @@ func RegisterWithManager(mgr ctrl.Manager) error {
 	go installPlanInformer.Run(nil)
 	if err := c.Watch(&source.Informer{
 		Informer: installPlanInformer,
-	}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{}); err != nil {
+	}, &handler.EnqueueRequestForObject{}, installPlanFilter{}); err != nil {
```
sthaha, line 165: good one!

pkg/controllers/grafana-operator/controller.go
```diff
@@ -286,13 +284,35 @@ func (r *reconciler) reconcileSubscription(ctx context.Context) reconcileResult
...
+			Kind:       "ClusterServiceVersion",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      subscription.Status.InstalledCSV,
+			Namespace: Namespace,
+		},
+	}
+	if err := r.k8sClient.Delete(ctx, &csv); err != nil {
+		return reconcileError(err)
+	}
```
sthaha, line 309: How about we refactor this to a `r.deleteSubscription()` ?

sthaha (approved):
nice! ..
May want to consider the suggested refactor

pkg/controllers/grafana-operator/controller.go
```diff
@@ -305,7 +325,6 @@ func (r *reconciler) approveInstallPlan(ctx context.Context) reconcileResult {
 	}
 
 	var approvePlan *v1alpha1.InstallPlan
-
 	for _, installPlan := range installPlans.Items {
```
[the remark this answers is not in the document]
fpetkovski: I am not entirely sure if that is safe because we are doing a `list`. So we could have installPlans without bundle lookups.

pkg/controllers/grafana-operator/controller.go
fpetkovski: I was thinking to extract an upgrade function instead. I'll submit a follow up PR :+1:

PR #102, opened 2021-12-14
refactor: reduce logging verbosity

fpetkovski:
The operator currently outputs messages even for operatons which are
no-op. This makes it hard to understand what the operator is doing.

This commit increases the log-level for operations which do not change
the cluster state, and only outputs a log message when a change is
actually made.

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -16,6 +16,7 @@ package grafana_operator
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	networkingv1 "k8s.io/api/networking/v1"
@@ -71,6 +72,8 @@ type reconciler struct {
 	grafanaClientset rest.Interface
 }
 
+type ReconcileFunc func(ctx context.Context) reconcileResult
+
 type reconcileResult struct {
 	ctrl.Result
 	err  error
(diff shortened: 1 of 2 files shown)


prashbnair (approved):
lgtm

pkg/controllers/grafana-operator/controller.go
```diff
@@ -170,49 +173,39 @@ func RegisterWithManager(mgr ctrl.Manager) error {
...
-
-	logger.Info("Reconciling Grafana Operator OperatorGroup")
-	if res := r.reconcileOperatorGroup(ctx); res.stop {
-		return result(res)
-	}
-
-	logger.Info("Reconciling Grafana Operator Subscription")
-	if res := r.reconcileSubscription(ctx); res.stop {
-		return result(res)
+	log.V(6).Info("Reconciling resources")
```
sthaha, line 189: I think we should log this to identify start of the reconcile loop

fpetkovski: On the one hand yes, this could be beneficial. But the way that we currently reconcile is by re-triggering the loop over and over until all resources are created or updated. We would therefore often have empty reconciliation loops which do not modify anything. This is most pronounced when the operator starts and reconciles all resources again. So we would end up with a lot of `Reconciling resources` messages which could would be hard to interpret. 

IMO the operator should emit an info log when something actually changes. Indicating what the operator is doing is for me a trace or debug level message.

pkg/controllers/grafana-operator/controller.go
```diff
@@ -114,7 +117,7 @@ func RegisterWithManager(mgr ctrl.Manager) error {
 	c, err := controller.New("grafana-operator", mgr, controller.Options{
 		MaxConcurrentReconciles: 1,
 		Reconciler:              r,
-		Log:                     logger,
+		Log:                     ctrl.Log,
```
fpetkovski, line 120: The controller name will added as a prefix by the `controller.New` function, so we don't need to create a new logger with the name of the controller.

pkg/controllers/grafana-operator/controller.go
sthaha: Agreed, it is indeed a trace log. Let's go ahead with this change.

sthaha (approved):

PR #103, opened 2021-12-14
fix: establish a watch on Grafana CRs ince the CRD is present

fpetkovski:
We currently start an informer on Grafana CRs as soon as the operator
starts. This is problematic during installations because the Grafana CRD
will only become available once the grafana operator is installed. As a
result, we quite often get flakes in the CI related to deletion and
restoring of Grafana related resources.

This commit modifies the behavior of the Grafana Operator controller to
establish a watch only when the CRDs have been installed in the cluster.
It also creates watches on all Grafana related resources in order to resolve 
the flakes happening in the CI.

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -40,13 +40,10 @@ import (
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/controllers/grafana-operator/controller.go
```diff
@@ -166,6 +160,31 @@ func RegisterWithManager(mgr ctrl.Manager) error {
...
+			if err := r.k8sClient.List(context.Background(), &datasources, client.InNamespace("default")); err != nil {
+				logger.V(6).Info("GrafanaDataSource CRD does not exist", "err", err)
+				return
+			}
+			grafanaInformer := r.grafanaInformer()
+			go grafanaInformer.Run(nil)
+			if err := c.Watch(&source.Informer{
+				Informer: grafanaInformer,
+			}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{}); err != nil {
+				logger.Error(err, "Could not establish a watch on Grafana CRs")
```
sthaha, line 177: if the watch couldn't be established wouldn't this result in just logging an error and never attempting to watch again? 

How about we add the watch dynamically in the reconcile loop? 
e.g. see: https://github.com/rhobs/monitoring-stack-operator/pull/100/files#diff-f7ca2b110a0e9c32221f61f0b87534e8cf46132d81b5ce267e5679396d969c8fR158

sthaha (approved):

sthaha:
nit: The commit message says ` Grafana CRs ince the CRD is present` - `ince` 🤔

sthaha (approved):

PR #107, opened 2021-12-15
fix: cleaning up grafana datasource on deleting monitoring stack

prashbnair:
this fix will clean up the grafana data source created by the monitoring
stack on deleting the monitoring stack

### DIFF
--- a/pkg/controllers/monitoring-stack/controller.go
+++ b/pkg/controllers/monitoring-stack/controller.go
@@ -46,11 +46,13 @@ import (
 
 	"github.com/go-logr/logr"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	goctrl "github.com/rhobs/monitoring-stack-operator/pkg/controllers/grafana-operator"
 )
 
 const (
 	grafanaDatasourceOwnerName      = "monitoring-stack-operator/owner-name"
 	grafanaDatasourceOwnerNamespace = "monitoring-stack-operator/owner-namespace"
+	finalizerName                   = "monitoring-stack-grafana-ds/finalizer"
 )
 
 type reconciler struct {
@@ -142,11 +144,14 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-
 	if requeue, err := r.createGrafanaDSWatch(ctx); err != nil {
 		return ctrl.Result{}, err
 	} else if requeue {
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	if !ms.ObjectMeta.DeletionTimestamp.IsZero() {
+		return r.cleanupResources(ctx, ms)
 	}
 
 	for _, patcher := range patchers {
@@ -164,8 +169,63 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		}
 	}
+	return r.setupFinalizer(ctx, ms)
+}
 
+func (r *reconciler) deleteGrafanaDS(ctx context.Context, ms *stack.MonitoringStack) error {
+	logger := r.logger.WithValues(stackName(ms)...)
+
+	gds := types.NamespacedName{Namespace: goctrl.Namespace, Name: GrafanaDSName(ms)}
+	grafanaDS := grafanav1alpha1.GrafanaDataSource{}
+	if err := r.k8sClient.Get(ctx, gds, &grafanaDS); err != nil {
+		// if the datasource is already deleted, take no further action
+		return client.IgnoreNotFound(err)
+	}
+
+	// grafana ds exists; so delete it
+	logger.WithValues("GrafanaDataSource", grafanaDS.Name).Info("Deleting GrafanaDataSource")
+	err := r.k8sClient.Delete(ctx, &grafanaDS)
+	return client.IgnoreNotFound(err)
+}
+
+func (r *reconciler) cleanupResources(ctx context.Context, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	logger := r.logger.WithValues(stackName(ms)...)
+
+	if !controllerutil.ContainsFinalizer(ms, finalizerName) {
+		logger.V(6).Info("Finalizer already removed")
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.deleteGrafanaDS(ctx, ms); err != nil {
+		logger.V(6).Info("Could not delete GrafanaDataSource", "err", err)
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("Removing finalizer")
+	controllerutil.RemoveFinalizer(ms, finalizerName)
+	err := r.k8sClient.Update(ctx, ms)
+	if errors.IsConflict(err) {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+	return ctrl.Result{}, err
+}
+
+func (r *reconciler) setupFinalizer(ctx context.Context, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	if controllerutil.ContainsFinalizer(ms, finalizerName) {
+		return ctrl.Result{}, nil
+	}
+	controllerutil.AddFinalizer(ms, finalizerName)
+	if err := r.k8sClient.Update(ctx, ms); err != nil {
+		if errors.IsConflict(err) {
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+		}
+		return ctrl.Result{}, err
+	}
 	return ctrl.Result{}, nil
+}
+
+func GrafanaDSName(ms *stack.MonitoringStack) string {
+	return fmt.Sprintf("ms-%s-%s", ms.Namespace, ms.Name)
 }
 
 func (r *reconciler) getStack(ctx context.Context, req ctrl.Request) (*stack.MonitoringStack, error) {
@@ -253,4 +313,8 @@ func (r *reconciler) createGrafanaDSWatch(ctx context.Context) (bool, error) {
 
 	return false, nil
 
+}
+
+func stackName(ms *stack.MonitoringStack) []interface{} {
+	return []interface{}{"Stack", ms.Namespace + "/" + ms.Name}
 }
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -8,7 +8,8 @@ import (
 	"testing"
 	"time"
 
-	grafana_operator "github.com/rhobs/monitoring-stack-operator/pkg/controllers/grafana-operator"
+	goctrl "github.com/rhobs/monitoring-stack-operator/pkg/controllers/grafana-operator"
+	msctrl "github.com/rhobs/monitoring-stack-operator/pkg/controllers/monitoring-stack"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -84,6 +85,9 @@ func TestMonitoringStackController(t *testing.T) {
 		}, {
 			name:     "Grafana data source is recreated on delete",
 			scenario: recreateDeleteGrafanaDS,
+		}, {
+			name:     "Deletion of monitoring stack cleans up grafana datasource",
+			scenario: cleanupGrafanaDS,
 		},
 	}
 
@@ -177,10 +181,10 @@ func recreateDeleteGrafanaDS(t *testing.T) {
 	assert.NilError(t, err, "failed to create a monitoring stack")
 
 	grafanaDS := grafanav1alpha1.GrafanaDataSource{}
-	f.AssertResourceEventuallyExists(datasourceName, grafana_operator.Namespace, &grafanaDS)
+	f.AssertResourceEventuallyExists(datasourceName, goctrl.Namespace, &grafanaDS)(t)
 
 	f.K8sClient.Delete(context.Background(), &grafanaDS)
-	f.AssertResourceEventuallyExists(datasourceName, grafana_operator.Namespace, &grafanaDS)
+	f.AssertResourceEventuallyExists(datasourceName, goctrl.Namespace, &grafanaDS)(t)
 }
 
 func validateStackLogLevel(t *testing.T) {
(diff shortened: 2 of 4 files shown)


pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if ms.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !containsString(ms.ObjectMeta.Finalizers, finalizerName) {
+			controllerutil.AddFinalizer(ms, finalizerName)
+			if err := r.k8sClient.Update(ctx, ms); err != nil {
+				if errors.IsConflict(err) {
+					return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+				}
+				return ctrl.Result{}, err
+			}
+		}
```
sthaha, line 182: This part is a bit confusing to understand ... Why does `cleanupResources` add a finalizer?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -46,11 +46,13 @@ import (
 
 	"github.com/go-logr/logr"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	grafana_operator "github.com/rhobs/monitoring-stack-operator/pkg/controllers/grafana-operator"
```
sthaha, line 49: it would be nice if package names didn't use `_`

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		}
 	}
+	return r.cleanupResources(ctx, ms)
+}
+
+func (r *reconciler) cleanupResources(ctx context.Context, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	if ms.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !containsString(ms.ObjectMeta.Finalizers, finalizerName) {
```
sthaha, line 174: this check shouldn't necessary since  `AddFinalizers` already does this check

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			controllerutil.AddFinalizer(ms, finalizerName)
```
sthaha, line 175: Shouldn't this be part of `creation` and updation?

pkg/controllers/monitoring-stack/controller.go
fpetkovski, line 174: There's https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#ContainsFinalizer

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err := r.k8sClient.Update(ctx, ms); err != nil {
+				if errors.IsConflict(err) {
+					return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+				}
+				return ctrl.Result{}, err
+			}
+		}
+	} else if containsString(ms.ObjectMeta.Finalizers, finalizerName) {
```
sthaha, line 183: should we check if the finalizer is set before deleting grafanadatasource?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+		if err := r.k8sClient.Get(ctx,
+			types.NamespacedName{Namespace: grafana_operator.Namespace, Name: fmt.Sprintf("ms-%s-%s", ms.Namespace, ms.Name)},
```
sthaha, line 186: I think we should have a `grafanaDSNameForStack(name, ns string) string` that returns the grafana-datasource name given a monitoring-stack's name and namespace so that this can be reused everywhere

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			grafanaDS); err != nil {
+			return ctrl.Result{}, err
+		}
 
+		if err := r.k8sClient.Delete(ctx, grafanaDS); err != nil {
```
sthaha, line 191: I think we should ignore `errors.IsNotExists()` when deleting

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +358,32 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+func cleanupGrafanaDS(t *testing.T) {
+	ms := newMonitoringStack(t, "test-cleanup-grafana-ds")
+	datasourceName := fmt.Sprintf("ms-%s-%s", ms.Namespace, ms.Name)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+
+	f.GetResourceWithRetry(t, datasourceName, "monitoring-stack-operator", grafanaDS)
```
sthaha, line 370: `f.AssertResourceExists` instead of `Get` and `Assert`

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +358,32 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+
+	assert.NilError(t, err, "grafana data source is not created")
+
+	f.K8sClient.Delete(context.Background(), ms)
+
+	if err = wait.Poll(5*time.Second, 20*time.Second, func() (done bool, err error) {
```
sthaha, line 376: can't we use `AssertResourceNeverExists` here?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,7 +166,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		}
 	}
+	return r.cleanupResources(ctx, ms)
```
sthaha, line 169: wouldn't this run all patchers unnecessarily when a monitoring stack is deleted? 
Shouldn't we handle deletion way early on in the reconcile loop .. i.e as soon as get the `ms` - MonitoringStack ?

pkg/controllers/monitoring-stack/controller.go
prashbnair: It would have been set in the `if` condition, this is just to confirm.

pkg/controllers/monitoring-stack/controller.go
prashbnair: Do you want to ignore the error when deleting, I am not sure about that.

pkg/controllers/monitoring-stack/controller.go
prashbnair: If we move it to the top, when the reconcile loop runs for the first time, the stack is not completely initialised. 
Shouldn't we be adding the finalizer only after all components have been created?

pkg/controllers/monitoring-stack/controller.go
prashbnair: We can change it but then we will need to add a check that `DeletionTimestamp` is not zero before we can actually delete the resource. That is what tells us that we are in delete mode.
I feel it better this way as the logic is clear.

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

test/e2e/monitoring_stack_controller_test.go
prashbnair: Done

test/e2e/monitoring_stack_controller_test.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
fpetkovski: I think the idea is to run this code at the top when the deletion timestamp is set. So the algorithm would be:
* If deletion timestamp is set, delete datasource, remove finalizer and update stack
* else If finalizer not set, set finalizer and update stack
* else reconcile child resources

pkg/controllers/monitoring-stack/controller.go
fpetkovski: I think this error is safe to ignore. Otherwise, we can end up in an endless loop if the child resource does not exist because errors are retried.

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
prashbnair: Done

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,10 +169,65 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	controllerutil.RemoveFinalizer(ms, finalizerName)
+	err := r.k8sClient.Update(ctx, ms)
+	if errors.IsConflict(err) {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+	return ctrl.Result{}, err
+}
+
+func (r *reconciler) setupFinalizer(ctx context.Context, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	if !controllerutil.ContainsFinalizer(ms, finalizerName) {
```
sthaha, line 214:
```suggestion
	if controllerutil.ContainsFinalizer(ms, finalizerName) {
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +359,24 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+	datasourceName := msctrl.GrafanaDSName(ms)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+
+	f.AssertResourceEventuallyExists(datasourceName, goctrl.Namespace, grafanaDS)
+
+	assert.NilError(t, err, "grafana data source is not created")
```
sthaha, line 373: is this true?
```suggestion
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +359,24 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+
+	f.K8sClient.Delete(context.Background(), ms)
+
```
sthaha, line 376:
```suggestion
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +359,24 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+	ms := newMonitoringStack(t, "test-cleanup-grafana-ds")
+	datasourceName := msctrl.GrafanaDSName(ms)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+
+	f.AssertResourceEventuallyExists(datasourceName, goctrl.Namespace, grafanaDS)
+
```
sthaha, line 372:
```suggestion
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +359,24 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
 
+func cleanupGrafanaDS(t *testing.T) {
+	ms := newMonitoringStack(t, "test-cleanup-grafana-ds")
+	datasourceName := msctrl.GrafanaDSName(ms)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+
```
sthaha, line 370:
```suggestion
```

sthaha (changes requested):
Almost good to go.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -164,10 +169,65 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	logger.Info("Removing finalizer")
+	controllerutil.RemoveFinalizer(ms, finalizerName)
+	err := r.k8sClient.Update(ctx, ms)
+	if errors.IsConflict(err) {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+	return ctrl.Result{}, err
+}
+
+func (r *reconciler) setupFinalizer(ctx context.Context, ms *stack.MonitoringStack) (ctrl.Result, error) {
```
sthaha, line 213: @prashbnair I think it is worth having  a test that validates that the finalizer is set on a monitoring stack 

Note ... This can be done is a follow up PR as well.

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -355,6 +359,20 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+func cleanupGrafanaDS(t *testing.T) {
+	ms := newMonitoringStack(t, "test-cleanup-grafana-ds")
+	datasourceName := msctrl.GrafanaDSName(ms)
+
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	grafanaDS := &grafanav1alpha1.GrafanaDataSource{}
+	f.AssertResourceEventuallyExists(datasourceName, goctrl.Namespace, grafanaDS)(t)
+
```
sthaha, line 371: We can add another validation here that the finalizer is set on the `ms` that was created above

sthaha (approved):
@prashbnair I am approving so that we can merge this PR that has been waiting for a longtime. However, I do think we need to validate that the finalizer is set.

PR #108, opened 2021-12-20
fix: optimise watches used in grafana operator

sthaha:
Previously grafana-operator used `source.Kind` to watch the
resources which was far less efficient than the singleResourceInformer.

This patch makes use of `source.NewKindWithCache` which allows building
a cache/informers that watches only the `monitoring-stack-operator`
namespace and only those resources that has the label
`app.kubernetes.io/managed-by=monitoring-stack-operator`

NOTE: `msoClient` is created to make use of the cache.

### DIFF
--- a/pkg/controllers/grafana-operator/controller.go
+++ b/pkg/controllers/grafana-operator/controller.go
@@ -15,26 +15,19 @@ package grafana_operator
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 	"time"
 
 	networkingv1 "k8s.io/api/networking/v1"
 
-	"github.com/rhobs/monitoring-stack-operator/pkg/eventsource"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	integreatlyv1alpha1 "github.com/grafana-operator/grafana-operator/v4/api/integreatly/v1alpha1"
+	"github.com/rhobs/monitoring-stack-operator/pkg/eventsource"
 	v1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/grafana-operator/controller.go
```diff
@@ -58,17 +56,18 @@ const (
...
+	managedBy    = "app.kubernetes.io/managed-by"
+	operatorName = "monitoring-stack-operator"
 )
 
 type reconciler struct {
 	controller              controller.Controller
 	grafanaWatchEstablished bool
-	k8sClient               client.Client
+	cache                   cache.Cache
+	msoClient               client.Client
```
fpetkovski, line 68: Can we keep this as `k8sClient`?

sthaha: I think we should differentiate so that the `msoClient` means a client that makes use of the `cache`  with informers watching only the `monitoring-stack-operator` namespace as opposed to `k8sClient` that we use else where which has the `mgr.GetCache()` attached/associated.

fpetkovski: Hm, should we call it `nsClient` then? I suspect the namespace will change at one point, and mso is a bit ambiguous

sthaha: agreed! should we delete the k8sClient .. which is now unused ?

sthaha: actually .. I delete it already :D

fpetkovski (approved):
Nice!

PR #110, opened 2022-01-17
Multi ns support

fpetkovski:


sthaha:
closing this for now. lets revisit this when there is a requirement.

PR #111, opened 2022-01-21
Changes to enable persistent storage for prometheus

prashbnair:


prashbnair:
@fpetkovski @sthaha Can you take a look and let me know if this how we want to proceed?

pkg/apis/v1alpha1/types.go
```diff
@@ -76,6 +80,23 @@ type MonitoringStackStatus struct {
 	// ??
 }
 
+type PrometheusConfig struct {
+	// Define persistent volume claim for prometheus
+	// +optional
+	PersistentVolumeClaim *PersistentVolumeClaim `json:"persistentVolumeClaim,omitempty"`
```
fpetkovski, line 86: I would use a `corev1.PersistentVolumeClaimSpec` directly here. As it is now, it's not possible for users to select volume claims by labels, which is a common use case for statefulsets

prashbnair: Done

fpetkovski (approved):

PR #115, opened 2022-04-05
fix: Increase resource(memory) limit of mso operator

Jooho:
jira: https://issues.redhat.com/browse/MON-2353

Signed-off-by: jooho <jlee@redhat.com>


sthaha (approved):
Thanks for the patch @Jooho

PR #117, opened 2022-04-05
ci: use Makefile for golangci-lint check

jan--f:
Without this we potentially run local checks and ghub action check
with two different versions.

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -197,6 +197,10 @@ tools: $(CONTROLLER_GEN) \
 		echo  $$(basename $(GOJSONTOYAML)) $(GOJSONTOYAML_VERSRION) >> $$tools_file ;\
 	}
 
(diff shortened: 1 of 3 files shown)


simonpasquier (approved):
lgtm

simonpasquier (approved):

Makefile.tools
```diff
@@ -20,7 +20,7 @@ OPM=$(TOOLS_DIR)/opm
 OPM_VERSION = v1.15.1
 
 GOLANGCI_LINT=$(TOOLS_DIR)/golangci-lint
-GOLANGCI_LINT_VERSION = v1.42.1
+GOLANGCI_LINT_VERSION = v1.45.2
```
sthaha, line 23: it seems like we have missed adding the tool to the cache file - ` .github/tools`
I will submit a patch.

PR #118, opened 2022-04-06
ci: add golangci-lint to the list of tools in cache

sthaha:


jan--f (commented):
:+1: 
I have two questions though: 
The `tools` target doesn't seem to install all tools (like `golangci-lint`)?
The `validate-tools` seems to only validate a subset of the tools. Is that intentional?

sthaha:
> 👍 I have two questions though: The `tools` target doesn't seem to install all tools (like `golangci-lint`)? The `validate-tools` seems to only validate a subset of the tools. Is that intentional?

Good catch! it isn't added to `tools` and `validate-tools` target

Makefile.tools
```diff
@@ -175,9 +175,10 @@ tools: $(CONTROLLER_GEN) \
 		$(OC) \
 		$(OPERATOR_SDK) \
 		$(OPM) \
-	 	$(PROMQ) \
-	 	$(CRDOC) \
-	 	jsonnet-tools
+		$(PROMQ) \
```
sthaha, line 178: this removes the `\t<space><space>` with just `\t`

jan--f (approved):

PR #121, opened 2022-04-07
refactor: use server side apply in mso controller

jan--f:
Server side apply has been in beta since kubernetes 1.16 and stable
since 1.22. It greatly reduces client complexity, since no
get-patch-update cycles are needed anymore.
Additionally server side apply allows for controllers to handle user
changes on managed resources much more easily, if that need arises in
the future.
This also adds a `GenerationChangedPredicate` to the GrafanaDataSource watch.

### DIFF
--- a/deploy/operator/monitoring-stack-operator-cluster-role.yaml
+++ b/deploy/operator/monitoring-stack-operator-cluster-role.yaml
@@ -26,6 +26,7 @@ rules:
   - create
   - delete
   - list
+  - patch
   - update
   - watch
 - apiGroups:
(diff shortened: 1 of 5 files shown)


jan--f:
Given that `LabelSelector` is marked as `structType=atomic` the e2e test shouldn't fail like this. Still investigating.

sthaha:
@jan--f , thanks, we have been wanting to do this !
could you please test that modifications outside the MSO to any of the resources created are reverted as well? Perhaps we need a test for that if one doesn't exist already.

jan--f:
I think the failing test is due to https://github.com/kubernetes/kubernetes/issues/109668

sthaha (approved):
super!! ❤️

PR #124, opened 2022-04-20
refactor!: removes the grafana operator

JoaoBraveCoding:
Problem some users have deciced they do not want to run the grafana
operator with cost or resources both being pointed as reasons
Issue: https://issues.redhat.com/browse/MON-2247


JoaoBraveCoding:
I'm currently blocked on two aspects:

**How to expose this configuration:** 
In an initial discussion with @sthaha I suggested exposing a knob on the OLM installation, however after researching this possibility I discovered [this issue](https://github.com/operator-framework/operator-lifecycle-manager/issues/2496) where an OLM engineer suggested two options, either the user modifies the deployment after OLM installs the operator (not ideal as I'm not sure this persists between upgrades of the operator) or the operator is configurable through a CRD. However IIUC MSO does not have such a CRD, with the closest candidate being `MonitoringStack` however the grafana operator is deployed before this CRD is reconciled.

Ideal scenario from my POV would be for the grafana operator to be configurable through the `MonitoringStack` CRD but this might have bigger implications

Possible scenario, we make it so MSO is configurable using ENV vars, this is supported by OLM as users can pass extra ENV vars through the Subscription resource  

**E2E tests:** 
If we are not able to dynamically change the configuration of the root operator it's hard to test the non-default side of this flag 

@simonpasquier @jan--f @sthaha input would be much appreciated :)

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -194,9 +199,11 @@ func (r *reconciler) cleanupResources(ctx context.Context, ms *stack.MonitoringS
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.deleteGrafanaDS(ctx, ms); err != nil {
-		logger.V(6).Info("Could not delete GrafanaDataSource", "err", err)
-		return ctrl.Result{}, err
+	if r.enableGrafanaOperator {
```
sthaha, line 202: Say we run operator with GO enabled and then re-run it after disabling it, shouldn't we delete existing GrafanaDataSources?

pkg/operator/operator.go
```diff
@@ -32,12 +32,14 @@ func New(metricsAddr string) (*Operator, error) {
...
 	}
 
-	if err := stackctrl.RegisterWithManager(mgr, stackctrl.Options{InstanceSelector: instanceSelector}); err != nil {
+	if err := stackctrl.RegisterWithManager(mgr, stackctrl.Options{InstanceSelector: instanceSelector, EnableGrafanaOperator: enableGrafanaOperator}); err != nil {
 		return nil, fmt.Errorf("unable to register monitoring stack controller: %w", err)
 	}
 
-	if err := goctrl.RegisterWithManager(mgr); err != nil {
-		return nil, fmt.Errorf("unable to register the grafana operator controller with the manager: %w", err)
+	if enableGrafanaOperator {
```
sthaha, line 39: if granfana-operator isn't enabled, IMO, we should delete any existing deployment, which IIUC, this wouldn't, thus the operator could have a dangling grafana deployed.

sthaha:
>Ideal scenario from my POV would be for the grafana operator to be configurable through the MonitoringStack CRD but this might have bigger implications

There will only be one grafana-operator deployement/subscription that MSO manages while there can be multiple Monitoring Stacks, thus IMHO, we shouldn't relate subcription of grafana operator to Monitoring Stack. It does make sense to provide an option in stack CRD to generate a grafana-datasource but that is a different discussion.


>Possible scenario, we make it so MSO is configurable using ENV vars, this is supported by OLM as users can pass extra ENV vars through the [Subscription resource](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md)

I think we should explore using ENV for configuring `GRAFANA_OPERATOR_NS` which by default can be `monitoring-stack-operator`


> E2E tests:
> If we are not able to dynamically change the configuration of the root operator it's hard to test the non-default side of this flag

Good point! 

One naive idea is to add another CI job that enables GrafanaOperator + a non-default namespace and run tests with `-run Grafana` while the default will `-run [^Grafana]` 
We could also redeploy operator (enabling grafana) after the initial test. 


Another option I think is to add a CRD to configure the Operator's behaviour - MonitoringStackOperatorConfig which allows operator's features to be toggled at runtime. This is akin to ClusterMonitoringOperator 's configmap except that it isn't a configmap, rather a CRD so that we can validate and report status of operators (PO, GO) back to the user.

JoaoBraveCoding:
Thanks a lot for the reply!

> There will only be one grafana-operator deployement/subscription that MSO manages while there can be multiple Monitoring Stacks, thus IMHO, we shouldn't relate subcription of grafana operator to Monitoring Stack. It does make sense to provide an option in stack CRD to generate a grafana-datasource but that is a different discussion.

Good point I completely overlooked that point! 100% agree. Indeed we could add that knob to the Monitoring Stack CRD however do you think we should do it in this PR or wait for the request and do it in another?

> I think we should explore using ENV for configuring GRAFANA_OPERATOR_NS which by default can be monitoring-stack-operator

Given the other point regarding the e2e tests I would be more inclined to introduce the MonitoringStackOperatorConfig CRD, having worked a lot with the the [Crane operator](https://github.com/konveyor/mig-operator) which is configured in this manner this seems to me like the best option to expose configuration. Since it would not only make testing easier/more maintainable but in my opinion it would also make the UX more consistent/enjoyable.

jan--f:
A cli flag that we can pass via ENV makes sense to me. I'm not exactly sure how configurable the grafana-operator installation via OLM is. Does it at some point make sense to reconsider if we want MSO to handle installing grafana-operator? If we want to keep doing this we should probably consider how we handle customization of this installation.
How would a config CRD for MSO work? We only ever have one operator, but having a CRD implies multiple resources can exist.

The CRD addition to trigger the datasource generation is a good idea. This can be done in a follow-up PR, but I think we should proactively add it. Decoupling MSO behavior from whether we install grafana-operator or not makes sense to me.

JoaoBraveCoding:
> A cli flag that we can pass via ENV makes sense to me. I'm not exactly sure how configurable the grafana-operator installation via OLM is. Does it at some point make sense to reconsider if we want MSO to handle installing grafana-operator? If we want to keep doing this we should probably consider how we handle customization of this installation. 

So regarding the customization of installation using a CRD vs ENV vars, my opinion is that CRD is better for two reasons:
1. Better UX  with some operator like the [Crane operator](https://github.com/konveyor/mig-operator) exposing configuration also thorough a CRD  as shown [here](https://github.com/konveyor/mig-operator#customizing-your-installation). Another popular example is the [TektonCD Operator](https://github.com/tektoncd/operator/blob/main/docs/TektonConfig.md)
2. Ease of tests, with the configuration being provided by a CRD like in ConfigMap of the CMO project we can change the configuration of the operator on the fly VS having to redeploy it. 

> How would a config CRD for MSO work? We only ever have one operator, but having a CRD implies multiple resources can exist.

When we create a CRD we can actually set the [scope of such a CRD](https://sdk.operatorframework.io/docs/building-operators/golang/crds-scope/#set-scope-marker-in-typesgo), this makes it so there is only on instance of such a CRD in the cluster. Furthermore the operator would then always look for the "Config" CRD always in the namespace where it was deployed.

> The CRD addition to trigger the datasource generation is a good idea. This can be done in a follow-up PR, but I think we should proactively add it. Decoupling MSO behavior from whether we install grafana-operator or not makes sense to me.

Yes, I 100% agree! So since it makes sense to add that knob to the MonitoringStack CRD I will try to squeeze that  feature in!

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,54 @@ type ThanosQuerierSpec struct {
 // It should always be reconstructable from the state of the cluster and/or outside world.
 type ThanosQuerierStatus struct {
 }
+
+// MonitoringStackConfig configures the deployment of the different operators managed
```
JoaoBraveCoding, line 158: @sthaha @jan--f Here's a preliminary preview of the knobs exposed by the new config CRD please let me know what do you think

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,54 @@ type ThanosQuerierSpec struct {
...
+}
+
+// MonitoringStackOperator exposes configuration options regarding the monitoring stack operator.
+// +kubebuilder:resource
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object 
+type MonitoringStackOperator struct {
+	// GrafanaDataSource exposes whether or not the monitoring stack should create 
+	// GrafanaDataSources or not.
+	// +kubebuilder:validation:Optional
+	GrafanaDataSource bool `json:"grafanaDataSource,omitempty"`
```
sthaha, line 193: 1. For the time being I think we should create grafana datasources if the `GrafanaOperator` is enabled. 

2. Creation of grafana-datasource that is linked to a monitoring-stack should be controlled through a stack and not as global setting (IMHO)

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,54 @@ type ThanosQuerierSpec struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   MonitoringStackConfigSpec   `json:"spec,omitempty"`
+	Status MonitoringStackConfigStatus `json:"status,omitempty"`
+}
+
+// MonitoringStackConfigSpec defines the desired state of MonitoringStackConfig.
+type MonitoringStackConfigSpec struct {
+	// +kubebuilder:validation:Optional
+	MonitoringStackOperator MonitoringStackOperator `json:"monitoringStackOperator,omitempty"`
```
sthaha, line 176: We shouldn't need this config (yet)

jan--f:
> 
> When we create a CRD we can actually set the [scope of such a CRD](https://sdk.operatorframework.io/docs/building-operators/golang/crds-scope/#set-scope-marker-in-typesgo), this makes it so there is only on instance of such a CRD in the cluster. Furthermore the operator would then always look for the "Config" CRD always in the namespace where it was deployed.

Ah thanks for the detailed explanation. TIL, that does sound a lot better then messing around with ENV. Lets go with the CRD approach is no one objects. I'll give this a proper review next week.

> Makes sense to me. If this isn't an immediate requirement, we should think of removing the entire controller and let user manage the deployment of grafana.

I think there are arguments to be made against installing the grafana operator. Lets gather more input on this requirement and then consider whether it makes sense to keep maintaining this code.

JoaoBraveCoding:
OK so after Jan's comment I started to try and implementing a first version of this, however I'm starting to think that I might have bitten more than what I can chew... so I would like to run something by you first.

My main concern is with the way I usually see operators implemented. Everything starts in `main.go` where we will eventually create a `manager` that is responsible for running controllers. Said controllers will register against said manager (in this case grafana, monitoringStack and thanos-querier) and then eventually `msg.Start(ctx)` will be called, to start the controllers.

Now, with what I know, I was imagining that the MSOConfig controller (controller responsible for reconciling the MSOConfig CRD) would have a reconcile loop where we would either enable or disable grafana. Given how MSO works nowadays, let's consider the case where the grafana was enabled and now we wanted to disable it. Then we would have to "de-register" the grafana controller from the manager, right? But IIUC this is not an option...

If all of the above is true, I was then thinking that for this to work we the MSOConfig controller would have to maintain two managers, one specifically for grafana and another for monitoring stack. Does this make sense or did I get lost somewhere?

sthaha:
> Now, with what I know, I was imagining that the MSOConfig controller (controller responsible for reconciling the MSOConfig CRD) would have a reconcile loop where we would either enable or disable grafana. Given how MSO works nowadays, let's consider the case where the grafana was enabled and now we wanted to disable it. Then we would have to "de-register" the grafana controller from the manager, right? But IIUC this is not an option. 

We wouldn't have to unregister the controller, rather the controller will need to undo any actions it had performed earlier (unsubscribe, delete datasources, etc ..)

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,41 @@ type ThanosQuerierSpec struct {
 // It should always be reconstructable from the state of the cluster and/or outside world.
 type ThanosQuerierStatus struct {
 }
+
+// MonitoringStackConfig configures the deployment of the different operators managed
+// by the Monitoring Stack Operator
+// +k8s:openapi-gen=true
+// +kubebuilder:resource
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
```
sthaha, line 163: could you please shed some light into why we want to make this cluster scoped?

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,41 @@ type ThanosQuerierSpec struct {
...
+
+// GrafanaOperator exposes configuration options regarding the grafana operator
+// +kubebuilder:resource
+type GrafanaOperator struct {
+	// Enable exposes whether or not the grafana operator should be deployed.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=false
+	Enable bool `json:"enable,omitempty"`
+	// Namespace exposes to which namespace the grafana operator should be deployed to.
+	// +kubebuilder:validation:Optional
```
sthaha, line 192: How about `//+kubebuilder:default:="monitoring-stack-operator"`

pkg/apis/v1alpha1/types.go
```diff
@@ -154,3 +154,41 @@ type ThanosQuerierSpec struct {
...
+// MonitoringStackConfigSpec defines the desired state of MonitoringStackConfig.
+type MonitoringStackConfigSpec struct {
+	// +kubebuilder:validation:Optional
+	GrafanaOperator *GrafanaOperator `json:"grafanaOperator,omitempty"`
+}
+
+// MonitoringStackConfigStatus defines the observed state of MonitoringStackConfig.
+// It should always be reconstructable from the state of the cluster and/or outside world.
+type MonitoringStackConfigStatus struct {
+}
```
sthaha, line 182: I think we can add a section for grafana operator installation status.

JoaoBraveCoding:
> Given that configuration of the operator is done through a config CR (and not using --enable-feature-flag), I don't think we need a separate controller to monitor the config cr rather each controller can watch the config and react depending on the state of the config, update the status of the config.

YES! This is a great idea! Way simpler than my original idea and way easier to implement! With this idea in mind, I feel that my other concerns do not make sense. To be seen in the meeting.

deploy/crds/kubernetes/grafanadatasource.yaml
```diff
@@ -0,0 +1,317 @@
+apiVersion: apiextensions.k8s.io/v1
```
sthaha, line 1: We shouldn't need this and our operator shouldn't own GrafanaDatasource CRD or create a grafana datasource.

deploy/dependencies/prometheus-operator-cluster-role-binding.yaml
```diff
@@ -10,4 +10,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: monitoring-stack-operator-prometheus-operator-sa
-    namespace: default
+    namespace: operators
```
sthaha, line 13: why is this change needed 🤔? 
Should this be a separate pr unrelated to deletion of Grafana?

deploy/operator/monitoring-stack-operator-cluster-role.yaml
```diff
@@ -16,14 +16,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
```
sthaha, line 19: I would guess the operator still needs to watch and list the namespaces.

go.mod
```diff
@@ -58,6 +56,8 @@ require (
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/openshift/api v3.9.0+incompatible // indirect
 	github.com/opentracing/opentracing-go v1.2.0 // indirect
+	github.com/operator-framework/api v0.10.3 // indirect
```
sthaha, line 59: we shouldn't need this as well - right?

pkg/controllers/monitoring-stack/components.go
```diff
@@ -359,7 +358,7 @@ func newGrafanaDataSource(ms *stack.MonitoringStack) *grafanav1alpha1.GrafanaDat
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        datasourceName,
-			Namespace:   goctrl.Namespace,
+			Namespace:   ms.Namespace,
```
sthaha, line 361: IMHO, we can remove creation of grafana datasource.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -173,7 +172,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (r *reconciler) deleteGrafanaDS(ctx context.Context, ms *stack.MonitoringStack) error {
 	logger := r.logger.WithValues(stackName(ms)...)
 
-	gds := types.NamespacedName{Namespace: goctrl.Namespace, Name: GrafanaDSName(ms)}
+	gds := types.NamespacedName{Namespace: ms.Namespace, Name: GrafanaDSName(ms)}
```
sthaha, line 175: I think the operator shouldn't create grafana datasources. Also not the grafana data source need to be created in the same namespace as the grafana operator, so it best to let users using Grafana Operartor deal with that.

deploy/operator/monitoring-stack-operator-cluster-role.yaml
JoaoBraveCoding: On this I'm actually not sure... since the operator will only have to manage the CRDs no? But maybe I'm missing some context

deploy/dependencies/prometheus-operator-cluster-role-binding.yaml
JoaoBraveCoding: Indeed this can come in a separate PR

go.mod
JoaoBraveCoding: Indeed, I ran `go mod tidy`

deploy/operator/monitoring-stack-operator-cluster-role.yaml
sthaha: You are right! I blame it on lack of coffee :D

pkg/operator/scheme.go
```diff
@@ -20,7 +19,6 @@ func NewScheme() *runtime.Scheme {
 	utilruntime.Must(apiextensionsv1.AddToScheme(scheme))
 	utilruntime.Must(monitoringv1.AddToScheme(scheme))
 	utilruntime.Must(integreatlyv1alpha1.AddToScheme(scheme))
```
sthaha, line 21: can't we remove this as well?

go.mod
```diff
@@ -5,8 +5,6 @@ go 1.17
 require (
 	github.com/go-logr/logr v1.2.0
 	github.com/grafana-operator/grafana-operator/v4 v4.0.1
-	github.com/operator-framework/api v0.10.3
```
sthaha, line 8: I would expected to see, grafana-operator removed as well 🤔 . Do you know where it is being referred?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -170,22 +155,6 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return r.setupFinalizer(ctx, ms)
```
sthaha, line 155: Do we need finalizer anymore?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -46,13 +42,10 @@ import (
...
 	"github.com/go-logr/logr"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
-	goctrl "github.com/rhobs/monitoring-stack-operator/pkg/controllers/grafana-operator"
 )
 
 const (
-	grafanaDatasourceOwnerName      = "monitoring-stack-operator/owner-name"
-	grafanaDatasourceOwnerNamespace = "monitoring-stack-operator/owner-namespace"
-	finalizerName                   = "monitoring-stack-grafana-ds/finalizer"
+	finalizerName = "monitoring-stack-grafana-ds/finalizer"
```
sthaha, line 48: we shouldn't need the finalizer that was set to keep the stack object until grafana-datasource is deleted.

JoaoBraveCoding: Removed finalizer as currently would not be used for anything.

deploy/crds/kubernetes/kustomization.yaml
```diff
@@ -8,4 +8,4 @@ resources:
   - prometheuses.yaml
   - prometheusrules.yaml
   - servicemonitors.yaml
-  - thanosrulers.yaml
+  - thanosrulers.yaml
```
sthaha, line 11: since the PR needs to be resubmitted, could we remove this unrelated change.

README.md
```diff
@@ -69,36 +69,35 @@ to the controller-gen CLI page in the [kubebuilder documentation](https://book.k
   ```sh
   kind create cluster --image kindest/node:v1.22.4
   ```
-  &nbsp; For more advanced tests, a 3 node cluster may be more suitable.
+  &nbsp; To run the e2e tests, a 3 node cluster is necessary (due to anti-affinity).
```
sthaha, line 72: IMHO, PR should contain only changes related to the feature. I would be nice to if this doc improvement is a new PR. 
Imagine git revert <this-commit> to get Grafana back. It would result in us losing this doc improvement which isn't what we really want.

sthaha:
@JoaoBraveCoding , nice work. Very closing to merging. It would be great if you considered those comments as well 🙇

README.md
JoaoBraveCoding: Sure makes perfect sense :) 
https://github.com/rhobs/monitoring-stack-operator/pull/135

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -145,14 +131,10 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if requeue, err := r.createGrafanaDSWatch(ctx); err != nil {
-		return ctrl.Result{}, err
-	} else if requeue {
-		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
-	}
 
 	if !ms.ObjectMeta.DeletionTimestamp.IsZero() {
```
sthaha, line 135: Do you see this happen? IIUC, this path shouldn't be hit since we don't have finalizers set anymore.

sthaha (approved):
good stuff, Joao!

PR #132, opened 2022-05-05
feat: enable Prometheus HA mode by default

sthaha:
This patch deploys prometheus in HA mode by ...
  1. Defaulting to 2 prometheus pods for each monitoring-stack.
  2. Setting Pod Anti Affinity to run both pods on different nodes.
  3. Adds a pod disruption budget with minAvailability set to 1
  
The number of replica to run is configurable through
monitoring-stack.Spec.prometheusConfig.Replicas

Fixes: https://issues.redhat.com/browse/MON-2105


sthaha (commented):
@simonpasquier @jan--f , could you please take a look at this implementation for running prometheus in HA mode. 

I will add a test to ensure that at least 2 pods are run by default.

dummy
```diff
@@ -0,0 +1 @@
+ha-prom - 1
```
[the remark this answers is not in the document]
sthaha: oh! yes!!! wonder how I missed this.

pkg/controllers/monitoring-stack/components.go
```diff
@@ -415,6 +423,25 @@ func newServiceAccount(name string, namespace string) *corev1.ServiceAccount {
...
 
+func applyDefaultConfig(src *stack.PrometheusConfig) *stack.PrometheusConfig {
+	if src == nil {
+		return &stack.PrometheusConfig{
+			// deploy 2 replica by default
+			Replicas: int32Ptr(2),
+		}
+	}
+
+	ret := &stack.PrometheusConfig{
```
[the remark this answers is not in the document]
sthaha: you are right I thought of using mergo https://github.com/imdario/mergo. 
The intention here is to not modify the src. hmm may be I can DeepCopy() :)

sthaha (commented):
Thank you for the review comments @JoaoBraveCoding

JoaoBraveCoding (dismissed):
LGTM, but I'll either @simonpasquier or @jan--f also review it as I might've missed something

pkg/controllers/monitoring-stack/components.go
```diff
@@ -693,6 +770,32 @@ func newAdditionalScrapeConfigsSecret(ms *stack.MonitoringStack, name string) *c
 	}
 }
 
+func newPrometheusPDB(ms *stack.MonitoringStack, instanceSelectorKey string, instanceSelectorValue string) *policyv1.PodDisruptionBudget {
```
sthaha, line 773: I need to test if this could cause issue if user sets replicas to 1.

pkg/controllers/monitoring-stack/components.go
```diff
@@ -475,6 +521,31 @@ func newPrometheus(
...
+							PodAffinityTerm: corev1.PodAffinityTerm{
+								LabelSelector: &metav1.LabelSelector{
+									MatchLabels: podLabels("prometheus", ms.Name),
+								},
+								TopologyKey: "topology.kubernetes.io/zone",
+							},
+						},
+					},
+				},
+			},
```
jan--f, line 548: Do we want to add this if its not necessarily applicable?
Once we move to SSA we could leave this to the users to add on their own.

pkg/controllers/monitoring-stack/components.go
```diff
@@ -427,11 +475,7 @@ func newPrometheus(
 		prometheusSelector = &metav1.LabelSelector{}
 	}
 
-	// create a default config if user hasn't specified one
-	config := ms.Spec.PrometheusConfig
-	if config == nil {
-		config = &stack.PrometheusConfig{}
-	}
+	config := applyDefaultConfig(ms.Spec.PrometheusConfig)
```
jan--f, line 478: Do we need to apply the default explicitly here? Doesn't the default in the CRD manifest take care of that?

jan--f (commented):
Just two questions, but overall this looks good.

pkg/controllers/monitoring-stack/components.go
sthaha: Wouldn't we want to ensure that Prometheus is run on different nodes so that one can be drained for upgrade while at least one prometheus is running? 
IMHO we should keep this until SSA is ready and remove optional customisations in a different cleanup PR.

pkg/controllers/monitoring-stack/components.go
sthaha: I think it can if we change `PrometheusConfig *PrometheusConfig` in MonitoringStackSpec to `PrometheusConfig PrometheusConfig` since it  `defaults` to `nil` at the moment.  

I don't have a strong opinion about the implementation. What do you feel?

sthaha (commented):
@jan--f , Thanks for the review. 

I have removed the applyDefaults, used kubebuilder:defaults (which I got working this time) and added a e2e test to ensure that the defaults are applied. 

Could you please take a look now?

pkg/controllers/monitoring-stack/components.go
sthaha: Good catch! I have modified the spec to apply defaults. My previous attempt didn't work since I had to be explicit at 2 different places - `config`, and `replicas`

jan--f (dismissed):
Thanks, :+1:

simonpasquier:
I had the impression that with only one replica, you shouldn't configure pod disruption budget (at least this is what we do in CMO) otherwise Kube would never be able to reschedule the pod in case of node drain.

sthaha:
> I had the impression that with only one replica, you shouldn't configure pod disruption budget (at least this is what we do in CMO) otherwise Kube would never be able to reschedule the pod in case of node drain.

Thats the impression I had as well. Except when I tested, it worked.  Could it be because no PVC was configured? 
To be on the safe-side, I will set the minAvailable to 0 if the replicas = 1.
@simonpasquier WDYT?

sthaha:
@simonpasquier @jan--f , I have modified the controller logic to remove existing resources (e.g. pdb) if the `patcher.Patch` returns `nil`.  

This allows users to update `replicas`  monitoring stack and pdb will be automatically cleaned up / created.

jan--f (approved):
lgtm :+1:

simonpasquier (commented):
late to the party but lgtm from my side too :)

sthaha:
Thanks @jan--f ,  @simonpasquier 🙇

PR #133, opened 2022-05-10
ci: publish versioned catalog images

sthaha:
Previously, catalog images only used to have :latest image tags. Since
Quay images without tags get garbage collected, it wasn't possible to
pin down a catalog image by its sha256.

This PR, creates a catalog image which has the same version as the
bundle and the operator image. Additionally, it also tags this
versioned image as `latest` so that it is possible to make continuous
deployments (useful for development channel) should the users choose it.

Fixes: https://issues.redhat.com/browse/MON-2498

### DIFF
--- a/.github/olm-publish/action.yaml
+++ b/.github/olm-publish/action.yaml
@@ -51,7 +51,7 @@ runs:
 
   - name: Build Catalog Image
     shell: bash
-    run: make catalog-image FIRST_OLM_RELEASE=false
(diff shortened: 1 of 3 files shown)


JoaoBraveCoding (approved):
LGTM!

sthaha:
@jan--f @simonpasquier are we okay with merging this?

jan--f (approved):
lgtm

PR #134, opened 2022-05-13
style: switch to human readable time format in logs

tremes:
Format log timestamp in the RFC3339. The motivation here is to have human readable time format.
E.g.  The log timestamp is  `1.6523301588633049e+09` is now formatted as `2022-05-13T13:23:10Z`

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -21,6 +21,7 @@ import (
 	"os"
 
 	"github.com/rhobs/monitoring-stack-operator/pkg/operator"
+	"go.uber.org/zap/zapcore"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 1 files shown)


jan--f (changes requested):
Can you please give a before/after example of the resulting timestamps in the commit message? I imagine that motivates the change sufficiently...?

sthaha (dismissed):

sthaha:
Prints 
```
2022-05-13T13:23:10Z	INFO	controller.monitoringstack	Starting EventSource	{"reconciler group": "monitoring.rhobs", "reconciler kind": "MonitoringStack", "source": "kind source: *v1.PodDisruptionBudget"}

```
As opposed to current

```
1.6523301588633049e+09	INFO	monitoring-stack-operator	Updating stack component	{"Stack": "e2e-tests/empty-stack", "Component": "Role.rbac.authorization.k8s.io/v1", "Name": "empty-stack-prometheus"}
```

sthaha:
@tremes Thank you for the patch, quite useful! 
I agree with Jan, that it is best to document the motivation in the commit message 👼 .

tremes:
@sthaha Thanks for the snippets. I was about to add them. Yeah the motivation is to have human readable time format.

sthaha:
@tremes , I think the suggestion was to amend the commit to add those.

tremes:
@sthaha Sorry. Reading too quick....is that better now pls?

sthaha (approved):

PR #136, opened 2022-05-17
chore(release): 0.0.8

sthaha:


CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [0.0.8](https://github.com/rhobs/monitoring-stack-operator/commit/) (2022-05-17)
```
sthaha, line 5: NOTE: not bumping to 0.1.0 because of a breaking change.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+
+
+### ⚠ BREAKING CHANGES
+
+* [MON-2247](https://issues.redhat.com/browse/MON-2247): MSO does not deploy grafana opeator anymore.
```
jan--f, line 10:
```suggestion
* [MON-2247](https://issues.redhat.com/browse/MON-2247): MSO does not deploy grafana operator anymore.
```

CHANGELOG.md
jan--f, line 5: This link url is the same for every release and points to master. Wouldn't it be better if it linked to the actual tag?

jan--f (changes requested):

CHANGELOG.md
sthaha: Agree ... 
I think this happens because we use `--skip.tag` and let the CI create and push the tag. However, referring to the tag that will be created later should be fine.

jan--f:
It'd be really nice to get #140 in for this bump.

jan--f (approved):

PR #140, opened 2022-05-18
feat: update prometheus-operator dependency in go.mod

jan--f:
This updates the prometheus-operator go dependency to v0.55.0. This
allows specifiying additional auth options in remote_write sections.


JoaoBraveCoding:
Since we are updating the PO, do you think we could also take the opportunity to update the PO deployment under `deploy/dependencies/prometheus-operator-deployment.yaml`? Otherwise we would be running PO CRD from v0.55.0 and a deployment from v0.50.0.

JoaoBraveCoding (commented):
>  We should probably automate something here?

100% agree.

LGTM 👍 just two last comments:
- we might want to ship `v0.55.1` since it contains a bugfix for `queryLogFile`.
- to make everything consistent we might also want to update the PO jsonnet version, under `jsonnet/jsonnetfile.json`, it's still pointing to `v0.52.1`.  AFAIK this is only being used to generate PrometheusRules and it will not change anything but it will make things consistent.

simonpasquier (approved):
lgtm

sthaha (changes requested):
Requesting Changes so that this does not get merged by mistake. 

Like @JoaoBraveCoding pointed out, I think we should update all the files in `deploy/dependencies` including cluster-roles, rolebindings etc to match what's upstream. 

Next we need to update the `deploy/crds/kubernetes` so that tests on kind cluster will run fine. 

Does updating to 0.55.1 also limit the openshift versions we support ?

jan--f:
> @jan--f I think you have list everything (including more than what I could think of).
> 
> I was wondering if we could replace the step (need not be in this PR but something worth thinking about) with generate using controller-gen the crds ? (~I haven't tried this~)
> 
> > Update deploy/crds/kubernetes with files from https://github.com/prometheus-operator/prometheus-operator/tree//example/prometheus-operator-crd/
> 
> This seems to work fine.
> 
> ```makefile
> .PHONY: generate-prom-operator-crds
> generate-prom-operator-crds: $(CONTROLLER_GEN)
> 	$(CONTROLLER_GEN) crd \
> 		paths=github.com/prometheus-operator/prometheus-operator/pkg/apis/... \
> 		output:dir=. \
> 		output:crd:dir=./deploy/crds/kubernetes
> ```
> 
> But generation of rbac will need to be copied from upstream since I didn't find controllers using `kubebuilder:rbac`

Which rbac resource are you referring to? I think the only ones we can't generate should be covered by kustomize below?

As for the CRD generation, the above snippet works indeed. However it seems we don't have any control over the version that `controller-gen` uses with that path, right?
How about we simply shallow-clone the p-o repo in a temp dir and generate from that? That allows us to specify the commit we want and might make debugging a bit simpler if we ever need it.

> 
> Also
> 
> > Update version tags in deploy/dependencies/*
> 
> Can we use some kustomize magic to do the replacement ?

Looks like it, check out https://github.com/rhobs/monitoring-stack-operator/pull/140/commits/d8d3d7f0fe40b500a95f4ef7ec8a2e95ed9215be. I'd even be inclined to just generate those resources from the upstream examples. Wdyt?

jan--f:
> As for the CRD generation, the above snippet works indeed. However it seems we don't have any control over the version that `controller-gen` uses with that path, right?

Or does `controller-gen` simply respect what we have in go.mod? I fail to find any documentation about this :thinking:

sthaha:
> Or does controller-gen simply respect what we have in go.mod? I fail to find any documentation about this 🤔

The initial implementation I had was to use `go mod vendor` and use the `vendor/github.com/..` to generate. That way we are explicit about the version. I think it is worth reading the source a bit to see how it picks up the package to parse.

deploy/dependencies/kustomization.yaml
```diff
@@ -1,9 +1,50 @@
 resources:
-- prometheus-operator-cluster-role-binding.yaml
-- prometheus-operator-cluster-role.yaml
-- prometheus-operator-deployment.yaml
-- prometheus-operator-service-account.yaml
+- https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.55.1/example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
```
sthaha, line 2: neat!! ❤️

jan--f:
Awesome, I like the go.mod and controller-gen setup. Will add this to the PR.

JoaoBraveCoding:
Awesome work! ❤️  Looking forward to how it looks in the end!

Makefile
```diff
@@ -61,6 +61,16 @@ docs: $(CRDOC)
...
+# correct version of the CRD manifests.
+.PHONY: generate-prom-op-crds
+generate-prom-operator-crds: $(CONTROLLER_GEN)
+	$(CONTROLLER_GEN) crd \
+		paths=github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/... \
+		output:dir=. \
+		output:crd:dir=./deploy/crds/kubernetes
+
 .PHONY: generate-crds
 generate-crds: $(CONTROLLER_GEN)
```
sthaha, line 75: I feel `generate-crds` should generate prom crds as well, WDYT?

deploy/crds/kubernetes/kustomization.yaml
```diff
@@ -1,11 +1,11 @@
----
+----
```
sthaha, line 1: isn't it 3 `-` -> `---` ? 🤔

sthaha:
Nice improvement @jan--f ❤️

sthaha (approved):

PR #143, opened 2022-05-24
WIP feat: report & update basic status condition

tremes:
This adds conditions to the MonitorinStackStatus type and
updates the controller logic so that the conditions are updated
during the reconciliation.


pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		err := reconciler(ctx, r.k8sClient, r.scheme)
 		// handle create / update errors that can happen due to a stale cache by
 		// retrying after some time.
 		if errors.IsAlreadyExists(err) || errors.IsConflict(err) {
 			logger.V(3).Info("skipping reconcile error", "err", err)
+			availableCondition.Status = metav1.ConditionFalse
+			availableCondition.Reason = "Failure"
+			availableCondition.Message = fmt.Sprintf("Monitoring stack failed to reconcile: %v", err)
+			err := r.updateStatusCondition(ctx, ms, availableCondition)
+			if err != nil {
```
sthaha, line 151: We shouldn't set "Failure" here since like the comment says, we are ignoring `AlreadyExists` and `Conflict` errors.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	reconcilers := stackComponentReconcilers(ms, r.instanceSelectorKey, r.instanceSelectorValue)
 	for _, reconciler := range reconcilers {
 		err := reconciler(ctx, r.k8sClient, r.scheme)
 		// handle create / update errors that can happen due to a stale cache by
 		// retrying after some time.
 		if errors.IsAlreadyExists(err) || errors.IsConflict(err) {
 			logger.V(3).Info("skipping reconcile error", "err", err)
+			availableCondition.Status = metav1.ConditionFalse
+			availableCondition.Reason = "Failure"
+			availableCondition.Message = fmt.Sprintf("Monitoring stack failed to reconcile: %v", err)
```
sthaha, line 149:
```suggestion
			updateFailed(&availableCondition)
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err != nil {
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{}, err
 		}
 	}
-	return ctrl.Result{}, nil
+	availableCondition.Status = metav1.ConditionTrue
+	availableCondition.Reason = "Ready"
+	availableCondition.Message = "Monitoring stack is up and running"
```
sthaha, line 169: This is misleading ... since the stack can fail for various other reasons. A better message may be "Monitoring stack has been deployed" or something on those lines.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
+	availableCondition := metav1.Condition{
+		Type:               "Available",
+		ObservedGeneration: ms.ObjectMeta.Generation,
+		LastTransitionTime: metav1.Time{Time: time.Now().UTC()},
+	}
+
```
sthaha, line 139:
```suggestion
```
IMHO construction of the condition should move to a different functions. say 
`new[Failed|Unknown|Ready]Condition(available, msg, ms.ObjectMeta.Generation)` 
where `available` is a `const available = "Available"`

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	availableCondition.Message = "Monitoring stack is up and running"
+	err = r.updateStatusCondition(ctx, ms, availableCondition)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
+}
+
+func (r *reconciler) updateStatusCondition(ctx context.Context, ms *stack.MonitoringStack, c metav1.Condition) error {
+	updatedMS := ms.DeepCopy()
```
sthaha, line 178: I think we can delay deepcloning until there is a need to update.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		if existingC.Reason == c.Reason {
+			return nil
+		}
+	}
+	updatedMS.Status.Conditions = append(newStatus.Conditions, c)
+	err := r.k8sClient.Status().Update(ctx, updatedMS)
+	if err != nil {
+		return err
+	}
+	return nil
```
sthaha, line 192:
```suggestion
	return r.k8sClient.Status().Update(ctx, updatedMS)
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+func (r *reconciler) updateStatusCondition(ctx context.Context, ms *stack.MonitoringStack, c metav1.Condition) error {
+	updatedMS := ms.DeepCopy()
+	newStatus := stack.MonitoringStackStatus{}
+
+	// if the condition exists and the reason hasn't changed then do no update
+	if existingC := getConditionByType(c.Type, updatedMS.Status.Conditions); existingC != nil {
+		if existingC.Reason == c.Reason {
+			return nil
+		}
+	}
```
sthaha, line 186:
```suggestion
	if existing := conditionForType(updatedMS.Status.Conditions, c.Type); existing != nil && existing.Reason == c.Reason {
		return nil
	}
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	updatedMS.Status.Conditions = append(newStatus.Conditions, c)
```
sthaha, line 187: won't this clear other conditions say `Progressing | Ready` that are set?

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -130,20 +131,74 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	err := r.k8sClient.Status().Update(ctx, updatedMS)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getConditionByType(ctype string, conditions []metav1.Condition) *metav1.Condition {
```
sthaha, line 195:
```suggestion
func conditionForType(conditions []metav1.Condition, x conditionType) *metav1.Condition {
```
I am inclined to create an enum for the types e.g .. 
```
type conditionType string
const available conditionType = "Availble"
```

pkg/controllers/monitoring-stack/controller.go
tremes: yes, but there are no other conditions at the moment

pkg/controllers/monitoring-stack/controller.go
tremes: OK

pkg/controllers/monitoring-stack/controller.go
tremes: What about "Monitoring stack has been reconciled" (or updated)? "Deployed" sounds like one time action to me.

pkg/controllers/monitoring-stack/controller.go
tremes: Hum so what should we return in this case please? Requeue as well with `ctrl.Result{RequeueAfter: 2 * time.Second}, nil` or do not update the status at all?

pkg/controllers/monitoring-stack/controller.go
tremes: OK. Thanks

pkg/controllers/monitoring-stack/controller.go
tremes: Yes why not.... I was thinking about adding something like `map[conditionType]Condition` to the `reconciler` type, but not sure if it's necessary right now.

pkg/controllers/monitoring-stack/controller.go
sthaha: In that case do even need to search and append? IMHO, we should try and follow the implementation in Prometheus Operator.

pkg/controllers/monitoring-stack/controller.go
sthaha: Creating a type to indicate condition type is necessary even if there is only one condition to avoid any typos and to have a single place to change / make corrections.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -39,6 +39,12 @@ import (
 
 	"github.com/go-logr/logr"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Status condition type available
+	available = "Available"
```
sthaha, line 47: this need to be a type instead of raw string so that it can methods can accept and return condition type instead of a raw string.

pkg/controllers/monitoring-stack/controller.go
sthaha: We should preserve the old behaviour. we shouldn't set any status here not return error because we are just ignoring an error due to stale cache.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -137,15 +143,76 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			failedCondition := newFailedCondition(available, err, ms.ObjectMeta.Generation)
+			err := r.updateStatusCondition(ctx, ms, failedCondition)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
 		if err != nil {
+			failedCondition := newFailedCondition(available, err, ms.ObjectMeta.Generation)
+			err := r.updateStatusCondition(ctx, ms, failedCondition)
```
sthaha, line 155: I think we can simplify this to `return r.updateStatusCondition(ctx, ms, failedCondition)` by making `r.updateStatusCondition` return `ctrl.Result, error`

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -137,15 +143,76 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err != nil {
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{}, err
```
sthaha, line 159:
```suggestion
			cond := newFailedCondition(available, err, ms.ObjectMeta.Generation)
			return r.updateStatusCondition(ctx, ms, cond)
```

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -137,15 +143,76 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		}
 	}
+	readyCondition := newReadyCondition(available, ms.ObjectMeta.Generation)
+	err = r.updateStatusCondition(ctx, ms, readyCondition)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 	return ctrl.Result{}, nil
```
sthaha, line 167:
```suggestion
	ready := newReadyCondition(available, ms.ObjectMeta.Generation)
	return r.updateStatusCondition(ctx, ms, ready)
```

pkg/controllers/monitoring-stack/controller.go
sthaha: I agree `Deployed` does sound like a one time action.  I feel `updated` sounds better than reconciled.

pkg/controllers/monitoring-stack/controller.go
simonpasquier: I guess it depends if there's only one place in the code where we update the conditions or if can done from different places. But I suppose that for now it doesn't matter much.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -140,12 +148,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			return r.updateStatusCondition(ctx, ms, availableCondition.Failed(msg, "Failure", ms.ObjectMeta.Generation))
 		}
 	}
+	msg := "Monitoring stack successfully updated"
+	return r.updateStatusCondition(ctx, ms, availableCondition.Ready(msg, "Ready", ms.ObjectMeta.Generation))
+}
+
+func (r *reconciler) updateStatusCondition(ctx context.Context, ms *stack.MonitoringStack, c metav1.Condition) (ctrl.Result, error) {
+	// if the condition exists and the reason hasn't changed then do no update
+	if existingC := getConditionByType(ms.Status.Conditions, c.Type); existingC != nil && existingC.Reason == c.Reason {
```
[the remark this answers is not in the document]
tremes: Yes I think it makes sense to update the conditions. I am looking at the prometheus-operator and if I am not mistaken the conditions are basically always updated, only the `LastTransitionTime` is updated when the condition status changes.

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -140,12 +148,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
 		if err != nil {
-			return ctrl.Result{}, err
+			msg := fmt.Sprintf("Monitoring stack failed to reconcile: %v", err)
```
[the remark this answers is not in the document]
tremes: I think the info (I guess the object name) should be available in the error message but let me try to verify it.

pkg/controllers/monitoring-stack/controller.go
jan--f: > Do we just want to replicate the status conditions from the Prometheus (if so then what about the alertmanager status?) or do we want to somehow "wrap" them (i.e introduce something like `PrometheusAvailable`, `PrometheusReconciled`)?

I think its fine to simply pull in or forward the `Prometheus` CR status into the `MonitoringStack` CR status. It seems natural and ideal the `Prometheus` status already adheres to [best practices](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).
This can also be the start of the status reporting. If we later find we want to add a condition for the operator itself or some other component that managed through a `MonitoringStack` CR, we'll simply add it.
So with that it probably makes sense to introduce our own condition type and embed the [PrometheusStatus](https://github.com/prometheus-operator/prometheus-operator/blob/89d905f764c81d3e55342fbd2919fcdba881c609/pkg/apis/monitoring/v1/types.go#L567) type. Then we can our own condition type (or others) later if needed.

I hope this makes sense, I'm far from an expert in reporting CR status.

pkg/controllers/monitoring-stack/controller.go
tremes: Outdated

pkg/controllers/monitoring-stack/controller.go
tremes: outdated

pkg/controllers/monitoring-stack/controller.go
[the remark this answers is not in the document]
tremes: Thanks

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,17 +151,50 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+}
+
+func newPrometheusCondition(pc monv1.PrometheusCondition, generation int64) metav1.Condition {
+	return metav1.Condition{
+		Type:               string(pc.Type),
+		Message:            pc.Message,
+		Reason:             pc.Reason,
+		LastTransitionTime: pc.LastTransitionTime,
+		Status:             metav1.ConditionStatus(pc.Status),
```
tremes, line 189: This can currently fails because the [`PrometheusConditionStatus`](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L632) doesn't conform to [`Kubernetes ConditionStatus`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ConditionStatus) (there's no degraded value).
So the question is what to do with that:
- fix/update in the Prometheus operator? 
- use the `PrometheusCondition` type here in OO? 
- use the current approach but introduce some tweak/hack to make compliant

simonpasquier: I would vote for the 2nd option but let OBO define its own ConditionStatus with a mapping from PrometheusCondition to ObservabilityCondition. WYDT?

simonpasquier: > So if I understand it correctly the suggestion is basically to introduce a custom condition type for OBO (something like ObservabilityCondtiion or MonitoringStackCondition) right?

yes it would be my suggestion.

pkg/controllers/monitoring-stack/controller.go
tremes: Thanks. I updated the PR. Please check

pkg/controllers/monitoring-stack/controller.go
```diff
@@ -140,12 +148,41 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			return r.updateStatusCondition(ctx, ms, availableCondition.Failed(msg, "Failure", ms.ObjectMeta.Generation))
```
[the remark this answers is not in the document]
tremes: Outdated

tremes:
Hmm so I am wondering why this https://github.com/rhobs/observability-operator/blob/main/Makefile#L83 is needed. I mean I think it's OK to generate the role, but why it's missing the get permissions? I am not able to read the Prometheus status then :-(

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,10 +151,44 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+func (r *reconciler) updateConditions(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	prom, err := r.mcClient.MonitoringV1().Prometheuses(req.Namespace).Get(ctx, ms.Name, metav1.GetOptions{})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	newStatus := stack.MonitoringStackStatus{}
+	for i := range prom.Status.Conditions {
+		pc := prom.Status.Conditions[i]
+		c := newCondition(pc, ms.Generation)
```
JoaoBraveCoding, line 168: nit: Since it's not used anywhere else in the `for` I would do it like this, but I don't have a strong opinion just a suggestion :)
```suggestion
		c := newCondition(prom.Status.Conditions[i], ms.Generation)
```

JoaoBraveCoding (commented):
Sorry for the delay in the review @tremes! From my side it looks good! I've only added a small nit, feel free to address it or simply disregard it ;)

Regarding your problem I think you simply have to fix the annotations in this line [1] and run `make generate` to solve your problem! 

[1] https://github.com/rhobs/observability-operator/blob/main/pkg/controllers/monitoring/monitoring-stack/controller.go#L64

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,10 +151,44 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	return r.updateConditions(ctx, req, ms)
+}
+
+func (r *reconciler) updateConditions(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack) (ctrl.Result, error) {
+	prom, err := r.mcClient.MonitoringV1().Prometheuses(req.Namespace).Get(ctx, ms.Name, metav1.GetOptions{})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	newStatus := stack.MonitoringStackStatus{}
+	for i := range prom.Status.Conditions {
```
simonpasquier, line 166: Rather than copying the Prometheus conditions to the MonitoringStack conditions verbatim, we need to think about how we want to combine the MonitoringStack conditions with the children's conditions.

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -77,8 +77,56 @@ type MonitoringStackSpec struct {
...
 type MonitoringStackStatus struct {
-	// TODO(sthaha): INSERT ADDITIONAL STATUS FIELDS -- observed state of prometheus
-	// ??
+	// Conditions provide status information about the MonitoringStack
+	// +listType=map
+	// +listMapKey=type
+	Conditions []Condition `json:"conditions"`
+}
+
+type ConditionStatus string
```
simonpasquier, line 86: Can we enumerate the possible values for clarity?

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -77,8 +77,56 @@ type MonitoringStackSpec struct {
...
+	Reason string `json:"reason"`
+	// message is a human readable message indicating details about the transition.
+	// This may be an empty string.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=32768
+	Message string `json:"message"`
+	// status of the condition
+	// +required
+	// +kubebuilder:validation:Required
```
simonpasquier, line 128: we can have a kubebuilder validation on the supported enum values.

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,10 +151,44 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
 		if err != nil {
-			return ctrl.Result{}, err
+			return r.updateConditions(ctx, req, ms)
```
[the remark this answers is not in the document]
tremes: It would be nice to mention the name &kind (in the message) of the resource which failed to reconcile. I am wondering if we can rely on the error message or if we will need to explicitly have/get the resource details. I'm afraid the latter seems more likely.....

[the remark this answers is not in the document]
simonpasquier: As a first approach, we can assume that the error will provide enough information. If not the case, we can improve later.

tremes:
I updated the PR based on the reviews. The OBO is reconciled faster than Prometheus containers (`init-config-reloader`) are ready (at least in my case) so the "MonitoringStackAvailable" condition is False....I guess we should wait for the containers to be ready or reconcile again? Or is that about the [controller settings](https://github.com/rhobs/observability-operator/blob/main/pkg/controllers/monitoring/monitoring-stack/controller.go#L98) (should we own the P-O or the Prometheus StatefulSet?)

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -98,14 +116,18 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
...
+		Owns(&monv1.Prometheus{}).
 		Owns(&monv1.Alertmanager{}).WithEventFilter(p).
 		Owns(&v1.Service{}).WithEventFilter(p).
 		Owns(&v1.ServiceAccount{}).WithEventFilter(p).
 		Owns(&rbacv1.Role{}).WithEventFilter(p).
 		Owns(&rbacv1.RoleBinding{}).WithEventFilter(p).
 		Owns(&monv1.ServiceMonitor{}).WithEventFilter(p).
 		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
+		Watches(
+			&source.Kind{Type: &monv1.Prometheus{}},
```
JoaoBraveCoding, line 128: Not quite sure I understand this part. I might be missing something, but isn't  `Owns(&monv1.Prometheus{}).` already responsible for triggering reconciliation when Prometheus CRs are updated?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if prometheusAvailable.Status != monv1.PrometheusConditionTrue {
+		ac.Status = stack.ConditionFalse
+		ac.Reason = prometheusNotAvailable
+		ac.Message = prometheusAvailable.Message
+		ac.LastTransitionTime = metav1.Now()
+		return
+	}
+
+	if ac.Status != stack.ConditionTrue {
```
JoaoBraveCoding, line 65: Why do we only update the time when the condition is not true?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,127 @@
...
+		ac.LastTransitionTime = metav1.Now()
+	}
+	ac.Status = stack.ConditionTrue
+	ac.Reason = monitoringStackAvailable
+	ac.Message = "Monitoring Stack is available"
```
JoaoBraveCoding, line 70: I'm afraid my low experience with golang might be bitting me but here in the end (and before every return) shouldn't we update again something like:
```golang
cm.conditions[stack.Available] = ac
```
Same for the other Update<Condition> functions.

JoaoBraveCoding:
> I updated the PR based on the reviews. The OBO is reconciled faster than Prometheus containers (`init-config-reloader`) are ready (at least in my case) so the "MonitoringStackAvailable" condition is False....I guess we should wait for the containers to be ready or reconcile again? Or is that about the [controller settings](https://github.com/rhobs/observability-operator/blob/main/pkg/controllers/monitoring/monitoring-stack/controller.go#L98) (should we own the P-O or the Prometheus StatefulSet?)

So IIUC what's happening is that the condition `MonitoringStackAvailable` never moves from `False` to `Available` correct? Now if my understanding of operators is correct, we shouldn't need to own anything else apart from the `Prometheus` CR. This CR should be updated with the correct status and inherently our operator should be notified of those updates and update the `MonitoringStack` CR accordingly.

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Ohh yeah good point. It doesn't make sense when I am thinking about it again. Thanks.

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: I don't think so. We update the reference in the map so it should be OK IMO.

pkg/controllers/monitoring/monitoring-stack/controller.go
tremes: Yes. Correct. This was just updated. Sorry. However I will raise another question here.

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -80,32 +85,35 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
...
-		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
+		For(&stack.MonitoringStack{}).WithEventFilter(predicate.ResourceVersionChangedPredicate{}).
+		Owns(&monv1.Prometheus{}).
+		Owns(&monv1.Alertmanager{}).
+		Owns(&v1.Service{}).
+		Owns(&v1.ServiceAccount{}).
+		Owns(&rbacv1.Role{}).
+		Owns(&rbacv1.RoleBinding{}).
+		Owns(&monv1.ServiceMonitor{}).
+		Owns(&policyv1.PodDisruptionBudget{}).
```
tremes, line 116: TBH I don't know how these filter predicates are supposed to work when there are more of them. So it would be good if some k8s controller-runtime expert double check. 
I was struggling (with original definition) with Prometheus being not available and no additional reconcile attempt (so the conditions were not updated correctly)

pkg/controllers/monitoring/monitoring-stack/conditions.go
JoaoBraveCoding: Correct my bad, didn't noticed we were using a reference

pkg/controllers/monitoring/monitoring-stack/controller.go
simonpasquier: I think you're on something. IIUC `*Builder.WithEventFilter()` applies to all watched objects. We need [Predicates.ApplyToOwns](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/builder#Predicates.ApplyToOwns) for the Prometheus resource:

```golang
Owns(&monv1.Prometheus{}, builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}))
```

tremes: Thanks @simonpasquier. It seems to work fine. Let me update the PR.

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -80,32 +86,35 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
...
-		Owns(&monv1.ServiceMonitor{}).WithEventFilter(p).
-		Owns(&policyv1.PodDisruptionBudget{}).WithEventFilter(p).
+		Owns(&monv1.Prometheus{}, builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		Owns(&monv1.Alertmanager{}).
+		Owns(&v1.Service{}).
+		Owns(&v1.ServiceAccount{}).
+		Owns(&rbacv1.Role{}).
+		Owns(&rbacv1.RoleBinding{}).
+		Owns(&monv1.ServiceMonitor{}).
+		Owns(&policyv1.PodDisruptionBudget{}).
```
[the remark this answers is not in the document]
tremes: OK

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,10 +154,30 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	ms.Status.Conditions = r.conditions.ToSlice()
+	err = r.k8sClient.Status().Update(ctx, ms)
+	if err != nil {
+		if errors.IsConflict(err) {
+			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+		}
+		return ctrl.Result{}, err
+	}
 	return ctrl.Result{}, nil
+
```
simonpasquier, line 180:
```suggestion
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -182,6 +182,39 @@ func reconcileStack(t *testing.T) {
 	assert.DeepEqual(t, expected.ServiceMonitorSelector, generated.Spec.ServiceMonitorSelector)
 	assert.Equal(t, expected.LogLevel, generated.Spec.LogLevel)
 	assert.Equal(t, expected.Retention, generated.Spec.Retention)
+
```
simonpasquier, line 185: I think that we can leverage the status field more generally in the e2e tests: whenever we use `f.K8sClient.Create(...)` and expect a success, we should rewrite it to `f.K8sClient.CreateAndWaitAvailable(...)` which would create the stack and wait for Available/Reconciled to be True. I'm not sold on the name of the function though ;)

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+	monitoringStackAvailable       = "MonitoringStackAvailable"
+	monitoringStackReconciled      = "MonitoringStackReconciled"
+	prometheusNotAvailable         = "PrometheusNotAvailable"
+	prometheusNotReconciled        = "PrometheusNotReconciled"
+	cannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
+)
+
+type conditions map[stack.ConditionType]*stack.Condition
+
+type conditionsMap struct {
```
simonpasquier, line 19: not sure if we need this additional type? couldn't we use `conditions` directly?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
+package monitoringstack
+
+import (
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	stack "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
```
simonpasquier, line 5:
```suggestion
	"github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	monitoringStackAvailable       = "MonitoringStackAvailable"
+	monitoringStackReconciled      = "MonitoringStackReconciled"
+	prometheusNotAvailable         = "PrometheusNotAvailable"
+	prometheusNotReconciled        = "PrometheusNotReconciled"
+	cannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
+)
+
+type conditions map[stack.ConditionType]*stack.Condition
```
simonpasquier, line 17: To avoid any issue with nil pointers
```suggestion
type conditions map[stack.ConditionType]stack.Condition
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+
+type conditionsMap struct {
+	conditions
+}
+
+// CreateConditions creates a new map with MonitoringStack status conditions.
+// The newly created conditions are in unknown state.
+func CreateConditions() conditionsMap {
```
simonpasquier, line 25:
```suggestion
func NewConditions() conditionsMap {
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+// UpdateAvailable gets existing "Available" condition and updates its parameters
+// based on the Prometheus "Available" condition
+func (cm *conditionsMap) UpdateAvailable(prometheusConditions []monv1.PrometheusCondition, generation int64) {
+	ac := cm.conditions[stack.Available]
+	ac.ObservedGeneration = generation
+
+	prometheusAvailable := getPrometheusCondition(prometheusConditions, monv1.PrometheusAvailable)
+
+	if prometheusAvailable == nil {
+		ac.Status = stack.ConditionFalse
```
simonpasquier, line 50: should be unknown?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (cm *conditionsMap) ToSlice() []stack.Condition {
+	s := make([]stack.Condition, 0, len(cm.conditions))
+	for _, v := range cm.conditions {
+		s = append(s, *v)
+	}
+	return s
+}
+
+func getPrometheusCondition(prometheusConditions []monv1.PrometheusCondition, t monv1.PrometheusConditionType) *monv1.PrometheusCondition {
```
simonpasquier, line 114: The function could return a value + error instead of a pointer and a non-nil error would mean that the condition type couldn't be found. This would be more idiomatic :)

```suggestion
func getPrometheusCondition(prometheusConditions []monv1.PrometheusCondition, t monv1.PrometheusConditionType) (monv1.PrometheusCondition, error) {
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,121 @@
...
+		ac.Message = cannotReadPrometheusConditions
+		ac.LastTransitionTime = metav1.Now()
+		return
+	}
+
+	if prometheusAvailable.Status != monv1.PrometheusConditionTrue {
+		ac.Status = stack.ConditionFalse
+		ac.Reason = prometheusNotAvailable
+		ac.Message = prometheusAvailable.Message
+		ac.LastTransitionTime = metav1.Now()
```
simonpasquier, line 61: Ideally we should only update the transition time if the condition has changed?
Also we don't update the transition time if the status is true.

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: I think we need the extra type, because we can't do receiver like `func (cm *conditions) UpdateReconciled`only `func (cm conditions) UpdateReconciled`, which means copying the map right...pls correct me if I am wrong

test/e2e/monitoring_stack_controller_test.go
tremes: Yes something like `GetAndAwait`. Perhaps I can create some function in the test framework or assertions.

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Updated. Thanks

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Updated. Thanks

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Updated. Thanks

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Updated. Thanks

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Well why? I think it's easier & better to work with pointers.

pkg/controllers/monitoring/monitoring-stack/conditions.go
simonpasquier: `func (cm conditions) UpdateReconciled(...)` would update the map despite being a value receiver I believe (see https://stackoverflow.com/questions/49176090/map-as-a-method-receiver)

pkg/controllers/monitoring/monitoring-stack/conditions.go
simonpasquier: you don't have to check for nil pointers so no risk of panics

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,8 +154,27 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	return r.updateStatus(ctx, req, ms, nil)
+}
+
+func (r *reconciler) updateStatus(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack, recError error) (ctrl.Result, error) {
+	prom, err := r.monClient.MonitoringV1().Prometheuses(req.Namespace).Get(ctx, ms.Name, metav1.GetOptions{})
+	if err != nil {
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+	}
+	r.conditions.UpdateAvailable(prom.Status.Conditions, ms.Generation)
+	r.conditions.UpdateReconciled(prom.Status.Conditions, ms.Generation, recError)
```
sthaha, line 170: Given that `conditions` is short lived and can be always derived from promethues.status.conditions should the reconciler hold onto `conditions` map? This also leads me to think what happens if there are 2 or prometheus instances (CR) wouldn't the condition map get overwritten every time and possibly be incorrect? would that even be a race condition? 



```suggestion
	ms.Status.Conditions = conditionsFromPromethues(prom.Status, ms.Generation, recError)
```

tremes: Hm good point. You're probably right that it's not safe to hold the conditions map in the controller. Let me try to update it.

pkg/controllers/monitoring/monitoring-stack/controller.go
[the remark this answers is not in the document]
tremes: Unfortunately this breaks it again - not enough reconciliations for the Prometheus.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,123 @@
...
+
+const (
+	monitoringStackAvailable       = "MonitoringStackAvailable"
+	monitoringStackReconciled      = "MonitoringStackReconciled"
+	prometheusNotAvailable         = "PrometheusNotAvailable"
+	prometheusNotReconciled        = "PrometheusNotReconciled"
+	cannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
+)
+
+func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.PrometheusCondition, generation int64, recError error) []v1alpha1.Condition {
```
sthaha, line 19: Thank you for changing the implementation. 
Could you please consider adding a unit test for different scenarios?

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -39,10 +40,14 @@ import (
 
 	"github.com/go-logr/logr"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+
+	monitoringclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
```
sthaha, line 44:
```suggestion
	promclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
```

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -61,7 +66,7 @@ type Options struct {
 //+kubebuilder:rbac:groups=monitoring.rhobs,resources=monitoringstacks/status,verbs=get;update
 
 // RBAC for managing Prometheus Operator CRs
-//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=alertmanagers;prometheuses;servicemonitors,verbs=list;watch;create;update;delete;patch
+//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=alertmanagers;prometheuses;servicemonitors,verbs=list;watch;create;update;delete;patch;get
```
sthaha, line 69: Having to add `get` probably indicates that we are skipping the cache which I think we shouldn't and has in the past led to hard to track reconcile errors due to race conditions caused by mismatch of objects in cache and the one returned by the server. Since the controller already watches prometheus, can't we use the dynamic client to `Get` prometheus from the Cache itself?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,123 @@
...
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	monitoringStackAvailable       = "MonitoringStackAvailable"
+	monitoringStackReconciled      = "MonitoringStackReconciled"
+	prometheusNotAvailable         = "PrometheusNotAvailable"
+	prometheusNotReconciled        = "PrometheusNotReconciled"
```
sthaha, line 15:
```suggestion
	AvailableReason       = "MonitoringStackAvailable"
	ReconciledReason      = "MonitoringStackReconciled"
	UnavailableReason         = "PrometheusNotAvailable"
	NotReconciledReason        = "PrometheusNotReconciled"
```

sthaha: Exported so that this can be used in tests

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -182,6 +182,23 @@ func reconcileStack(t *testing.T) {
...
+	availableMs := f.GetStackWhenAvailable(t, ms.Name, ms.Namespace)
+	availableC := getConditionByType(availableMs.Status.Conditions, stack.AvailableCondition)
+	assert.Check(t, availableC.Reason == "MonitoringStackAvailable", "unexpected Available condition reason")
```
sthaha, line 188:
```suggestion
	assert.Check(t, availableC.Reason == stack.AvailableReason, "unexpected Available condition reason")
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: OK. Good point.

pkg/controllers/monitoring/monitoring-stack/controller.go
tremes: Ahh interesting. The dynamic client you're talking about is the `k8sClient` available in the controller (obtained from the manager) right? I can try that.

sthaha (commented):
looks good to me! 🎊 
I will let others take another look before merging 
cc: @jan--f @simonpasquier @JoaoBraveCoding

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -143,9 +141,31 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+func (r *reconciler) updateStatus(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack, recError error) (ctrl.Result, error) {
+	var prom monv1.Prometheus
+	key := types.NamespacedName{
+		Name:      ms.Name,
+		Namespace: ms.Namespace,
+	}
+	err := r.k8sClient.Get(ctx, key, &prom)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
```
JoaoBraveCoding, line 159: Should we log here something? I'm afraid that if we don't log anything here that if we enter this condition regularly we will not be able to tell that the operator is stuck. WDYT?

JoaoBraveCoding (commented):
LGTM! Just a small comment

pkg/controllers/monitoring/monitoring-stack/controller.go
tremes: I think the `controller-runtime` logs every reconciliation so such situation should be visible from the logs IMO but we can add some log statement here saying something like failing to get the Prometheus instance XYZ?

JoaoBraveCoding: If `controller-runtime` logs it then let's ignore my comment! I'll test it to be sure 😉

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,125 @@
...
+	ReconciledReason               = "MonitoringStackReconciled"
+	PrometheusNotAvailable         = "PrometheusNotAvailable"
+	PrometheusNotReconciled        = "PrometheusNotReconciled"
+	CannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
+	AvailableMessage               = "Monitoring Stack is available"
+	ReconciledMessage              = "Monitoring Stack successfully reconciled"
+)
+
+func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.PrometheusCondition, generation int64, recError error) []v1alpha1.Condition {
+	if len(msConditions) == 0 {
```
simonpasquier, line 22: when would we enter this branch?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if err != nil {
+		ac.Status = v1alpha1.ConditionUnknown
+		ac.Reason = PrometheusNotAvailable
+		ac.Message = CannotReadPrometheusConditions
+		return ac
+	}
+
+	if prometheusAvailable.Status != monv1.PrometheusConditionTrue {
+		ac.Status = v1alpha1.ConditionStatus(prometheusAvailable.Status)
```
simonpasquier, line 75: though there's a 1:1 mapping for now, it's probably safer to have an explicit map between the 2 types.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		rc.Status = v1alpha1.ConditionFalse
+		rc.Message = err.Error()
+		rc.Reason = "FailedToReconcile"
+		return rc
+	}
+	prometheusReconciled, err := getPrometheusCondition(prometheusConditions, monv1.PrometheusReconciled)
+
+	if err != nil {
+		rc.Status = v1alpha1.ConditionFalse
```
simonpasquier, line 100: in `updateAvailable()`, we set the status to unknown. Any reason to have it different here?

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: I think in the very first reconciliation when there's already `MonitoringStack` available, but with no conditions.

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Yes, good catch. Thanks

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -88,24 +90,19 @@ func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
 		instanceSelectorValue: split[1],
 		grafanaDSWatchCreated: false,
 	}
-
-	// We only want to trigger a reconciliation when the generation
-	// of a child changes. Until we need to update our the status for our own objects,
-	// we can save CPU cycles by avoiding reconciliations triggered by
-	// child status changes.
```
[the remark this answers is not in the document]
tremes: OK

pkg/controllers/monitoring/monitoring-stack/controller.go
sthaha: I think @JoaoBraveCoding raises a good point. Should we ignore all errors? how about ignoring only NotFound and treat the rest as actual errors?

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Updated

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -182,6 +183,23 @@ func reconcileStack(t *testing.T) {
...
 	assert.Equal(t, expected.LogLevel, generated.Spec.LogLevel)
 	assert.Equal(t, expected.Retention, generated.Spec.Retention)
+
+	availableMs := f.GetStackWhenAvailable(t, ms.Name, ms.Namespace)
+	availableC := getConditionByType(availableMs.Status.Conditions, stack.AvailableCondition)
+	assert.Check(t, availableC.Reason == monitoringstack.AvailableReason, "unexpected Available condition reason")
+	reconciledC := getConditionByType(availableMs.Status.Conditions, stack.ReconciledCondition)
+	assert.Check(t, reconciledC != nil, "failed to find Reconciled status condition for %s monitoring stack", ms.Name)
+	assert.Check(t, reconciledC.Status == stack.ConditionTrue, "unexpected Reconciled condition status")
+	assert.Check(t, reconciledC.Reason == monitoringstack.ReconciledReason, "unexpected Reconciled condition reason")
```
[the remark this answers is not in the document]
tremes: Add the helper function

pkg/controllers/monitoring/monitoring-stack/controller.go
tremes: Yes probably yes. Thanks. Let me update it.

pkg/controllers/monitoring/monitoring-stack/conditions.go
jan--f: You resolved this by changing `updateAvailable` to set `ConditionFalse`. I was thinking `ConditionUnknown` makes more sense here.
Can you outline why you went with `ConditionFalse` instead?

pkg/controllers/monitoring/monitoring-stack/conditions.go
jan--f: Its probably best to use `prometheusStatusToMSStatus(prometheusReconciled.Status)` here as well?

pkg/controllers/monitoring/monitoring-stack/conditions.go
tremes: Ahh yeah right. `ConditionUnknown` makes probably more sense when there's an error reading the Prometheus conditions. I just noticed I missed some other things so I am going to update. Thanks

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -77,8 +77,76 @@ type MonitoringStackSpec struct {
 // MonitoringStackStatus defines the observed state of MonitoringStack.
 // It should always be reconstructable from the state of the cluster and/or outside world.
 type MonitoringStackStatus struct {
-	// TODO(sthaha): INSERT ADDITIONAL STATUS FIELDS -- observed state of prometheus
-	// ??
+	// Conditions provide status information about the MonitoringStack
+	// +listType=map
+	// +listMapKey=type
```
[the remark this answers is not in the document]
simonpasquier: I assume that @tremes took it from the Prometheus Operator;)
I don't think that it matters much since as you said, 1) the controller always updates the status (no patch) and 2) nobody else should be modifying the resource status.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,141 @@
...
+	SuccessfullyReconciledMessage  = "Monitoring Stack is successfully reconciled"
+)
+
+func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.PrometheusCondition, generation int64, recError error) []v1alpha1.Condition {
+	if len(msConditions) == 0 {
+		return []v1alpha1.Condition{
+			{
+				Type:               v1alpha1.AvailableCondition,
+				Status:             v1alpha1.ConditionUnknown,
+				Reason:             "None",
```
simonpasquier, line 28: (nit) a const would be nice :)

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -0,0 +1,141 @@
...
+				LastTransitionTime: metav1.Now(),
+			},
+			{
+				Type:               v1alpha1.ReconciledCondition,
+				Status:             v1alpha1.ConditionUnknown,
+				Reason:             "None",
```
simonpasquier, line 34: here too

pkg/apis/monitoring/v1alpha1/types.go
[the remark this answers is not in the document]
jan--f: Fair points! I would still argue for the "correct" markers if only to inform future developers of the intention.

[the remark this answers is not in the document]
tremes: OK. Thanks

sthaha (commented):
looks good to me (again) 🎊

jan--f:
I'll give it a once over later today. 
@sthaha  I'd like to hold off the on the release for a little bit. I have 0.0.13 sort of ear marked to test our CPaaS pipeline, but I'm still waiting on reviews there. We can go ahead if that ends up taking too long, maybe we can hold off till next week?

simonpasquier:
lgtm too :tada: 
not officially approving it based on the latest @jan--f comment :)

jan--f (approved):
Thanks, this looks great!

simonpasquier:
:1st_place_medal: congratulations @tremes, it wasn't an easy one :)

PR #153, opened 2022-06-02
Build e2e specific bundle

jan--f:
This is an alternative approach to #152.

Not sure if this fully works yet, but I suppose the approach is obvious and I'd appreciate feedback.


.github/olm-publish/action.yaml
```diff
@@ -43,7 +43,7 @@ runs:
 
   - name: Build Bundle Image
     shell: bash
-    run: make bundle-image
+    run: make bundle bundle-image
```
sthaha, line 46: IMHO, bundle-image should depend on bundle.

Makefile
```diff
@@ -151,8 +151,22 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
 		 	$(BUNDLE_METADATA_OPTS)
 	$(OPERATOR_SDK) bundle validate ./bundle
 
+.PHONY: e2e-bundle
+e2e-bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
```
sthaha, line 155: I feel we should test using the same bundle that would deployed in production to avoid any surprises.

deploy/e2e-olm/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+resources:
```
sthaha, line 1: do you think, `deploy/production` or `deploy/osd` would be a better name/target than `e2e-olm` since this represents the production env ?

sthaha:
@jan--f should we close this?

Issue #149, opened 2022-06-02
Prometheus Operator fails to run on 4.11.0-0.ci-2022-05-28 or later


sthaha:
The operator fails with the following status 

```
- lastTransitionTime: "2022-06-02T03:36:46Z"
    lastUpdateTime: "2022-06-02T03:36:46Z"
    message: 'pods "monitoring-stack-operator-prometheus-operator-5874559dd8-" is
      forbidden: unable to validate against any security context constraint: [provider
      "anyuid": Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.runAsUser:
      Invalid value: 65534: must be in the ranges: [1000400000, 1000409999], provider
      "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2":
      Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden:
      not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden:
      not usable by user or serviceaccount, provider "machine-api-termination-handler":
      Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2":
      Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden:
      not usable by user or serviceaccount, provider "hostaccess": Forbidden: not
      usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable
      by user or serviceaccount, provider "privileged": Forbidden: not usable by user
      or serviceaccount]'
    reason: FailedCreate
    status: "True"
```

After manually editing the deployment to set runAsUser from 65534 to 1000400000, it shows the following warning which should also be addressed.

```
Warning: would violate PodSecurity "restricted:latest": seccompProfile (pod or container "prometheus-operator" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
deployment.apps/monitoring-stack-operator-prometheus-operator edited
```

simonpasquier:
The reason for the 65534 uid is to ensure that the pod gets assigned a stable uid/gid and avoid permission issues with persistent volumes. More context in https://github.com/openshift/cluster-monitoring-operator/pull/1660#pullrequestreview-972742840.

PR #154, opened 2022-06-02
fix: Set seccomp profiles and grant SAs necessary premissions to run

JoaoBraveCoding:
When running in namespace with Pod Security Standard profile "restricted"
we need to set RunAsNonRoot and SeccompProfile to all workloads running
on that namespace. Futhermore on OpenShift to run with a SeccompProfile
set we need to grant service accounts premisisons to use the SCC
nonroot-v2 
Fixes https://github.com/rhobs/observability-operator/issues/149

### DIFF
--- a/pkg/controllers/monitoring/monitoring-stack/components.go
+++ b/pkg/controllers/monitoring/monitoring-stack/components.go
@@ -7,6 +7,7 @@ import (
 	stack "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -20,6 +21,8 @@ import (
 )
 
 const AdditionalScrapeConfigsSelfScrapeKey = "self-scrape-config"
+const PrometheusUserFSGroupID = 65534
+const AlertmanagerUserFSGroupID = 65535
 
 type reconcileFunction func(ctx context.Context, c client.Client, scheme *runtime.Scheme) error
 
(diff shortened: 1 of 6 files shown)


pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -163,7 +172,14 @@ func newPrometheus(
 					},
 					Key: AdditionalScrapeConfigsSelfScrapeKey,
 				},
-				Storage:        storageForPVC(config.PersistentVolumeClaim),
+				Storage: storageForPVC(config.PersistentVolumeClaim),
+				SecurityContext: &corev1.PodSecurityContext{
```
simonpasquier, line 176: should we set `fsGroup` too?

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -163,7 +172,15 @@ func newPrometheus(
...
+					FSGroup:      pointer.Int64(65534),
```
simonpasquier, line 177: (nit) we could have a `const` for this

JoaoBraveCoding:
So CI was failing because if I set on the AM deployment the `securityContext: runAsNonRoot: true` then the AM pods do not start with the following error:
```
 Warning  Failed     3s (x5 over 37s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (nobody), cannot verify user is non-root (pod: "alertmanager-self-scrape-0_e2e-tests(6840fcd1-b25a-4d95-bcb4-87da5ded49d8)", container: config-reloader)
```
I checked and indeed the AM dockerimage uses ` nobody`...  but I'm afraid I might be missing something because it feels weird to me that if the problem is indeed with the dockerimage that only now the problem is surfacing. Do know anything about this @simonpasquier

sthaha (approved):

sthaha:
Looks good to me 😍
I will let @simonpasquier approve the PR.

sthaha (commented):
removing approval for @simonpasquier to approve this PR. 
We should consider releasing as soon as this merges.

simonpasquier (approved):

PR #158, opened 2022-06-06
chore: add a script to setup kind locally

sthaha:
This PR adds a script to setup kind cluster locally, so that it is
mimics the CI setup. Notable things in the setup are ...

1. Use the same kind config used in CI
2. Modify control-plane node to be used as infra node
3. Set up OLM
4. Have container registry running in kind for testing bundle deployment


README.md
```diff
@@ -64,6 +64,12 @@ to the controller-gen CLI page in the [kubebuilder documentation](https://book.k
 
 ### Running the operator in kind
 
+#### Automated
+
+Run `./hack/kind/setup.sh` to create a kind cluster that has all the
```
JoaoBraveCoding, line 69: Correct me if I'm wrong but running this will not actually deploy the operator correct? If so I think we should clarify in the readme what instructions are to set up a cluster and what are to run the operator. WDYT of this structure: 
1. Setting up a dev cluster to run OO
     - Automated
     - Manual
2. Running OO on a cluster
     - Catalog install + OLM install
     - Run local
     - Run with local registry

README.md
```diff
@@ -64,6 +64,12 @@ to the controller-gen CLI page in the [kubebuilder documentation](https://book.k
...
+required dependencies (OLM, infra-node, registry).
```
JoaoBraveCoding, line 70: Do you think it also makes sense to add a teardown script? 🤔 
If not I feel that at least deprovision instructions would be nice
```
kind delete cluster --name obs-operator
```

hack/kind/setup.sh
```diff
@@ -0,0 +1,105 @@
...
+
+  curl http://local-registry:30000 || {
+    die "Failed to reach local-registry running"
+  }
+  echo -e "      ---------------------------------- \n"
+
+}
+
+setup_cluster() {
+  kind create cluster \
```
JoaoBraveCoding, line 67: Should we try making the script idempotent and validate if a cluster is not already provisioned?

hack/kind/setup.sh
```diff
@@ -0,0 +1,105 @@
...
+    info "echo \"127.0.0.1\tlocal-registry\" | sudo tee -a /etc/hosts"
+    die "/etc/hosts does not contain local-registry entry"
+  }
+
+  kubectl apply -f ./hack/kind/registry.yaml -n operators
+  kubectl rollout status deployment local-registry -n operators
+
+  kubectl wait --for=condition=Available deploy local-registry -n operators --timeout=300s
+
+  curl http://local-registry:30000 || {
```
JoaoBraveCoding, line 59: In my machine, this step actually failed. I commented out the steps before `run-registry` and re-ran the script and it worked. So I feel like here we should actually retry the connection for a couple of sec/min.

JoaoBraveCoding (commented):
Sorry for the late late review... great work! Just some comments but we should be able to merge this soon 😉

README.md
sthaha: The script does print how to delete the cluster but no harm in mentioning it here. We do have to keep the docs updated if anything changes in the script.

hack/kind/setup.sh
sthaha: I am surprised why this happens .. The deployment did say it is available (see above) . 
I will add a return / max-timeout to curl.

hack/kind/setup.sh
sthaha: I don't think we should strive to do that since it also means we have to validate each step before executing it - olm setup, registry, etc. It should be more like delete and `./setup.sh` if things go wrong

hack/kind/setup.sh
```diff
@@ -0,0 +1,105 @@
...
+  setup_olm
+  run_registry
+
+  info "Waiting for cluster boot to complete ..."
+  echo -e "      ---------------------------------- \n"
+  kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout=300s
+  echo -e "      ---------------------------------- \n"
+
+  info "NOTE: To delete the cluster, run:
+    kind delete cluster --name obs-operator \n"
```
sthaha, line 94: @JoaoBraveCoding here is where the script outputs how to delete the cluster.

README.md
sthaha: > running this will not actually deploy the operator correct?

That's right, the intention of the setup is to not run the operator but to setup the bare minimal cluster so that operator can be run. It in fact does not even apply the `deploy/common/kubernetes` and `crds` because often you may want to test without these dependencies.

sthaha: > Running OO on a cluster

I like the idea!
How about in a follow up patch so that this one is minimal and only introduces one change - setup a cluster.

JoaoBraveCoding: > I actually feel we should abandon the manual part and rely on an automated setup so that everyone has a similar environment.

100% agree, the more consistent we make environments the better so I'm super down for this 👍 

>  How about in a follow up patch so that this one is minimal and only introduces one change - setup a cluster.

Sounds good ✅

hack/kind/setup.sh
JoaoBraveCoding: No problem, I'll then just make a small suggestion to mention what users should run if they face an issue with the script.

hack/kind/setup.sh
JoaoBraveCoding: Yes, my bad 😅

hack/kind/setup.sh
JoaoBraveCoding: Tried it again and it ran perfectly 👌

PR #159, opened 2022-06-09
feat: update prometheus-operator dependency in go.mod

tremes:
This updates the prometheus-operator go dependency to v0.57.0


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -60,9 +60,8 @@ type MonitoringStackSpec struct {
 
 	// Time duration to retain data for. Default is '120h',
 	// and must match the regular expression `[0-9]+(ms|s|m|h|d|w|y)` (milliseconds seconds minutes hours days weeks years).
-	// +kubebuilder:validation:Pattern="^[0-9]+(ms|s|m|h|d|w|y)$"
```
tremes, line 63: I hope this is safe to remove when there's validation at https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L385

jan--f: Yeah that looks great, nice improvement!

jan--f (changes requested):
Two places still need the update:

- [ ] The dependency resources need the bump: https://github.com/rhobs/observability-operator/blob/f83e38b5df749390a4212525ab601486c7e4c2da/deploy/dependencies/kustomization.yaml#L2-L5
- [ ] The mixins from prometheus-operator need the bump: https://github.com/rhobs/observability-operator/blob/f83e38b5df749390a4212525ab601486c7e4c2da/jsonnet/jsonnetfile.json#L11

jan--f (approved):
Thanks!

PR #161, opened 2022-06-15
fix: grants SA of components access to nonroot SCC

JoaoBraveCoding:
Necessary for cases where nonroot-v2 still does not exist e.g OpenShift 4.10. Otherwise prometheus-operator pods would not be able to start


pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -71,7 +71,7 @@ type Options struct {
 //+kubebuilder:rbac:groups=extensions;networking.k8s.io,resources=ingresses,verbs=get;list;watch
 
 // RBAC for delegating the use of SCC nonroot-v2 needed for OpenShift
```
simonpasquier, line 73:
```suggestion
// RBAC for delegating the use of SCC nonroot-v2 (for OpenShift >= 4.11) and nonroot (for OpenShift >= 4.12)
```

simonpasquier (approved):
lgtm

PR #164, opened 2022-06-16
fix: remove SeccompProfile

### LINKED ISSUE #163: SecurityContexts in MSO
**Background:**
Upon installing the `CatalogSource` and `Subscription` to deploy the Observability Operator, two deployments are created in the `openshift-operators` namespace: observability-operator, and observability-operator-prometheus-operator. 

**Problem:**
These deployments define `SecurityContext` that is blocked by the OpenShift Admission Controller(s).

output
```
 message: 'pods "observability-operator-prometheus-operator-75df4c6c7-9rql8" is
      forbidden: unable to validate against any security context constraint: [pod.metadata.annotations.seccomp.security.alpha.kubernetes.io/pod:
      Forbidden: seccomp may not be set pod.metadata.annotations.container.seccomp.security.alpha.kubernetes.io/prometheus-operator:
      Forbidden: seccomp may not be set provider "anyuid": Forbidden: not usable by
      user or serviceaccount pod.metadata.annotations.seccomp.security.alpha.kubernetes.io/pod:
      Forbidden: seccomp may not be set spec.containers[0].securityContext.runAsUser:
      Invalid value: 65534: must be in the ranges: [1000400000, 1000409999] pod.metadata.annotations.container.seccomp.security.alpha.kubernetes.io/prometheus-operator:
      Forbidden: seccomp may not be set provider "hostmount-anyuid": Forbidden: not
      usable by user or serviceaccount provider "machine-api-termination-handler":
      Forbidden: not usable by user or serviceaccount provider "hostnetwork": Forbidden:
      not usable by user or serviceaccount provider "hostaccess": Forbidden: not usable
      by user or serviceaccount provider "node-exporter": Forbidden: not usable by
      user or serviceaccount provider "privileged": Forbidden: not usable by user
      or serviceaccount]'
```

We feel strongly that `anyuid` should not be overwritten, as we will be running this on OSD clusters. This exposes us _somewhat_** and the premise of running on OCP, and especially OSD is the security aspect. 

We were using `anyuid` as a temporary solution but it has security ramifications, and it does not work anymore. 
What are the recommendations for this? Thanks

simonpasquier:
It seems that built-in SCCs (except `privileged`) don't allow to set the
seccomp profile [1]. Until this is fixed in OpenShift, we should leave
it to the default (which is already `RuntimeDefault`).

[1] https://github.com/openshift/cluster-kube-apiserver-operator/issues/1325

Fixes #163

### DIFF
--- a/deploy/dependencies/kustomization.yaml
+++ b/deploy/dependencies/kustomization.yaml
@@ -42,8 +42,6 @@ patches:
                 terminationMessagePolicy: FallbackToLogsOnError
             securityContext:
               runAsNonRoot: true
-              seccompProfile:
-                type: RuntimeDefault
   - patch: |-
       - op: remove
         path: /spec/template/spec/nodeSelector
--- a/deploy/operator/observability-operator-deployment.yaml
+++ b/deploy/operator/observability-operator-deployment.yaml
@@ -22,8 +22,6 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
-        seccompProfile:
-          type: RuntimeDefault
       containers:
         - name: operator
           image: observability-operator:0.0.1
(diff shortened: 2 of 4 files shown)


JoaoBraveCoding (approved):
LGTM! However I'll open an issue to track we comply with pod security admission on 4.12

PR #167, opened 2022-06-28
Remove startingCSV from subscription.yaml

leifmadsen:
Remove the startingCSV line from the subscription.yaml as it's not necessary and requires installation and updates in OLM to start at an older release rather than just installing the latest. This can have issues if the catalog source (index image) does not have the appropriate starting CSV release. (Additionally this blocked installation of observability-operator.v0.0.11 since v0.0.10 wasn't available in the index image, and resulted in an out of sync condition with the latest release.)


JoaoBraveCoding:
Thank you for the contribution ❤️ 

Could you please just update the commit message to be compliant with the [convention we require](https://github.com/rhobs/observability-operator#commit-message-convention)

leifmadsen:
> Thank you for the contribution heart
> 
> Could you please just update the commit message to be compliant with the [convention we require](https://github.com/rhobs/observability-operator#commit-message-convention)

Done

simonpasquier (approved):

simonpasquier:
:wave: the commit message would need an update
https://github.com/rhobs/observability-operator/runs/7109437423?check_suite_focus=true

JoaoBraveCoding:
Thanks for fixing it, I've enabled auto-merge for when the tests pass ❤️

simonpasquier (approved):

PR #168, opened 2022-06-30
fix: updates resources for OO and p-o

### LINKED ISSUE #166: Review resources set in components
Currently, our deployment of Prometheus Operator (p-o) set's the following resources:
```yaml
        resources:
          limits:
            cpu: 200m
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 100Mi
```
This is an extremely tight window of resources that p-o might hit easily, some users already reported to have hit OOM.
AFAIK we should never set limits on infrastructure components, so I suggest us removing completely the `limits`.

JoaoBraveCoding:
Problem: we pull the Prometheus Operator (p-o) deployment from
the upstream repo as a dependency. However this manifest sets
a very low limit to the p-o resources, this limit is easily hit
when the operator is managing multiple Prometheus instances.

Solution: remove current limits and load testing on OO and p-o
and observe the resources they consume. Establish a baseline for
both and then multiply that baseline by 3 and give some headroom

Issue https://issues.redhat.com/browse/MON-2648
Closes https://github.com/rhobs/observability-operator/issues/166

### DIFF
--- a/deploy/dependencies/kustomization.yaml
+++ b/deploy/dependencies/kustomization.yaml
@@ -39,6 +39,9 @@ patches:
                   requests:
                     cpu: 5m
                     memory: 150Mi
+                  limits:
+                    cpu: 100m
+                    memory: 500Mi
                 terminationMessagePolicy: FallbackToLogsOnError
             securityContext:
               runAsNonRoot: true
--- a/hack/loadtest/test.sh
+++ b/hack/loadtest/test.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -e -u -o pipefail
+
+# Functions that given a number it creates a namespace
+# and in that namespace it creates a monitoring stack
+create_monitoring_stack() {
+    stack_number=$1
+    MS_NAME=stack-$stack_number
+    NAMESPACE=loadtest-$stack_number
+
+    monitoring_stack=$(cat <<- EOF
+apiVersion: monitoring.rhobs/v1alpha1
+kind: MonitoringStack
+metadata:
+  name: ${MS_NAME}
+  namespace: ${NAMESPACE}
+  labels:
+    load-test: test
+spec:
+  logLevel: debug
+  retention: 15d
+  resourceSelector:
+    matchLabels:
+      load-test-instance: ${MS_NAME}
+EOF
+)
+
+    kubectl create namespace $NAMESPACE --dry-run=client -o yaml | kubectl apply -f -
+    echo "$monitoring_stack" | kubectl -n $NAMESPACE apply -f -
+
+}
+
+cleanup() { 
+    kubectl delete ns loadtest-{1..10}
+}
+
+main() {
+  # Goal: create 10 monitoring stack CRs, wait for OO to
+  # reconcile and then clean-up
+  for ((i=1; i<=10; i++)); do
+    create_monitoring_stack $i
+  done
+  # Give some time for OO to reconcile all the MS
+  # and create the necessary resources
+  sleep 180
+  
+  cleanup
+}
+ main "$@"
--- a/deploy/operator/kustomization.yaml
+++ b/deploy/operator/kustomization.yaml
@@ -16,3 +16,19 @@ images:
 - name: observability-operator
   newTag: 0.0.11
 namespace: operators
+
+patches:
+- patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/resources
+      value:
+        requests:
+          cpu: 5m
+          memory: 50Mi
+        limits:
+          cpu: 50m
+          memory: 150Mi
+  target:
+    group: apps
+    kind: Deployment
+    version: v1
--- a/docs/assess-resources.md
+++ b/docs/assess-resources.md
@@ -1,0 +1,42 @@
+
+# Procedure to assess resources used by Observability Operator
+
+1. Provision an OpenShift cluster
+
+2. Run `oc apply -f hack/olm/catalog-src.yaml` to install the Observability Operator (OO) catalogue.
+
+3. Using the UI install OO
+
+4. Scale down the following deployments, so we can remove the currently set limits on OO:
+
+```bash
+# Scale down the cluster version operator
+oc -n openshift-cluster-version scale deployment.apps/cluster-version-operator --replicas=0
+# Scale down the OLM operator
+oc -n openshift-operator-lifecycle-manager scale deployment.apps/olm-operator --replicas=0
+```
+
+5. Edit the OO and Prometheus Operator deployment to remove it's limits with:
+
+```bash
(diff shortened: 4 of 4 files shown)


simonpasquier (approved):
lgtm

should we do the same for the observability operator deployment?

apahim:
please do set proper resources. see comment in https://issues.redhat.com/browse/MON-2648

simonpasquier (changes requested):
Following Amador's comment, we need to increase limits, not removing them.

hack/loadtest/test.sh
```diff
@@ -0,0 +1,44 @@
+set -e
```
sthaha, line 1:
```suggestion
#! /usr/bin/env bash

set -eux -o pipefail
```

hack/loadtest/test.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Functions that given a number it creates a namespace
+# and in that namespace it creates a monitoring stack
+function create_monitoring_stack() {
```
sthaha, line 5:
```suggestion
create_monitoring_stack() {
```

hack/loadtest/test.sh
```diff
@@ -0,0 +1,44 @@
...
+    stack_number=$1
+    export MS_NAME=stack-$stack_number
+    export NAMESPACE=loadtest-$stack_number
+
+    monitoring_stack="apiVersion: monitoring.rhobs/v1alpha1
```
sthaha, line 10: it may be better to read from the samples, or a test-data directory.

hack/loadtest/test.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Give some time for OO to reconcile all the MS
+# and create the necessary resources
+sleep 180
+
+# Cleanup
+for ((i=1; i<=10; i++))
+do
+    kubectl delete ns loadtest-$i 
+done
```
sthaha, line 44:
```suggestion
cleanup() { 
    kubectl delete ns loadtest-{1..10}
}

main() {

  # Goal: create 10 monitoring stack CRs, wait for OO to
  # reconcile and then clean-up

  for ((i=1; i<=10; i++)); do
    create_monitoring_stack $i
  done

  # Give some time for OO to reconcile all the MS
  # and create the necessary resources
  sleep 180
  
  cleanup
}

 main "$@"
```

hack/loadtest/test.sh
```diff
@@ -0,0 +1,44 @@
+set -e
+
+# Functions that given a number it creates a namespace
+# and in that namespace it creates a monitoring stack
+function create_monitoring_stack() {
+    stack_number=$1
+    export MS_NAME=stack-$stack_number
```
sthaha, line 7: we don't need to export these

deploy/dependencies/kustomization.yaml
```diff
@@ -39,6 +39,9 @@ patches:
                   requests:
                     cpu: 5m
                     memory: 150Mi
+                  limits:
+                    cpu: 100m
+                    memory: 500Mi
```
sthaha, line 44: I am a bit confused why the PR which says remove limits from PO adds limits 🤔

hack/loadtest/test.sh
sthaha: I think it is better to use heredocs instead e.g. 

read -r -d '' VAR <<- EO_YAML
     apiVersion: 
     kind:
     metadata: ...
EO_YAML

docs/assessResources.md
```diff
@@ -0,0 +1,42 @@
+
+# Procedure to assess resources used by Observability Operator
```
sthaha, line 2: nice!

docs/assessResources.md
```diff
@@ -0,0 +1,42 @@
+
```
sthaha, line 1: Could we please keep all filenames lowercase?

docs/assessResources.md
```diff
@@ -0,0 +1,42 @@
...
+2. Run `oc apply -f hack/olm/catalog-src.yaml` to install the Observability Operator (OO) catalogue.
+
+3. Using the UI install OO
+
+4. Scale down the following deployments, so we can remove the currently set limits on OO:
+
+```bash
+# Scale down the cluster version operator
+oc -n openshift-cluster-version scale deployment.apps/cluster-version-operator --replicas=0
```
sthaha, line 14: should we scale down the CVO? Can't we edit the cluster-version to put OLM in unmanaged mode? 🤔

docs/assessResources.md
```diff
@@ -0,0 +1,42 @@
...
+# Scale down the OLM operator
+oc -n openshift-operator-lifecycle-manager scale deployment.apps/olm-operator --replicas=0
+```
+
+5. Edit the OO and Prometheus Operator deployment to remove it's limits with:
+
+```bash
+oc -n openshift-operators edit deployment.apps/observability-operator
```
sthaha, line 22: just a thought - can we use `patch`  instead?

sthaha:
FYI: I am looking e2e-test-olm failure #170

hack/loadtest/test.sh
JoaoBraveCoding: From StackOverflow:
> I think it's worth mentioning that if you have set -o errexit (a.k.a set -e) in your script and you use this then it will terminate your script because read returns a non-zero return code when it reaches EOF.

Still, I've updated it to use heredoc

hack/loadtest/test.sh
```diff
@@ -0,0 +1,50 @@
...
+spec:
+  logLevel: debug
+  retention: 15d
+  resourceSelector:
+    matchLabels:
+      load-test-instance: ${MS_NAME}
+EOF
+)
+
+    kubectl create namespace $NAMESPACE --dry-run=client -o yaml | kubectl apply -f -
```
sthaha, line 29: can't we `kubectl create ns $namespace` ?

sthaha:
@JoaoBraveCoding I felt running cleanup on SIGINT will be a good addition. PTAL - https://github.com/JoaoBraveCoding/observability-operator/pull/1 👼

simonpasquier (approved):
lgtm

simonpasquier:
> @JoaoBraveCoding Does this patch completely addresses https://issues.redhat.com/browse/MON-2648
> 
> Should there be a change for PO as well?

IIUC this PR updates the limits for both OBO and PO. If you mean updating the limits in https://github.com/prometheus-operator/prometheus-operator, the values are based on what is required to run the end-to-end tests.

JoaoBraveCoding:
> @JoaoBraveCoding Does this patch completely addresses https://issues.redhat.com/browse/MON-2648
> 
> Should there be a change for PO as well?

Yes, but that is also done in this PR in the file (`deploy/dependencies/kustomization.yaml`). I'll let you press merge @sthaha if everything looks good to you 😄

Issue #142, opened 2022-05-19
Reenable `reconcileRevertsManualChanges` test


jan--f:
Once the prometheus crd yaml manifest carries the expected atomic marker on `LabelSelector` fields, we should reenable this test. Until then we'll skip.

See also:
https://github.com/kubernetes-sigs/controller-tools/issues/676
<link to prometheus-operator issue>

jan--f:
This is fixed in https://github.com/kubernetes-sigs/controller-tools/pull/692 and via https://github.com/kubernetes-sigs/controller-tools/pull/693 shipped in [controller-tools v0.9.1](https://github.com/kubernetes-sigs/controller-tools/releases/v0.9.1).

PR #173, opened 2022-07-06
chore: update controller-gen to v0.9.2

jan--f:
This mostly pulls in many fixes from controller-tools. Mainly
controller-gen now respects type marker in types and generated the CRD
manifests accordingly. This means that SSA now functions correctly with
the generated manifests. Accordingly this also reenables the
`revert-test`.
The other changes added here are mostly whitespace changes in the
manifests, the status being no longer included in the CRD manifests (see
https://github.com/kubernetes-sigs/controller-tools/issues/456 for
details) and we seem to have forgotten to regenerate some manifests
after a version bump.

Fixes: https://github.com/rhobs/observability-operator/issues/142


simonpasquier (approved):

sthaha:
lgtm except for the broken CI 🤔

jan--f:
Ok the CRD manifest errors where due to a bug in controller-gen. v0.9.2 ships a fix for this.

simonpasquier (dismissed):

hack/kind/setup.sh
```diff
@@ -77,11 +77,23 @@ setup_cluster() {
...
 }
 
+install_kubectl() {
+    if ! command -v kubectl &> /dev/null
+then
+    info "kubectl not found, attempting to install"
+    curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+    chmod +x kubectl
+    export PATH="$(pwd):$PATH"
+fi
```
sthaha, line 87: Unrelated change :( 

Also ... IMHO, we shouldn't install tools / make changes to users system other than in locations known/announced to them. E.g. `make tools` is explicit and installs to `<prj>/tmp/bin`

sthaha (changes requested):
Requesting changes only due to an unrelated change (`hack/setup.sh`). 
Merging this would squash the commits (only policy enabled for this repo)

jan--f:
> Requesting changes only due to an unrelated change (`hack/setup.sh`). Merging this would squash the commits (only policy enabled for this repo)

Thanks good catch. I had that change in for testing. I'm actually proposing it in #175. Dropped it now.

sthaha (approved):

PR #175, opened 2022-07-08
chore: kind setup script, check for kubectl

jan--f:
And if not found install it.


hack/kind/setup.sh
```diff
@@ -77,11 +77,21 @@ setup_cluster() {
   kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout=300s
 }
 
+install_kubectl() {
+    if ! command -v kubectl &> /dev/null
+then
+    info "kubectl not found, attempting to install"
```
sthaha, line 83:
```suggestion
    if ! command -v kubectl >& /dev/null; then  # 2 corrections >& and nit:  ; then to follow the convention
      info "kubectl not found, attempting to install"
```

hack/kind/setup.sh
```diff
@@ -77,11 +77,21 @@ setup_cluster() {
...
+    curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
sthaha, line 84: nice! 
Where would this install kubectl to?

hack/kind/setup.sh
```diff
@@ -77,11 +77,21 @@ setup_cluster() {
...
+fi
+
+}
+
 main() {
   ## NOTE: all paths are relative to the root of the project
   cd "$PROJECT_ROOT_DIR"
-  mkdir -p tmp/logs $(dirname $KUBECONFIG)
+  mkdir -p tmp/logs "$(dirname $KUBECONFIG)"
```
sthaha, line 92: good find! 
Someone's got a ` ` in their path ;-)

hack/kind/setup.sh
jan--f: According to `man bash` `&>` is the preferred version of the otherwise equivalent versions.

hack/kind/setup.sh
jan--f: `kubectl` gets installed in the repo root with this. Hence the `PATH` addition (which I now also remembered to push).

I opted for this for now since I run this automated and it's nice if a script can just get its own dependency. I'd be up for a check before too, though I'd prefer to not have it cough up an error, require a manual step and then another run...at least not when the remedy is pretty straight forward and should work on all linux systems.

hack/kind/setup.sh
```diff
@@ -77,11 +77,22 @@ setup_cluster() {
   kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout=300s
 }
 
+install_kubectl() {
+    if ! command -v kubectl &> /dev/null; then
+    info "kubectl not found, attempting to install"
+    curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
sthaha, line 83: How about using tmp/bin instead of project root so that `git status` and `git add` doesn't include kubectl by mistake?

hack/kind/setup.sh
sthaha: wow! didn't know this was the preferred option.  Force of habit :)

sthaha (approved):

Issue #160, opened 2022-06-09
Expose ability to enable remote write receiver for Prometheus controller


leifmadsen:
**Request**

Enable the remote-write receiver functionality (https://prometheus.io/docs/prometheus/latest/feature_flags/#remote-write-receiver) for the Prometheus controller in Observability Operator

**Use Case**

Future versions of OpenStack will implement an `openstack observability` CLI command which will deploy Prometheus in Agent mode to locally scrape and remote-write data without providing local data storage or TSDB functionality. Storage of data from the cloud into a centralized data system (without the need to have tenants etc as this is an administrative domain environment) for the purposes of monitoring for anomalies in the OpenStack platform in order to alert administrators of issues and to provide basic graphing capabilities is the reason for this request.

simonpasquier:
:wave: IIUC you want use the Observability Operator to deploy a Prometheus instance that would receive metrics from Prometheus agents?

leifmadsen:
> wave IIUC you want use the Observability Operator to deploy a Prometheus instance that would receive metrics from Prometheus agents?

Correct. Instead of collecting metrics on OSP and transporting them via QDR and exposing them via Smart Gateway middleware for scraping by Prometheus on OCP, we want to scrape locally on OCP and remote-write them to another Prometheus for the storage domain.

PR #176, opened 2022-07-13
feature: add option to pass EnableRemoteWriteReceiver to Prometheus CR

jan--f:
Fixes: https://github.com/rhobs/observability-operator/issues/160


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -96,6 +96,9 @@ type PrometheusConfig struct {
 	// Define ExternalLabels for prometheus
 	// +optional
 	ExternalLabels map[string]string `json:"externalLabels,omitempty"`
+	// Enable Prometheus to be used as a receiver for the Prometheus remote write protocol. Defaults to the value of `false`.
+	// +optional
+	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
```
sthaha, line 101: shouldn't this be `*bool` so that `false` isn't omitted?

jan--f: Iiuc a pointer allows us to distinguish unset from `false`/empty. I don't think we need this here, do we?

jan--f: This is also consistent with [prometheus-operator's spec](https://github.com/prometheus-operator/prometheus-operator/blob/50ec1f57f19289cda613d484f3e66327bfeb3328/pkg/apis/monitoring/v1/types.go#L169).

sthaha: I tested this against kind and it actually works as expected. My understanding was that `bool` value of false is the zero value and if you applied the following 

```yaml
spec:
  prometheusConfig:
    enableRemoteWriteReceiver: false
```
The zero value of false will be omitted (just like json.Marshal does) and `kubectl get -o yaml monitoringstack foobar` will **not** print `enableRemoteWriteReceiver: false` but to the contrary, it does in fact print it. I think json marshalling should have been changed to accomodate this behaviour.

sthaha (approved):

PR #177, opened 2022-07-14
Internal build changes

jan--f:
Two changes to support internal build pipelines. This introduces no functional changes, only a new `build` target and the addition of OLM artifacts to the tracked files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,9 +94,11 @@ generate-deepcopy: $(CONTROLLER_GEN)
 .PHONY: generate
 generate: generate-crds generate-deepcopy generate-kustomize generate-prometheus-rules docs
 
-operator: generate
-	go build -o ./tmp/operator ./cmd/operator/...
+.PHONY: operator
+operator: generate build
 
+build:
+	go build -o ./tmp/operator ./cmd/operator/...
 
 .PHONY: operator-image
 operator-image: generate
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -69,6 +69,8 @@ jobs:
 
       - run: make --always-make generate && git diff --exit-code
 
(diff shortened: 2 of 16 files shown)


sthaha:
@jan--f what is the reason behind tracking the bundles which can be generated from source? 

If we need to track bundles, we need to also ensure that the bundles are up to date by adding another CI job like `Generate & format` that ensures that the bundle is kept in sync with code changes. 

Alternatively, if we want to track only when released, I think it is better do that in the automation we have that makes the release.

Makefile
```diff
@@ -92,11 +92,13 @@ generate-deepcopy: $(CONTROLLER_GEN)
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./pkg/apis/..."
 
 .PHONY: generate
-generate: generate-crds generate-deepcopy generate-kustomize generate-prometheus-rules docs
+generate: generate-crds generate-deepcopy generate-kustomize generate-prometheus-rules bundle docs
```
sthaha, line 95: doesn't this create a circular dependency? `generate -> bundle -> generate` ?

jan--f: Yup you are correct of course. Tbh I'm not exactly sure why I added this :thinking: 

Removed that dependency.

sthaha:
@jan--f , could you please also modify the CI job - (make generate) to instead call make bundle so that we keep generate bundle in sync with the code changes?

jan--f:
Ok I added a step to the `Generate and format` check. After this is merged we can go for a new release. This is all I need to test the internal build pipelines I think.

jan--f:
and it pointed out that I forgot to run `make bundle` :tada: 

We'll probably need to run make bundle as part of the release cutting too, since the version goes into the bundle resource..?

simonpasquier (approved):

sthaha (approved):

PR #178, opened 2022-07-14
chore: use Podman as the default container runtime

tremes:
This updates the Makefile to use Podman as the default container runtime.
If Podman is not available then fallback to docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ IMAGE_BASE ?= observability-operator
 
 VERSION ?= $(shell cat VERSION)
 OPERATOR_IMG = $(IMAGE_BASE):$(VERSION)
+CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 
 # running `make` builds the operator (default target)
 all: operator
@@ -100,11 +101,11 @@ operator: generate
 
 .PHONY: operator-image
 operator-image: generate
-	docker build -f build/Dockerfile . -t $(OPERATOR_IMG)
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
 
 .PHONY: operator-push
 operator-push:
-	docker push ${OPERATOR_IMG}
+	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) ${OPERATOR_IMG}
 
 .PHONY: test-e2e
 test-e2e:
@@ -153,11 +154,11 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
 
 .PHONY: bundle-image
 bundle-image: bundle ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	$(CONTAINER_RUNTIME) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
 
 .PHONY: bundle-push
 bundle-push: ## Build the bundle image.
-	docker push $(BUNDLE_IMG)
+	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) $(BUNDLE_IMG)
 
 # A comma-separated list of bundle images e.g.
 # make catalog-build BUNDLE_IMGS=example.com/operator-bundle:v0.1.0,example.com/operator-bundle:v0.2.0).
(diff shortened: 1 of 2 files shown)


tremes:
Hmm `podman push` is failing with following error when push to the local-registry:
```
Error: trying to reuse blob sha256:c8b72adf0d7c5590f429424d6ecaf56211b483a70340d8243ee49d779db8a3ba at destination: pinging container registry local-registry:30000: Get "https://local-registry:30000/v2/": http: server gave HTTP response to HTTPS client
```
So we have following options IMO:

-  use the `--tls-verify=false` flag
- provide the path to the certs 
- run the registry with http (similar to first option)

JoaoBraveCoding:
> `podman push` is failing with following error when push to the local-registry

I'm pretty sure the Makefile is used to publish official releases so I'm not a big fan of adding `--tls-verify=false` So I'm more in favour of either: 
- Runnig the local registry in HTTP mode, if possible.
- Mentioning in our README that when using `podman` the user should configure it like the following. Got this idea by reading [1].
```
[[registry]]
location="localhost:30000"
insecure=true
```

[1] https://www.redhat.com/sysadmin/manage-container-registries

sthaha:
> Mentioning in our README that when using podman the user should configure it like the following. Got this idea by reading [1].
```
[[registry]]
location="localhost:30000"
insecure=true
```

I feel setting the registry.conf works betters since we can do the same in CI as well. The other options is to explicitly pass an option to `make` e.g.
 `make <operator|bundle|catalog>-push INSECURE_REGISTRY=true` or 
`make <operator|bundle|catalog>-push PUSH_OPTION=--tls-verify=false`

jan--f:
Iiuc the `registries.conf` is respected when its either at `/etc/containers/registries.conf` or `$HOME/.config/containers/registries.conf`, so a user has to potentially change their existing configuration.

With that in mind I think I prefer the `make <operator|bundle|catalog>-push PUSH_OPTION=--tls-verify=false` approach. We have no file io to do in the automated tests and the command is explicit about what happens.

sthaha (dismissed):
good stuff :)

Makefile
```diff
@@ -100,11 +101,11 @@ operator: generate
 
 .PHONY: operator-image
 operator-image: generate
-	docker build -f build/Dockerfile . -t $(OPERATOR_IMG)
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
 
 .PHONY: operator-push
 operator-push:
-	docker push ${OPERATOR_IMG}
+	$(CONTAINER_RUNTIME) push $(PUSH_OPTION) ${OPERATOR_IMG}
```
jan--f, line 108: Apologies, one last nit. 
```suggestion
	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) ${OPERATOR_IMG}
```
Seems better? I think this can pass multiple options to push right? Otherwise lgtm

jan--f (changes requested):

Makefile
tremes: Ahh yeah. No problem. Let me update this because it's in multiple places.

jan--f (approved):
Thanks!

jan--f:
Hmm I've seen this error in multiple PRs now, seems like we have a race somewhere maybe?
```
2022-07-20T08:00:20Z	ERROR	controller.monitoringstack	Reconciler error	{"reconciler group": "monitoring.rhobs", "reconciler kind": "MonitoringStack", "name": "tq-ms-combo", "namespace": "e2e-tests", "error": "roles.rbac.authorization.k8s.io \"tq-ms-combo-prometheus\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227
```
Its usually resolved with a re-run.

jan--f:
> I'm trying to reproduce the CI error locally, but without success....and now it passed....hmm

Yeah as I said, it seems to be unrelated to this PR. I'll create an issue for this and we can go ahead with this PR here.

PR #184, opened 2022-07-26
feat: add API option to disable Alertmanager deployment

tremes:
This adds a new configuration option for the MonitoringStack CRD that
allows users to disable Alertmanager deployment.

This is a first draft. I would like to add:
- removing of other Alermanager's resources (service, role, etc.)
- some tests


pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -192,3 +195,24 @@ func (r *reconciler) getStack(ctx context.Context, req ctrl.Request) (*stack.Mon
...
+		if errors.IsNotFound(err) {
+			return
+		}
+		logger.Error(err, "failed to get Alertmanager ", "name", ms.Name)
+	}
+	err = r.k8sClient.Delete(ctx, &am)
+	if err != nil {
+		logger.Error(err, "failed to delete Alertmanager ", "name", ms.Name)
+	}
+	logger.Info("Successfully removed Alermanager", "name", ms.Name)
```
tremes, line 217: Lower case and typo....note to myself :)

simonpasquier (commented):
I think that we could have 2 types of reconciler:
* patchReconciler which patches an existing resource to the desired state (e.g. defaultReconciler as it exists today).
* deleteReconciler which ensures that a given resource doesn't exist in the cluster.

`stackComponentReconcilers()` would know which reconciler to use for which resource.

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -72,6 +72,10 @@ type MonitoringStackSpec struct {
 	// +optional
 	// +kubebuilder:default={replicas: 2}
 	PrometheusConfig *PrometheusConfig `json:"prometheusConfig,omitempty"`
+
+	// Disables Alertmanager deployment
+	// +optional
+	DisableAlertmanager bool `json:"disableAlertmanager,omitempty"`
```
sthaha, line 78: I think it is better to have an `AlertManagerConfig` and have `Disabled` under it. So that the spec reads
```
spec:
  alertmanager:
    disabled: true
```

tremes: I've thought about it and I think we can do it, but at the same time it seems like the only config option for the Alertmanager (but one never knows....).

sthaha: Yes, the idea is to namespace so that additional configs do not require changes to existing api.

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -511,12 +545,17 @@ func newAlerts(t *testing.T) *monv1.PrometheusRule {
 	return rule
 }
 
-func newMonitoringStack(t *testing.T, name string) *stack.MonitoringStack {
+func newMonitoringStack(t *testing.T, name string, disableAlertManager bool) *stack.MonitoringStack {
```
simonpasquier, line 548: (nit) to avoid grow the list of arguments indefinitely, I'd recommend using something like the functional options pattern.

```golang
func newMonitoringStack(t *testing.T, name string, options ...func(*stack.MonitoringStack)) *stack.MonitoringStack {
    ms := &stack.MonitoringStack{
    ...
    }
    for _, opt := range opts {
        opt(ms)
    }
    ...
    return ms
}
```

Then use it like

```
newMonitoringStack(t, name,
    func(ms *stack.MonitoringStack) {
        ms.Spec.AlertmanagerConfig.DisableAlertmanager = true
    },
)
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -168,6 +172,12 @@ type PrometheusConfig struct {
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
 }
 
+type AlertmanagerConfig struct {
+	// Disables Alertmanager deployment
+	// +optional
+	DisableAlertmanager bool `json:"disableAlertmanager,omitempty"`
```
simonpasquier, line 178: Do we need to repeat Alertmanager?

```suggestion
	Disabled bool `json:"disabled,omitempty"`
```

simonpasquier (commented):
I think that the operator should disable [Alertmanager configuration](https://github.com/rhobs/observability-operator/blob/93a144b0bea739752dfdfd7b4c6f59944ac9e6bd/pkg/controllers/monitoring/monitoring-stack/components.go#L191-L201) in Prometheus too.

test/e2e/framework/assertions.go
```diff
@@ -279,6 +280,24 @@ func (f *Framework) GetStackWhenAvailable(t *testing.T, name, namespace string)
 	return ms
 }
 
+func (f *Framework) AssertAlertmanagerRemoved(t *testing.T, name, namespace string) {
```
sthaha, line 283:
```suggestion
func (f *Framework) AssertAlertmanagerAbsent(t *testing.T, name, namespace string) {
```

test/e2e/framework/assertions.go
```diff
@@ -279,6 +280,24 @@ func (f *Framework) GetStackWhenAvailable(t *testing.T, name, namespace string)
...
+	}
+	err := wait.Poll(5*time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+		err := f.K8sClient.Get(context.Background(), key, &am)
+		if errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err == wait.ErrWaitTimeout {
+		t.Fatal(fmt.Errorf("alertmanager %s/%s was not removed", namespace, name))
```
sthaha, line 297: i
```suggestion
		t.Fatal(fmt.Errorf("alertmanager %s/%s is present when expected to be absent", namespace, name))
```

test/e2e/monitoring_stack_controller_test.go
sthaha: calling `newMonitoringStack(t, "foobar", false)` is confusing to read. [The infamous boolean trap!](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

while `newMontoringStack(t, "foobar", withoutAlertManager)` is clearer 
where `func withoutAlertManager(ms *stack.MonitoringStack) { ms.Spec.AlertmanagerConfig.Disabled = true }`

tremes:
> I think that the operator should disable [Alertmanager configuration](https://github.com/rhobs/observability-operator/blob/93a144b0bea739752dfdfd7b4c6f59944ac9e6bd/pkg/controllers/monitoring/monitoring-stack/components.go#L191-L201) in Prometheus too.

@simonpasquier  I guess it means to remove the spec and the endpoint right (there's no extra switch right, when looking at https://pkg.go.dev/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1#PrometheusSpec)?

pkg/apis/monitoring/v1alpha1/types.go
tremes: we don't :)

simonpasquier:
@tremes correct you need something like

```golang
prometheus := &monv1.Prometheus{
...
}

if !ms.Spec.AlertmanagerConfig.DisableAlertmanager {
    prometheus.Spec.Alerting = &monv1.AlertingSpec{
        ....
    }
} 
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -168,6 +172,12 @@ type PrometheusConfig struct {
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
 }
 
+type AlertmanagerConfig struct {
+	// Disables Alertmanager deployment
```
simonpasquier, line 176:
```suggestion
	// Disables the deployment of Alertmanager. 
	// Note that in this case, Prometheus doesn't send alerts. Disabling Alertmanager is mainly useful when using the monitoring stack in a collect-and-forward mode.
```

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -41,27 +42,57 @@ func defaultReconciler(component client.Object, ms *stack.MonitoringStack) recon
...
+
+	for _, amObj := range alertManagerObjects(ms, instanceSelectorKey, instanceSelectorValue) {
+		if ms.Spec.AlertmanagerConfig.Disabled {
+			reconcileFunctions = append(reconcileFunctions, removeReconciler(amObj, ms))
+		} else {
+			reconcileFunctions = append(reconcileFunctions, defaultReconciler(amObj, ms))
+		}
+	}
+
+	return reconcileFunctions
```
jan--f, line 95: Another option would be to have the reconcile function inspect `ms.Spec.AlertmanagerConfig.Disabled` similarly to what `prometheusPDBReconciler` is doing.
This would avoid treating the alertmanager resources differently in this location.
But that's nothing that prevents this from merging...

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -72,6 +72,10 @@ type MonitoringStackSpec struct {
 	// +optional
 	// +kubebuilder:default={replicas: 2}
 	PrometheusConfig *PrometheusConfig `json:"prometheusConfig,omitempty"`
+
+	// Define Alertmanager config
+	// +optional
+	AlertmanagerConfig *AlertmanagerConfig `json:"alertmanagerConfig,omitempty"`
```
simonpasquier, line 78: This avoids checking for nil values.

```suggestion
	AlertmanagerConfig AlertmanagerConfig `json:"alertmanagerConfig,omitempty"`
```

pkg/controllers/monitoring/monitoring-stack/components.go
simonpasquier: Or instead of `prometheusPDBReconciler`, the controller could decide to use `defaultReconciler` (replicas > 1) or `removeReconciler`.

tremes: Hmm so the suggestions are either have multiple functions and keep `stackComponentReconcilers` simple or have two functions (`defaultReconciler` and `removeReconciler` + possibly more in the futture) and have the logic in the `stackComponentReconcilers` right? 

I think both are possible...

jan--f: Just to offer a third version:
We could have a generic reconciler function that takes an additional argument: `removeReconciler(component client.Object, ms *stack.MonitoringStack, deleteCondition bool) reconcileFunction`.

I'm not the biggest fan of repeating blocks of
```
if condition1 {
		reconcileFunctions = append(reconcileFunctions, removeReconciler(obj1, ms))
} else {
	reconcileFunctions = append(reconcileFunctions, defaultReconciler(obj1, ms))
}
if condition2 {
		reconcileFunctions = append(reconcileFunctions, removeReconciler(obj2, ms))
} else {
	reconcileFunctions = append(reconcileFunctions, defaultReconciler(obj2, ms))
}
...
```
Is there and advantage of the above I'm not seeing?

jan--f: also to reiterate: I don't think this discussion needs to hold up this PR. I have a branch to put the reconcile logic in a module and use it in the thanos querier controller too.

tremes: > Is there and advantage of the above I'm not seeing?

I don't think so. If you have it prepared in your branch then I would leave it.

jan--f: See #185 for what I had in mind. I'd very much appreciate feedback.

With that in mind, I'm happy to move this forward and discuss the reconcile bit on the other PR.

sthaha: I feel the code reads cleaner if the decision making part moves to a separate function that returns the right reconciler to use. 

e.g. the code would read as 


```go
reconcilers =  []reconcileFunction {
   defaultReconciler(...),
   pdbReconciler(ms), 
}

reconcilers = append(reconcilers, ...alertmanagerReconcilers(ms))
return reconcilers
```
where `pdbReconciler` will return the `remove|default` reconciler
and `alertmanagerReconcilers` will return the appropiate reconcilers

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -41,27 +42,62 @@ func defaultReconciler(component client.Object, ms *stack.MonitoringStack) recon
 	}
 }
 
+func removeReconciler(component client.Object, ms *stack.MonitoringStack) reconcileFunction {
```
sthaha, line 45:
```suggestion
func deleteReconciler(component client.Object, ms *stack.MonitoringStack) reconcileFunction {
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -168,6 +172,12 @@ type PrometheusConfig struct {
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
 }
 
+type AlertmanagerConfig struct {
+	// Disables the deployment of Alertmanager.
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
```
sthaha, line 178: can we also set `default: false`?

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -72,6 +72,10 @@ type MonitoringStackSpec struct {
 	// +optional
 	// +kubebuilder:default={replicas: 2}
 	PrometheusConfig *PrometheusConfig `json:"prometheusConfig,omitempty"`
+
+	// Define Alertmanager config
+	// +optional
+	AlertmanagerConfig AlertmanagerConfig `json:"alertmanagerConfig,omitempty"`
```
sthaha, line 78:
```suggestion
	// +kubebuilder:default={disabled: false}
	AlertmanagerConfig *AlertmanagerConfig `json:"alertmanagerConfig,omitempty"`
```

pkg/controllers/monitoring/monitoring-stack/components.go
sthaha: To match `client.Delete`

pkg/apis/monitoring/v1alpha1/types.go
tremes: Is that required? It's the zero value.

simonpasquier: > Is that required? It's the zero value.

agreed, I don't feel we need to specify a default.

sthaha: We need to specify default so that it appears when `oc get -o yaml` an already applied resource where user has relied on the defaults. see: replicas e.g.

pkg/controllers/monitoring/monitoring-stack/components.go
tremes: So do you want me to update it? It seems unnecessary given the Jan's proposal (in different PR).

pkg/apis/monitoring/v1alpha1/types.go
tremes: OK

pkg/controllers/monitoring/monitoring-stack/components.go
tremes: OK

pkg/apis/monitoring/v1alpha1/types.go
tremes: OK

simonpasquier:
I'd be fine with merging this one and discussing the follow-up options in #185.

sthaha (dismissed):

sthaha (changes requested):
@tremes I had approved it and was going to merge when I realised that the commit-lint fails :(

Could you please fix and resubmit?

JoaoBraveCoding:
Merging it as per @sthaha comment

JoaoBraveCoding (approved):

pkg/controllers/monitoring/monitoring-stack/components.go
jan--f: I'm ok with leaving this as is.

jan--f (approved):

sthaha (approved):

PR #185, opened 2022-07-29
refactor: introduce common componentUpdater module

jan--f:
This provides default component update functions that can be used across
controllers. For more complex update functions controllers can still
create custom functions.

### DIFF
--- a/pkg/componentUpdater/componentUpdater.go
+++ b/pkg/componentUpdater/componentUpdater.go
@@ -1,0 +1,41 @@
+package componentUpdater
+
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type UpdateFunction func(ctx context.Context, c client.Client, scheme *runtime.Scheme) error
+
+func DefaultUpdater(component client.Object, controller metav1.Object) UpdateFunction {
+	return func(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
+		if controller.GetNamespace() == component.GetNamespace() {
+			if err := controllerutil.SetControllerReference(controller, component, scheme); err != nil {
+				return err
+			}
+		}
+
+		if err := c.Patch(ctx, component, client.Apply, client.ForceOwnership, client.FieldOwner(fmt.Sprintf("%s/%s", controller.GetNamespace(), controller.GetName()))); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func DefaultDeleteOrUpdate(component client.Object, controller metav1.Object, deleteCondition bool) UpdateFunction {
+	if deleteCondition {
+		return func(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
+			if err := c.Delete(ctx, component); client.IgnoreNotFound(err) != nil {
+				return err
+			}
+			return nil
+		}
+	} else {
+		return DefaultUpdater(component, controller)
+	}
+}
--- a/pkg/controllers/monitoring/thanos-querier/controller.go
+++ b/pkg/controllers/monitoring/thanos-querier/controller.go
@@ -30,7 +30,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -105,14 +104,9 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, err
 	}
 
-	components := thanosComponents(querier, sidecarServices)
-	componentLabels := map[string]string{
-		"app.kubernetes.io/instance":   querier.Name,
-		"app.kubernetes.io/part-of":    "ThanosQuerier",
-		"app.kubernetes.io/managed-by": "observability-operator",
-	}
-	for _, c := range components {
-		err := r.reconcileComponent(ctx, querier, ensureLabels(c, componentLabels))
+	reconcilers := thanosComponentReconcilers(querier, sidecarServices)
+	for _, reconciler := range reconcilers {
+		err := reconciler(ctx, r, r.scheme)
 		// handle creation / updation errors that can happen due to a stale cache by
 		// retrying after some time.
 		if apierrors.IsAlreadyExists(err) || apierrors.IsConflict(err) {
@@ -187,34 +181,4 @@ func (r reconciler) findQueriersForMonitoringStack(ms client.Object) []reconcile
 		}
 	}
 	return requests
-}
-
-// Create an object after setting the owner reference to the passed
-// ThanosQueriers.
-// Uses server-side-apply.
-func (r reconciler) reconcileComponent(ctx context.Context, thanos *msoapi.ThanosQuerier, component client.Object) error {
-	if thanos.Namespace == component.GetNamespace() {
-		if err := controllerutil.SetControllerReference(thanos, component, r.scheme); err != nil {
-			return err
-		}
-	}
-
-	if err := r.Patch(ctx, component, client.Apply, client.ForceOwnership, client.FieldOwner("thanos-querier-controller")); err != nil {
-		return err
-	}
-	return nil
-}
-
-func ensureLabels(obj client.Object, wantLabels map[string]string) client.Object {
-	existingLabels := obj.GetLabels()
-	if existingLabels == nil {
-		obj.SetLabels(wantLabels)
-		return obj
-	}
-	for name, val := range wantLabels {
-		if _, ok := existingLabels[name]; !ok {
-			existingLabels[name] = val
-		}
-	}
-	return obj
 }
--- a/pkg/controllers/monitoring/monitoring-stack/components.go
+++ b/pkg/controllers/monitoring/monitoring-stack/components.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"reflect"
 
+	"github.com/rhobs/observability-operator/pkg/componentUpdater"
+
 	stack "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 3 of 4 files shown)


pkg/componentUpdater/componentUpdater.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		if err := c.Patch(ctx, component, client.Apply, client.ForceOwnership, client.FieldOwner(fmt.Sprintf("%s/%s", controller.GetNamespace(), controller.GetName()))); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func DefaultDeleteOrUpdate(component client.Object, controller metav1.Object, deleteCondition bool) UpdateFunction {
```
sthaha, line 30: I have an idea (un tested) for better readability 

```go

func For(component client.Object, controller metav1.Object,  conditions ...Conditions) UpdateFunction {
  if len(conditions) == 0 {
     return patcher(component, controller)
  }

  return conditions[0].Updater
}
```

lets us call 
```go
reconciler.For(newFoobar(), ms)
reconciler.For(newFoobar(), ms,  reconciler.DeleteIf{ ms.Spec.Replica >= 1 } )
reconciler.For(newFoobar(), ms,  reconciler.CreateIf{ ms.Spec.Bar >= 1 } )

```

where 
```go
func(d DeleteIf) Updater() UpdateFunction {
   if d.Condition {
     return deleteReconciler
   }
   return patcher
}
```

tremes: Yeah I think it's something similar as proposed in the https://github.com/rhobs/observability-operator/pull/184#discussion_r931148861 and I like that approach. Not sure about naming though.....

pkg/componentUpdater/componentUpdater.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type UpdateFunction func(ctx context.Context, c client.Client, scheme *runtime.Scheme) error
```
simonpasquier, line 13: (nit) I think that instead of a function type, an interface might be more idiomatic

```golang
type Reconciler interface {
    Reconcile(context.Context, client.Client, *runtime.Scheme) error
}

// UpdateReconciler ensures that a resource exists with the desired state.
type UpdateReconciler struct {
    ms *stack.MonitoringStack
    resource client.Object
}

func (r *UpdateReconciler) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
    if ms.Namespace == component.GetNamespace() {
    ...
    return nil
}

// DeleteReconciler ensures that a resource doesn't exist.
type DeleteReconciler struct {
    resource client.Object
}

func (r *DeleteReconciler) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
    if err := c.Delete(ctx, r.resource); client.IgnoreNotFound(err) != nil {
    ...
}
```

We could have specialized reconcilers depending on the monitoring stack defintion:

```golang
// Executes a reconciler based on conditions.
type AlertmanagerReconciler struct {
   r Reconciler
}

func NewAlertmanagerReconciler(ms *stack.MonitoringStack) *AlertmanagerReconciler {
  var r Reconciler
  if ms.Spec.Alertmanager.Disabled {
    r = &DeleteReconciler{...}
  } else {
    r = &UpdateReconciler{...}
  }
  return &AlertmanagerReconciler{r: r}
}

func (r *AlertmanagerReconciler) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
    return r.r.Reconcile(ctx, c, scheme)
}
```

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,75 @@
...
+	if cond {
+		return OptionalResourceReconciler{
+			r: NewUpdateReconciler(r, c),
+		}
+	} else {
+		return OptionalResourceReconciler{
+			r: NewDeleteReconciler(r),
+		}
+	}
+}
```
jan--f, line 75: I went with this name after some time banging my head against the wall. I think this is better than the other options because this reconciler is really for optional components. That they get deleted in some cases is an implementation details.
Consequently the condition that is passed is expected to be true if the optional resource should be created. This reverses the options we have discussed so far, but I think this makes more sense.

pkg/componentUpdater/componentUpdater.go
jan--f: This is a good point. Even if an interface isn't strictly necessary (it has only one function) it is more idiomatic and can improve "testability".

pkg/componentUpdater/componentUpdater.go
jan--f: While I very much like syntax that is readable and conveys intention, I would probably argue that the go proverb [Clear is better than clever](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s) takes precedence here.
If we'd write a library for many users, additional cleverness might be warranted, but here we just want to create some resources.

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,75 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type Reconciler interface {
+	Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error
+}
+
+type UpdateReconciler struct {
```
sthaha, line 17: I think we can avoid stuttering by omit `Reconciler` e.g.`reconciler.NewUpdateReconciler` would read `reconciler.Updater`

pkg/componentUpdater/componentUpdater.go
sthaha: For our usecase, I agree it shouldn't matter much at all. It is an internal implementation detail. 

Having said that .. 
I am not sure I would put it under `too clever` category. My take on that talk is that these are good guidelines (not rules) to follow and can be broken. 

I feel the intention is clear and follows a very simple interface - A condition returns the method to reconcile.


Contrast this `reconciler.NewOptionalReconciler(foobar, ms, ms.Spec.Bar >= 1)` over `reconciler.New(foobar, ms, reconciler.DeleterIf{  ms.Spec.Bar >= 1 })` I would argue that the former requires you to understand the implementation detail of the `NewOptionalReconciler` while the latter does not (the implementation detail doesn't matter). 

But like I said, it is an internal implementation detail ...

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -42,7 +42,7 @@ import (
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 )
 
-type reconciler struct {
+type msController struct {
```
sthaha, line 45: Since this isn't a controller - could we rename it back to `reconciler` ? 

see : NewControllerManagedBy which returns a `Builder` whose build method returns a `controller.Controller`

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,75 @@
...
+	controller metav1.Object
+	resource   client.Object
+}
+
+func (r UpdateReconciler) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
```
sthaha, line 22: I think it would be nice for all `Reconcile` to return `ctrl.Result, error` so that each reconcile method can send back to the caller that the request just needs to be requeued.

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,75 @@
...
+
+type OptionalResourceReconciler struct {
+	r Reconciler
+}
+
+func (r OptionalResourceReconciler) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
+	return r.r.Reconcile(ctx, c, scheme)
+}
+
+func NewOptionalResourceReconciler(r client.Object, c metav1.Object, cond bool) OptionalResourceReconciler {
```
sthaha, line 65: Since the function already decides which reconciler to use, can't we  remove the need for an Optional reconciler and instead return the concrete reconciler itself ? e.g.

```
func NewDeleterOrUpdater(cond bool, obj client.Object, ctrl metav1.Object) Reconciler {
   if cond {
     return Deleter{obj}
   }
   return Updater{obj, ctrl}
}
```
The caller would call 

```
reconciler.NewDeleterOrUpdater(ms.Spec.Replicas >= 2, newAlertmanager(), ctrl)
```

pkg/reconciler/reconciler.go
jan--f: Right now the reconciler doesn't know anything about how `controller-runtime` (re)schedules events. To me that decision belongs in the controller.
Logging happens there, as well as deciding how long to hold up the requeued event. That is nothing the reconciler needs to have knowledge about.

pkg/controllers/monitoring/monitoring-stack/controller.go
jan--f: The motivation was to not have several `reconciler` names that do different things. But I agree, `*Controller` is also not great.

Maybe the component that is added here would be better named a `ResourceManager`? That would be consistent with https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/ at least?

pkg/reconciler/reconciler.go
simonpasquier: I agree with @sthaha, it feels simpler.

pkg/controllers/monitoring/monitoring-stack/controller.go
simonpasquier: resourceManager sounds good to me

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,75 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type Reconciler interface {
+	Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error
+}
+
+type UpdateReconciler struct {
+	controller metav1.Object
```
simonpasquier, line 18: I feel that `controller` is a bit overloaded.

```suggestion
	controlledBy metav1.Object
```

pkg/controllers/monitoring/thanos-querier/controller.go
```diff
@@ -65,7 +64,7 @@ type reconciler struct {
 // RegisterWithManager registers the controller with Manager
 func RegisterWithManager(mgr ctrl.Manager) error {
 	logger := ctrl.Log.WithName("thanos-querier")
-	r := &reconciler{
+	r := &thanosController{
```
simonpasquier, line 67: (nit) we might want to replace all `r` variables with `tc` or `c`.

simonpasquier (commented):
This looks very good! Only had a few comments.

pkg/reconciler/reconciler.go
sthaha: I kind of agree with on the requirement to remove coupling however, I do feel that the reconcile method should be able to not just `Patch` or `Delete` but also be able to communicate back to the "controller / caller" that the request need to be retried without an error ( i.e. the controller should return `Result{Requeue : true, After 5 * time.Second }`.  While in other cases return an error just how it does now. 

I am open to changing this in future when such a requirement arise.

pkg/controllers/monitoring/monitoring-stack/controller.go
sthaha: `Patcher | Updater` ? has a `Patch|Update`

jan--f:
Ok I addressed the review comments and tried to clarify things with better names and comments. wdyt?

pkg/reconciler/reconciler.go
simonpasquier: I'd agree with @sthaha that `reconciler.NewUpdateReconciler` is a bit verbose. `reconciler.Updater` and `reconciler.Deleter` sound good to me.
Not sure how to rename `NewOptionalResourceReconciler()` though. Maybe it can stay the same?

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,69 @@
...
+		return err
+	}
+	return nil
+}
+
+func NewDeleteReconciler(r client.Object) DeleteReconciler {
+	return DeleteReconciler{resource: r}
+}
+
+func NewOptionalResourceReconciler(r client.Object, c metav1.Object, cond bool) Reconciler {
```
simonpasquier, line 63:
```suggestion
// NewOptionalResourceReconciler ensures that a resource is present or absent depending on the `cond` value (true: present).
func NewOptionalResourceReconciler(resource client.Object, owner metav1.Object, cond bool) Reconciler {
```

simonpasquier (commented):
A few minor comments. I'm not decided if `NewOptionalResourceReconciler()` is needed or if the caller should instantiate the correct reconciler based on the condition but that's a very minor concern that shouldn't block the PR.

pkg/reconciler/reconciler.go
jan--f: Thanks! Added.

jan--f:
How about `NewOptionalUpdater`? With the comment you suggested I think this name makes sense.

simonpasquier (approved):
lgtm

PR #186, opened 2022-08-02
feat: add health probe

tremes:
This adds basic health probe endpoint (using the controller-runtime
options) and exposes the port via the existing OBO service

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -29,14 +29,16 @@ import (
 
 func main() {
 	var (
-		namespace   string
-		metricsAddr string
+		namespace       string
+		metricsAddr     string
+		healthProbeAddr string
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
 	opts := zap.Options{
 		Development: true,
 		TimeEncoder: zapcore.RFC3339TimeEncoder,
@@ -50,7 +52,7 @@ func main() {
 		"namespace", namespace,
 		"metrics-bind-address", metricsAddr)
 
-	op, err := operator.New(metricsAddr)
+	op, err := operator.New(metricsAddr, healthProbeAddr)
 	if err != nil {
 		setupLog.Error(err, "cannot create a new operator")
 		os.Exit(1)
--- a/deploy/operator/observability-operator-deployment.yaml
+++ b/deploy/operator/observability-operator-deployment.yaml
@@ -45,5 +45,13 @@ spec:
             requests:
               cpu: 100m
               memory: 150Mi
+          readinessProbe:
+             httpGet:
+               path: /healthz
+               port: 8081
+          livenessProbe:
+             httpGet:
+               path: /healthz
+               port: 8081
       serviceAccountName: observability-operator-sa
       terminationGracePeriodSeconds: 30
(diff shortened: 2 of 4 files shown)


cmd/operator/main.go
```diff
@@ -50,7 +50,7 @@ func main() {
 		"namespace", namespace,
 		"metrics-bind-address", metricsAddr)
 
-	op, err := operator.New(metricsAddr)
+	op, err := operator.New(metricsAddr, ":9090")
```
tremes, line 53: Do we want it as a config option as well? And what about the default value? Is `9090` OK?

pkg/operator/operator.go
```diff
@@ -39,6 +41,11 @@ func New(metricsAddr string) (*Operator, error) {
 		return nil, fmt.Errorf("unable to register the thanos querier controller with the manager: %w", err)
 	}
 
+	ok := func(_ *http.Request) error { return nil }
```
tremes, line 44: Do we want to do anything extra with the request here?

pkg/operator/operator.go
```diff
@@ -39,6 +41,11 @@ func New(metricsAddr string) (*Operator, error) {
...
+	if err := mgr.AddHealthzCheck("health probe", ok); err != nil {
```
tremes, line 45: I guess we want to add some `ready` check as well. I am wondering what the difference would be.....

cmd/operator/main.go
simonpasquier: An argument would be handy. I can't think of a reason for changing the default but at least we'll be safe. Maybe pick 8081 (instead of 9090) to follow 8080.

pkg/operator/operator.go
simonpasquier: for now, the service is always (healthy/ready) or (not healthy/not ready) so I don't think we need to differentiate and implementing only the health handler is fine.

pkg/operator/operator.go
simonpasquier: I think it's ok. We could even use https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.3/pkg/healthz#Ping

sthaha (commented):
Since this is  new feature, it would be nice to if it can be tested as well. 
WDYT? @simonpasquier @jan--f

tremes:
@sthaha that's what I am working on :)

deploy/operator/observability-operator-service.yaml
```diff
@@ -13,3 +13,5 @@ spec:
   ports:
   - name: metrics
     port: 8080
+  - name: healthz
```
simonpasquier, line 16: Do we need to expose the healthz port?

simonpasquier:
> Since this is new feature, it would be nice to if it can be tested as well.

I suppose that if we add liveness/readiness probes to the OBO deployment, it will be tested automatically?

deploy/operator/observability-operator-service.yaml
tremes: My thinking was that it had to be reachable within the cluster. Is that wrong assumption?

simonpasquier: the probe will only be used by the kubelet, it doesn't have to be exposed in the cluster.

tremes: Ahh OK didn't know that. I'll remove it and then I guess it brings up the question about testability of this.

sthaha:
> I suppose that if we add liveness/readiness probes to the OBO deployment, it will be tested automatically?

indeed! the operator in a ready state is a good test!

pkg/operator/operator.go
tremes: Updated.

pkg/operator/operator.go
tremes: OK thanks.

deploy/operator/observability-operator-deployment.yaml
```diff
@@ -45,5 +45,13 @@ spec:
...
               cpu: 100m
               memory: 150Mi
+          readinessProbe:
+             httpGet:
+               path: /healthz
+               port: 8081
+          livenessProbe:
+             httpGet:
+               path: /healthz
+               port: 8081
```
tremes, line 55: Do we want both? I guess so

simonpasquier: yes for me

simonpasquier (approved):
lgtm

sthaha (approved):

PR #188, opened 2022-08-05
MON-2677: chore: add app-interface artifact

jan--f:
First version based on an internal testing template and
https://github.com/openshift/addon-operator/blob/main/hack/olm-registry/olm-artifacts-template.yaml

### DIFF
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -1,0 +1,58 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: selectorsyncset-template
+
+parameters:
+  - name: REGISTRY_IMG
+    required: true
+    value: quay.io/rhobs/observability-operator-catalog
+  - name: CHANNEL
+    value: development
+    required: true
+  - name: IMAGE_DIGEST
+    required: true
+    value: eee088e78ecbaafdc572188398ceb463e63c365f1023a0301742be2a8e0e4c29
+
+objects:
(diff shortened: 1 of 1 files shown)


sthaha (commented):
@jan--f , Looks magic to me 🪄 

Could you please link to a JIRA (if that exists) and add doc on how to update this? 
Additionally, should this be added to bundle?

jan--f:
Magic indeed :)

So iiuc this specifies a [`SelectorSyncSet`](https://github.com/openshift/hive/blob/master/docs/syncset.md), which gives `hive` a list of resources to maintain across clusters.

In other words this is a sort of manifest that lists all resources (and potentially patches to related resources) that OBO needs in order to run on a cluster. We'll probably want to automate the creation of this based on our various `kustomization.yaml` files.

I think we'll need to following:

- [ ] The operator resources referenced in `deploy/operator/kustomization.yaml`. (I guess we can exclude the rule definitions for now)
- [ ] The dependencies referenced in `deploy/dependencies/kustomization.yam`.
- [ ] (Maybe) the CRD manifests referenced in `deploy/crds/kubernetes/kustomization.yaml`.

I'll start the work on automating this.

jan--f:
ok I think this is ready. This version is mostly a copy of what AppSRE has tested with already. I only added a couple of parameters so the catalog resource, image digest and channel can be changed.

sthaha (approved):
✨ 🪄

PR #189, opened 2022-09-07
fix: do not report Available=degraded condition

tremes:
The convention is to report only true,false or unknown as the condition status value.

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/types.go
+++ b/pkg/apis/monitoring/v1alpha1/types.go
@@ -96,10 +96,9 @@ type ConditionStatus string
 type ConditionType string
 
 const (
-	ConditionTrue     ConditionStatus = "True"
-	ConditionDegraded ConditionStatus = "Degraded"
-	ConditionFalse    ConditionStatus = "False"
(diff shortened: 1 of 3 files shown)


simonpasquier (commented):
I wonder if we shouldn't have different Reasons depending on the Prometheus status. E.g.
* Prometheus(Condition=Available,Status=Degraded) -> MonitoringStack(Condition=Available,Status=False,Reason=PrometheusDegraded)
* Prometheus(Condition=Available,Status=False) -> MonitoringStack(Condition=Available,Status=False,Reason=PrometheusNotAvailable)

sthaha (changes requested):
Could the commit message please include link to the Jira and a bit of reasoning behind this change? 
 
E.g.

```
fix: report Available=false when prometheus is degraded

This PR changes the status.condition to adhere to the convention to
report a condition as  true, false, or unknown. 

jira: link

signed-by: 
```

simonpasquier (approved):
lgtm

sthaha (approved):
/lgtm 

Thanks you @tremes for the patch 🙇

PR #190, opened 2022-09-07
Add logging for reconcile errors

csibbitt:
* Needed this to help debug why my prometheus was not deployed

### DIFF
--- a/pkg/controllers/monitoring/monitoring-stack/controller.go
+++ b/pkg/controllers/monitoring/monitoring-stack/controller.go
@@ -145,35 +145,32 @@ func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
 		if err != nil {
-			return rm.updateStatus(ctx, req, ms, err)
+			return rm.updateStatus(ctx, req, ms, err), err
 		}
 	}
 
-	return rm.updateStatus(ctx, req, ms, nil)
+	return rm.updateStatus(ctx, req, ms, nil), nil
 }
 
-func (rm resourceManager) updateStatus(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack, recError error) (ctrl.Result, error) {
+func (rm resourceManager) updateStatus(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack, recError error) ctrl.Result {
 	var prom monv1.Prometheus
+	logger := rm.logger.WithValues("stack", req.NamespacedName)
 	key := client.ObjectKey{
 		Name:      ms.Name,
 		Namespace: ms.Namespace,
 	}
 	err := rm.k8sClient.Get(ctx, key, &prom)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
-		}
-		return ctrl.Result{}, err
+		logger.Info("Failed to get prometheus object", "err", err)
+		return ctrl.Result{RequeueAfter: 2 * time.Second}
 	}
 	ms.Status.Conditions = updateConditions(ms.Status.Conditions, prom.Status.Conditions, ms.Generation, recError)
 	err = rm.k8sClient.Status().Update(ctx, ms)
 	if err != nil {
-		if errors.IsConflict(err) {
-			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
-		}
-		return ctrl.Result{}, err
+		logger.Info("Failed to update status", "err", err)
+		return ctrl.Result{RequeueAfter: 2 * time.Second}
 	}
-	return ctrl.Result{}, nil
+	return ctrl.Result{}
 }
 
 func (rm resourceManager) getStack(ctx context.Context, req ctrl.Request) (*stack.MonitoringStack, error) {


pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -145,6 +145,7 @@ func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
 		if err != nil {
+			logger.Error(err, "failed to reconcile")
```
sthaha, line 148: IIUC, the default behaviour of the controller-runtime reconcile loop is to log any error is `Reconcile` returns it. Do you not see these errors in logs?

csibbitt: I did not see any errors in the logs without this code.

My repro is like this:

- Deploy OCP 4.8 w/ Cluster Monitoring Operator
- Deploy ObO
- Create a MonitoringStack with `enableRemoteWriteReceiver: true`
- Observe that no Prometheus is created
- Check logs on ObO[1] and the ObO Prometheus Operator - they give no indication of a problem
- Wait a long time (over the weekend) and never see any log messages or Prometheus appear
- Apply this patch and redeploy
- Observe that there is an error in reconciling [2]

[1] It just repeats this over and over:
```
2022-09-06T16:14:23Z    INFO    observability-operator  Reconciling monitoring stack    {"stack": "observability-operator/sample-monitoring-stack"}
```
[2] New, helpful error message:
```
2022-09-06T16:54:59Z    ERROR   observability-operator  failed to reconcile      {"stack": "observability-operator/sample-monitoring-stack", "error": "failed to create typed patch object (observability-operator/sample-monitoring-stack; monitoring.coreos.com/v1, Kind=Prometheus): .spec.enableRemoteWriteReceiver: field not declared in schema"}
```

I'm also somewhat surprised that the framework did not report errors which appear to be returned (I didn't read further to figure out why); are they logged somewhere else? Please give my reproduction a test, or find another way to fail the reconciliation and let me know if I missed something, but I spent hours with no error message to follow up on and this helped me spot the problem (out of date CRD) in seconds.

sthaha: `updateStatus` being passed an `err` seems quite odd 🤔

sthaha: @csibbitt would you mind fixing the `updateStatus` to log the `recError` as an `INFO` ?

sthaha (changes requested):

pkg/controllers/monitoring/monitoring-stack/controller.go
simonpasquier: I think that in this case, we want to return the reconciliation error instead of the status update error.
 
```
if err != nil {
    statusErr = rm.updateStatus(ctx, req, ms, err)
    if statusErr != nil {
        logger.Error(statusErr, "Failed to update status")
    }
    return err
}
```

csibbitt: @sthaha It's possible that the error was reflected in the status of the MonitoringStack object and I simply missed it. I think logging is still a good idea.

I appreciate the effort going into making sure we do this the best way possible. I think Simon's latest suggestion makes the most sense if it results in displaying the error in the logs like we think it will. (though it might need to say `return ctrl.Result{}, err` I'll try this out locally when I'm done with my current experiments. :)

sthaha: Given that updateStatus can fail due to various reasons, (e.g. Get fails, updating status fails,) the error it returns should the reason for failure and log the `recError` which is passed to the function. 

WDYT @simonpasquier

jan--f: I think I agree with Simon. We should make sure the `Reconcile` function returns errors that are caused by the reconciliation logic.
I'd go as far as not `return rm.updateStatus(ctx, req, ms, err)` from `Reconcile` but to log these errors and return errors that pertain to reconciliation. The argument being that an error in updating the status is not tightly related to reconciliation of resources.

sthaha: I see your point but consider 2 places where updateStatus is called. 
```
r.updateStatus(ctx, req, ms, err) // when there is a know reconcile error
r.updateStatus(ctx, req, ms, nil) // when everything ran fine
```
In the first case if we returned the err, then operator will try and reconcile until everything runs fine (or for every) as opposed to current behaviour where operator logs failure in the status and only retries if the status couldn't be updated. 

Now, in the second case .. say there  was an issue updating the status, should the operator retry and update the status again? If we returned the `recError -> nil`, then operator won't retry, will it? 

How about we return recError if not nil else the error from statusUpdate?

simonpasquier: My initial idea was not to change the happy path (e.g. return error from updateStatus if any) but I feel like it makes the Reconcile() method less readable (sometimes updateStatus error is propagated to the caller, sometimes it's not).
Since `updateStatus` already returns a `crtl.Result` value, what if it didn't return any additional error but rather log errors caught when getting/updating the status and a "retry later" result?

```
func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
        ...
        for _, reconciler := range reconcilers {
                err := reconciler.Reconcile(ctx, rm.k8sClient, rm.scheme)
                // handle create / update errors that can happen due to a stale cache by
                // retrying after some time.
                        logger.V(3).Info("skipping reconcile error", "err", err)
                        return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
                }
                if err != nil {
                        return rm.updateStatus(ctx, req, ms, err), err
                }
        }

        return rm.updateStatus(ctx, req, ms, nil), nil
}

func (rm resourceManager) updateStatus(ctx context.Context, req ctrl.Request, ms *stack.MonitoringStack, recError error) ctrl.Result {
        ...
        err = rm.k8sClient.Status().Update(ctx, ms)
        if err != nil {
                logger.Info(statusErr, "Failed to update status")
                return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
        }
        return ctrl.Result{}, nil
}
```

sthaha: I like that @simonpasquier 🤗

csibbitt: I'll finish this up based on Simon's recommendation. Thanks all.

csibbitt:
I think this takes into account all the feedback, and the error logs are quite nice:

```
2022-09-22T21:18:41Z    INFO    observability-operator  Reconciling monitoring stack    {"stack": "openshift-operators/sample-monitoring-stack"}                                         
2022-09-22T21:18:41Z    INFO    observability-operator  Failed to get prometheus object {"stack": "openshift-operators/sample-monitoring-stack", "err": "Prometheus.monitoring.coreos.com
 \"sample-monitoring-stack\" not found"}                                                
2022-09-22T21:18:41Z    ERROR   controller.monitoringstack      Reconciler error        {"reconciler group": "monitoring.rhobs", "reconciler kind": "MonitoringStack", "name": "sample-mo
nitoring-stack", "namespace": "openshift-operators", "error": "failed to create typed patch object (openshift-operators/sample-monitoring-stack; monitoring.coreos.com/v1, Kind=Prometheu
s): .spec.enableRemoteWriteReceiver: field not declared in schema"}                                                                                                                      
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem                                                        
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266                                                                                     
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227
```

I have not yet been able to test the working path on a newer OCP or @sthaha 's rhobs-prometheus-operator-0.58.0 branch, but it seems pretty safe to me.

simonpasquier (dismissed):
lgtm

sthaha (dismissed):
Thanks for the fix @csibbitt 🙇

sthaha (approved):
Thanks again, @csibbitt ! Merging the patch this time as the CI gods have given their blessings :D

PR #191, opened 2022-09-13
feat: switch to file-based OLM catalogs

sthaha:
In order to switch to new file-based OLM catalogs, following changes were made

  * upgrade to newer version of operator-sdk tools (operator-sdk, olm)
  * Generate existing bundles
  * Add 0.0.13 bundle to the new catalog image by
    *  creating new catalog index dockerfile
    *  adding package definition for observability-operator *  add 0.0.13 bundle and channels

  *  Add olm/update-channels.sh script to update olm catalog channels
  * Fix GitHub workflow to adapt to constraints introduced by the new file-based catalog

**NOTE:** This PR introduces a change to CI post-merge action - release.

Previously, when a commit is merged to main, the CI job used to publish a catalog image which did not require any change to the code. With the file-based catalog, the new bundle that is published needs to be added to the catalog index.yaml file. Hence, the CI job release that calls the `olm-publish` action now adds a commit on top the commit that got merged into `main` and **pushes** that commit to the repo if a catalog image could be published successfully.

JIRA: https://issues.redhat.com/browse/MON-2791


.github/workflows/release.yaml
```diff
@@ -87,12 +87,14 @@ jobs:
...
-          npx standard-version -i RELEASE_BODY.md --release-as $(cat VERSION) --skip.commit
+          npx standard-version -i RELEASE_BODY.md  \
+            --release-as $(cat VERSION) \
+            --skip.commit
 
       - name: Publish tag
         id: publish_tag
         run: |
-          git push --follow-tags
+          # git push --follow-tags
```
sthaha, line 97: delete this

.github/workflows/release.yaml
```diff
@@ -87,12 +87,13 @@ jobs:
...
           git config user.email release-bot@monitoring.rhobs.io
-          npx standard-version -i RELEASE_BODY.md --release-as $(cat VERSION) --skip.commit
+          npx standard-version -i RELEASE_BODY.md  \
+            --release-as $(cat VERSION) \
+            --skip.commit
 
       - name: Publish tag
         id: publish_tag
         run: |
-          git push --follow-tags
```
tremes, line 95: So this step will just echo the tag name?

sthaha: git push --follow-tags will push the tag that got created in an earlier step

olm/observability-operator-index/index.yaml
```diff
@@ -0,0 +1,91 @@
...
+  mediatype: image/svg+xml
+name: observability-operator
+schema: olm.package
+---
+schema: olm.channel
+package: observability-operator
+name: development
+entries:
+  - name: observability-operator.v0.0.13
+### DEVELOPMENT_CHANNEL_MARKER ###
```
sthaha, line 16: Once this patch merges and we can release 0.0.14 successfully, we should probably revisit this part and see if we can cause kustomize to add the entries. 
For every new version we add it should add something like 

```
  - name: observability-operator.v<new-version>
    replaces: observability-operator.v<previous-version>
```

to appropriate channels.

.github/workflows/release.yaml
tremes: Can we remove this step then?

sthaha: no, it is important to push the tag ... just like the old release workflow. 
What is the concern 🤔

tremes: Updated. Thanks.

.github/workflows/release.yaml
```diff
@@ -83,25 +83,24 @@ jobs:
...
       - name: Generate release notes
+        id: release_notes
         run: |
           # Create the release notes for the Github release
           git config user.name rhobs-release-bot
           git config user.email release-bot@monitoring.rhobs.io
-          npx standard-version -i RELEASE_BODY.md --release-as $(cat VERSION) --skip.commit
+          npx standard-version -i RELEASE_BODY.md  \
+            --release-as $(cat VERSION) \
+            --skip.commit
```
sthaha, line 93: Note: unlike old release workflow, this one does not push the tag and only creates a  draft release without a git tag. The git tag gets created  only after the commit that updates index.yaml is added (after running `olm-publish` workflow) and pushed

Makefile
```diff
@@ -195,11 +186,15 @@ endif
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-image
```
simonpasquier, line 187: Is the comment block still up-to-date?

tremes:
I tried to release my custom catalog image and deploy it to my 4.12 cluster and it works fine.

Makefile
sthaha: Pretty sure it is not even without checking. I am happy to do a follow up cleanup PR, so that the release workflow can be tested with this pr getting merged

simonpasquier: fine by me :)

.github/olm-publish/action.yaml
```diff
@@ -51,8 +51,19 @@ runs:
...
+      git add olm/observability-operator-index/index.yaml
+      git commit -m "ci(bot): update catalog image"
 
   - name: Publish catalog image
     shell: bash
-    run: make catalog-push
+    run: |
+      make catalog-push
+      # update main only if catalog could be successfully published
+      git push origin HEAD:main
```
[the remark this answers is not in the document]
sthaha: True .. and this needs a bit more thought ...

We currently have all images (operartor, bundle, catalog) built in CI. Operator and Bundle are built and pushed to quay.io when a commit merges into main. The new file-based catalog requires us to update the index.yaml file with the bundle information and for that the bundle and operator images must exist; i.e. those must be built in CI.

I can think of the follow options, please feel free to add me :) 

* Do away with maintaining catalog. Users, to test the operator will have to use `operator-sdk` to run the operator bundle like we currently do in development
* Move catalog index to a separate release-branch where the CI is allowed to push, so when a patch merges, the CI checks out the `olm-catalog` branch, updates the index.yaml and pushes it.
* Allow CI to push to main  (using special token)
* Make catalog updates manual but this is quite error prone and I wouldn't consider this as an option
* ... any other thoughts?

olm/update-channels.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+err() {
+  echo "ERROR: $*"
+}
+
+declare -r CATALOG_INDEX_FILE="olm/observability-operator-index/index.yaml"
+
+update_channel() {
+  local channel=$1; shift
```
jan--f, line 18:
```suggestion
  local channel="$1"; shift
```

olm/update-channels.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+main() {
+  cd "$(git rev-parse --show-toplevel)"
+  local channels=$1; shift
+  local bundle=$1; shift
+
+  echo "channels: $channels | bundle: $bundle"
+
+  # convert coma seperated list to an array
```
jan--f, line 48:
```suggestion
  # convert comma seperated list to an array
```

.github/olm-publish/action.yaml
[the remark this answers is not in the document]
jan--f: Would it be an option to have CI build the artifacts in a branch, then open a PR against master? Once this merges CI can push to quay?
This is still adding a manual step, but one could argue that's the intention of branch protection.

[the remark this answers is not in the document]
sthaha: > Would it be an option to have CI build the artifacts in a branch, then open a PR against master?

I see 2 issues with that approach 
1. the images need that are built in response to a merge need to be held somewhere 
2.  The channels in the index.yaml needs a  have a proper dependency graph. This isn't guaranteed if PRs are created based on the code that got merged. 

E.g. 
main: 0.0.13
pr1 - merges -> creates 0.0.14-abc, replaces 0.0.13
pr2 - merges -> creates 0.0.14-xyz, replaces 0.0.13

The expected dependency graph based on the above is 
main: 0.0.13
pr1 - merges -> creates 0.0.14-abc, replaces 0.0.13
pr2 - merges -> creates 0.0.14-xyz, replaces **0.0.14-abc**

Moreover there will be conflict that would need resolution.

olm/observability-operator-index/index.yaml
jan--f: :+1:  yep was thinking the same when I looked at `update-channels.sh` :) 
If kustomize doesn't work maybe at least something else that actually interprets yaml.

jan--f (commented):
Nice work!
I think it would be great to have more comments around the opm usage, especially the Dockerfile and `opm serve`. The documentation wasn't super straight forward to find and outlining what these pieces actually do might be a service to future developers.

Makefile
sthaha: I have removed the stale comment albeit, we still use semver.

.github/olm-publish/action.yaml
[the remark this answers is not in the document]
simonpasquier: I've discussed it offline with @sthaha and the "separate branch for maintaining the catalog index" option looks good to me but I'm not knowledgeable enough about OLM to have a strong opinion on it.

olm/observability-operator-index/index.yaml
sthaha: I have seen other teams write (python) script to parse and update yaml 
but I think maintaining separate files for channels and using kustomize to assemble it might work. I took the easy route (as you can see :D )

sthaha:
@jan--f 
> I think it would be great to have more comments around the opm usage, especially the Dockerfile and opm serve. The documentation wasn't super straight forward to find and outlining what these pieces actually do might be a service to future developers.

As soon as this merges, I am happy to write a docs/ops/release.md where we can explain the release workflow. 

For OLM Catalog I used the following refs
* https://olm.operatorframework.io/docs/tasks/creating-a-catalog/
* https://docs.okd.io/4.11/operators/admin/olm-managing-custom-catalogs.html#olm-managing-custom-catalogs-fb

Where do we want these references to be updated?

.github/olm-publish/action.yaml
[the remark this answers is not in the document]
sthaha: Thanks for the update @simonpasquier . 
Marking this as resolved.

sthaha:
Closing this in favour of https://github.com/rhobs/observability-operator/pull/195 where the catalog index file will be maintained in the `olm-catalog` branch

PR #194, opened 2022-09-16
chore: new file-based-catalog based on 0.0.13

sthaha:
This PR creates a new file-based OLM catalog index based on 0.0.13 bundle
that is already published to quay.io

This is achieved by running

```
make catalog-image VERSION=0.0.13 IMAGE_BASE=quay.io/rhobs/observability-operator
```


sthaha:
cc: @jan--f @simonpasquier

olm/update-channels.sh
```diff
@@ -0,0 +1,59 @@
...
+ sed -e \
+   "s|^\($marker\)|  - name: $bundle\n    replaces: $previous_entry\n\1|" \
+   -i "$CATALOG_INDEX_FILE"
+
+}
+
+main() {
+  cd "$(git rev-parse --show-toplevel)"
+  local channels=$1; shift
+  local bundle=$1; shift
```
jan--f, line 44:
```suggestion
  local channels="$1"; shift
  local bundle="$1"; shift
```

Makefile
```diff
@@ -8,6 +8,7 @@ IMAGE_BASE ?= observability-operator
 
 VERSION ?= $(shell cat VERSION)
 OPERATOR_IMG = $(IMAGE_BASE):$(VERSION)
+OPERATOR_BUNDLE=observability-operator.v$(VERSION)
```
jan--f, line 11:
```suggestion
OPERATOR_BUNDLE=observability-operator-v$(VERSION)
```
Isn't this a more common way of adding a version to the name? I went looking for examples and found https://olm.operatorframework.io/docs/best-practices/channel-naming/ which also seems to suggest `name-version`. Might be less surprising for anyone who tries to parse this?

jan--f (changes requested):
Just two small things want to raise.

Makefile
sthaha: I am a bit confused about this comment. 
We aren't changing the existing channel names `stable, candidate, development` in this patch. 

If it is about the change proposed, that wouldn't work since `opm render <bundle-image>`  uses the format in PR. E.g. 

```
❯ opm render quay.io/rhobs/observability-operator-bundle:0.0.13 --output=yaml | ag '^name:'
name: observability-operator.v0.0.13
```

olm/update-channels.sh
sthaha: Thanks, I have fixed and updated the commit.

Makefile
jan--f: Got it. I just pointed it out cause the name format looked odd to me. But since its not under our control I'll merge this.

jan--f (approved):
lgtm

PR #197, opened 2022-09-20
Makefile and readme additions

jan--f:
Just a few updates that should lower hurdles for our build and release process. Commit messages have more details.

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -72,10 +72,6 @@ $(GOLANGCI_LINT) golangci-lint: $(TOOLS_DIR)
 		GOBIN=$(TOOLS_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) ;\
 	}
 
-# NOTE: kustomize does not support `go install` hence this workaround to install
-# it by creating an tmp module and using go get to download the precise version
-# needed for the project
-# see: https://github.com/kubernetes-sigs/kustomize/issues/3618
 .PHONY: kustomize
 $(KUSTOMIZE) kustomize: $(TOOLS_DIR)
 	@{ \
@@ -88,7 +84,11 @@ $(KUSTOMIZE) kustomize: $(TOOLS_DIR)
 $(OPERATOR_SDK) operator-sdk: $(TOOLS_DIR)
 	@{ \
 		set -ex ;\
-		[[ -f $(OPERATOR_SDK) ]] && exit 0 ;\
+		[[ -f $(OPERATOR_SDK) ]] && \
+		[[ $(shell $(OPERATOR_SDK) version | awk '{print $$3}' | tr -d ,) = $(OPERATOR_SDK_VERSION) ]] && { \
+			echo "sdk up to date"; \
+			exit 0 ;\
+		}; \
 		OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
 		curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_$${OS}_$${ARCH} ;\
 		chmod +x $(OPERATOR_SDK) ;\
--- a/README.md
+++ b/README.md
@@ -28,6 +28,16 @@ Refer `Running the Operator in Kind `below to run the operator in Kubernetes.
 
 In order to contribute to this project, make sure you have go 1.17 on your local machine.
 
+### Dependencies
+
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -28,6 +28,14 @@ Refer `Running the Operator in Kind `below to run the operator in Kubernetes.
 
 In order to contribute to this project, make sure you have go 1.17 on your local machine.
 
+### Dependencies
+
+The build system assumes the following binaries are in the `PATH`:
+
+```
+make git golang npm
```
sthaha, line 36:
```suggestion
make git go npm
```

Makefile.tools
```diff
@@ -88,7 +84,13 @@ $(KUSTOMIZE) kustomize: $(TOOLS_DIR)
...
 	@{ \
 		set -ex ;\
-		[[ -f $(OPERATOR_SDK) ]] && exit 0 ;\
+		if [ -f $(OPERATOR_SDK) ]; then \
+			if [ $(shell $(OPERATOR_SDK) version | awk '{print $$3}' | tr -d ,) = $(OPERATOR_SDK_VERSION) ]; then \
+				echo "sdk up to date"; \
+				exit 0 ;\
+			fi; \
+			echo "sdk needs update" ;\
+		fi; \
```
sthaha, line 93: We can simplify nested using `&&` and totally do away with `if` e.g. the snippet below works 


```suggestion
		[[ -x $(OPERATOR_SDK) ]] && \
		[[  "$$( $(OPERATOR_SDK) version | awk '{print $$3}' | tr -d , )" == \"$(OPERATOR_SDK_VERSION)\" ]] && { \
				echo "operator-sdk up to date"; \
				exit 0 ;\
		}; \
```

I think `[[ -x ` is better than `[[ -f`

README.md
jan--f: Yeah fair point since I call out binaries. `go` packages are often called golang...

Makefile.tools
jan--f: - Not sure we should move to `-x`. Directories can be executable as well. We really only want to check if a file exists so `-e` would probably me most appropriate. `-f` ensures its a regular file, if this is not executable, I think the error message from trying to execute it should suffice?
- `[[ ]]` vs `if`, you got a point for consistency's sake. Generally I'd be more inclined to be more explicit about the logic structure. But this will unlikely change a lot, and if it does I'd probably argue for not using bash anymore. This change was already too tricky to get right really.

jan--f: case in point, this already does not work on macos... _sigh_

sthaha: > Not sure we should move to -x. Directories can be executable as well. We really only want to check if a file exists so -e would probably me most appropriate. -f ensures its a regular file, if this is not executable, I think the error message from trying to execute it should suffice?

`./dir` gets you `./dir is a directory` whereas `./file` tells you `./tmp/bin/foo: Permission denied` and hence ` -x ` but I agree we don't have to fret on this 

> [[ ]] vs if, you got a point for consistency's sake

I don't have strong opinion on this. I was pointing at avoiding nested if by using `&&` 
```
if [[ -f foo ]] && [[ "$(./foo | awk .. )" == "bar" ]]; then
  
fi
```


Which should work on bash 3.2 as well.

sthaha: > case in point, this already does not work on macos... sigh

```
bash: -c: line 3: conditional binary operator expected
bash: -c: line 3: syntax error near `v1.23.0'
bash: -c: line 3: `	[[  = v1.23.0 ]] && { \'
```
That seems to because of missing `"` around  `"$$( ... )"`

Makefile.tools
```diff
@@ -88,7 +84,11 @@ $(KUSTOMIZE) kustomize: $(TOOLS_DIR)
 $(OPERATOR_SDK) operator-sdk: $(TOOLS_DIR)
 	@{ \
 		set -ex ;\
-		[[ -f $(OPERATOR_SDK) ]] && exit 0 ;\
+		[[ -f $(OPERATOR_SDK) ]] && \
+		[[ $(shell $(OPERATOR_SDK) version | awk '{print $$3}' | tr -d ,) = $(OPERATOR_SDK_VERSION) ]] && { \
```
sthaha, line 88:
```suggestion
		[[ "$(shell $(OPERATOR_SDK) version | awk '{print $$3}' | tr -d ,)" = $(OPERATOR_SDK_VERSION) ]] && { \
```

sthaha:
@jan--f Shall we close this since the Makefile in `main` now has the fix to verify the operator-sdk version?

jan--f:
Yep lets close it. I added the dependency note for the readme in #209

PR #202, opened 2022-10-13
fix: do not update MonitoringStack status in case of Prometheus

tremes:
generation difference

This adds check comparing the Prometheus observedGeneration to the current MonitoringStack generation. If there is a difference then the MonitoringStack status is not updated.

### DIFF
--- a/pkg/controllers/monitoring/monitoring-stack/conditions.go
+++ b/pkg/controllers/monitoring/monitoring-stack/conditions.go
@@ -21,7 +21,7 @@ const (
 	NoReason                       = "None"
 )
 
-func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.PrometheusCondition, generation int64, recError error) []v1alpha1.Condition {
+func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, generation int64, recError error) []v1alpha1.Condition {
 	if len(msConditions) == 0 {
 		return []v1alpha1.Condition{
 			{
@@ -43,13 +43,13 @@ func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.
 	for _, mc := range msConditions {
 		switch mc.Type {
 		case v1alpha1.AvailableCondition:
-			available := updateAvailable(mc, promConditions, generation)
+			available := updateAvailable(mc, prom, generation)
 			if !available.Equal(mc) {
 				available.LastTransitionTime = metav1.Now()
 			}
 			updatedConditions = append(updatedConditions, available)
 		case v1alpha1.ReconciledCondition:
-			reconciled := updateReconciled(mc, promConditions, generation, recError)
+			reconciled := updateReconciled(mc, prom, generation, recError)
 			if !reconciled.Equal(mc) {
 				reconciled.LastTransitionTime = metav1.Now()
 			}
@@ -62,15 +62,18 @@ func updateConditions(msConditions []v1alpha1.Condition, promConditions []monv1.
 
 // updateAvailable gets existing "Available" condition and updates its parameters
 // based on the Prometheus "Available" condition
-func updateAvailable(ac v1alpha1.Condition, prometheusConditions []monv1.PrometheusCondition, generation int64) v1alpha1.Condition {
-	ac.ObservedGeneration = generation
-
-	prometheusAvailable, err := getPrometheusCondition(prometheusConditions, monv1.PrometheusAvailable)
+func updateAvailable(ac v1alpha1.Condition, prom monv1.Prometheus, generation int64) v1alpha1.Condition {
+	prometheusAvailable, err := getPrometheusCondition(prom.Status.Conditions, monv1.PrometheusAvailable)
 
 	if err != nil {
 		ac.Status = v1alpha1.ConditionUnknown
 		ac.Reason = PrometheusNotAvailable
 		ac.Message = CannotReadPrometheusConditions
+		return ac
+	}
+	// MonitoringStack status will not be updated if there is a difference between the Prometheus generation
+	// and the Prometheus ObservedGeneration. This can occur, for example, in the case of an invalid Prometheus configuration.
+	if prometheusAvailable.ObservedGeneration != prom.Generation {
 		return ac
 	}
 
@@ -87,13 +90,13 @@ func updateAvailable(ac v1alpha1.Condition, prometheusConditions []monv1.Prometh
 	ac.Status = v1alpha1.ConditionTrue
 	ac.Reason = AvailableReason
 	ac.Message = AvailableMessage
+	ac.ObservedGeneration = generation
 	return ac
 }
 
 // updateReconciled updates "Reconciled" conditions based on the provided error value and
 // Prometheus "Reconciled" condition
-func updateReconciled(rc v1alpha1.Condition, prometheusConditions []monv1.PrometheusCondition, generation int64, err error) v1alpha1.Condition {
-	rc.ObservedGeneration = generation
+func updateReconciled(rc v1alpha1.Condition, prom monv1.Prometheus, generation int64, err error) v1alpha1.Condition {
 
 	if err != nil {
 		rc.Status = v1alpha1.ConditionFalse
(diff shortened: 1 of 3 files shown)


tremes:
@thetechnick can you please check whether this makes sense? :)

tremes:
This is for the https://issues.redhat.com/browse/MON-2796

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -74,6 +72,10 @@ func updateAvailable(ac v1alpha1.Condition, prometheusConditions []monv1.Prometh
 		return ac
 	}
 
+	if prometheusAvailable.ObservedGeneration != generation {
```
simonpasquier, line 75: hmm if I understand correctly, we want to check the Prometheus condition's  observedGeneration against the Prometheus object's generation.

tremes: Well I've discussed with Nico and maybe I still don't fully get it, but my understanding is following:
- MonitoringStack spec is updated -> generation is bumped
- OBO is reflecting the change in the Prometheus

thetechnick: You have to compare Prometheus.metadata.generation with `Prometheus.status.condition[].observedGeneration` and
not `MonitoringStack.metadata.generation` with `Prometheus.status.condition[].observedGeneration`

Under perfect conditions, it would not actually matter, but if anyone with UPDATE permission to Prometheus is updating Prometheus directly you will end up with:
`MonitoringStack.metadata.generation = 1` and `Prometheus.metadata.generation = 3`

image (8)

tremes: OK thanks. I will update it. And the case (I tried to describe above) can't happen (maybe not...)?

tremes:
@simonpasquier, @thetechnick can you please check again?

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -27,7 +27,11 @@ func TestUpdateConditions(t *testing.T) {
 			originalMSConditions: []v1alpha1.Condition{},
 			generation:           1,
 			recError:             nil,
-			prometheusConditions: []monv1.PrometheusCondition{},
+			prometheus: monv1.Prometheus{
+				Status: monv1.PrometheusStatus{
+					Conditions: []monv1.PrometheusCondition{},
+				},
+			},
```
simonpasquier, line 34:
```suggestion
			prometheus: monv1.Prometheus{},
```

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -111,9 +119,12 @@ func TestUpdateConditions(t *testing.T) {
...
 			},
-			generation:           1,
-			recError:             nil,
-			prometheusConditions: []monv1.PrometheusCondition{},
+			generation: 1,
+			recError:   nil,
+			prometheus: monv1.Prometheus{
+				Status: monv1.PrometheusStatus{
+					Conditions: []monv1.PrometheusCondition{},
+				}},
```
simonpasquier, line 127:
```suggestion
			prometheus: monv1.Prometheus{},
```

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -176,10 +194,68 @@ func TestUpdateConditions(t *testing.T) {
 					Reason:             PrometheusNotReconciled,
 				}},
 		},
+		{
+			name: "Prometheus observed generation is different to the Prometheus generation",
```
simonpasquier, line 198: (nit)
```suggestion
			name: "Prometheus observed generation is different than the Prometheus generation",
```

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -176,10 +194,68 @@ func TestUpdateConditions(t *testing.T) {
...
+					Message:            SuccessfullyReconciledMessage,
+					LastTransitionTime: transitionTime,
+				}},
+			sameTransitionTimes: true,
+		},
 	}
 
 	for _, test := range tt {
-		res := updateConditions(test.originalMSConditions, test.prometheusConditions, test.generation, test.recError)
+		res := updateConditions(test.originalMSConditions, test.prometheus, test.generation, test.recError)
```
sthaha, line 258: Would these unit tests suffice? I wonder if we should add an e2e test that tests this scenario in real world. 
I believe having an e2e validates this specific requirement is met and can potentially catch an error in PO and/or ObO should things change later.

sthaha (changes requested):
Could you please add the Jira link to the commit ?

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
sthaha: (nit on nit) I think the common use is `different from` but if my understanding is right, it is fine to use `from / to / than` 

Not native speaker so take this with a large grain on salt :D

tremes: :) https://www.deepl.com/translator tells me to use `from` :)

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
simonpasquier: An e2e test would be great indeed. It'd mean to scale the prometheus operator to zero, modify the Prometheus resource and check that the status doesn't change.

tremes: Hmm does it mean we need to know the namespace where the OBO is installed (to be able to scale down the P-O)? Can we find it deterministically? I can imagine we could use hardcoded value for kind, but it's not very nice....other option is to list all the deployments and try to find by name (again not very nice IMO). Or am I missing something?

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
tremes: Thanks

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
sthaha: > It'd mean to scale the prometheus operator to zero ... 

That may not be an option since OLM can possibly reset it albeit it doesn't seem to at the moment but we shouldn't rely on this behaviour. 

Would it possible to somehow set the `ms.Spec.PrometheusConfig` to be something invalid that results in PO not updating the ObservedGeneration ?

tremes: Hmm ok I can try....but I guess it will be a bit tricky. I think the time when there will be the difference between the generations will be very very short.

tremes: Yes I think Simon is right that we would need to scale down the P-O. I was experimenting manually and when you update Prometheus config then the P-O updates the status conditions (observedGeneration) almost immediately. When I scaled down the P-O then I was able to reproduce such situation.

>Would it possible to somehow set the ms.Spec.PrometheusConfig to be something invalid that results in PO not updating the ObservedGeneration ?

I guess P-O should always update the status or not? 

Anyway I would suggested to merge this PR and then maybe work on the e2e test.

sthaha:
@simonpasquier @jan--f, could you please take another look?

simonpasquier (approved):
lgtm

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
jan--f: I'd be ok with merging this without the e2e test. Iiuc this test would mostly test correct behavior of PO anyway?

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -74,6 +72,10 @@ func updateAvailable(ac v1alpha1.Condition, prometheusConditions []monv1.Prometh
 		return ac
 	}
 
+	if prometheusAvailable.ObservedGeneration != prom.Generation {
+		return ac
+	}
+
```
jan--f, line 78: Should we leave a note for future devs on what the reasoning is for not updating the conditions?
I'd be :+1: on that since this PR already needed a bit of discussion about the motivation for this.
Otherwise this lgtm.

tremes: Yeah why not. Let me add some explanation there.

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
tremes: I think it's not just PO behaviour. We want to be sure OBO works as expected, but otherwise yes....let's work on the e2e test in different PR and don't block this one.

jan--f (dismissed):
Thanks!

sthaha:
@tremes could you please rebase on to latest main to get the CI passing

sthaha (approved):

PR #207, opened 2022-10-26
fix: target management clusters in SSS

jan--f:

### DIFF
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -22,7 +22,7 @@ objects:
     spec:
       clusterDeploymentSelector:
         matchLabels:
-          api.openshift.com/addon-reference-addon: 'true'
+          api.openshift.com/managed: "true"
       resourceApplyMode: Sync
       resources:
         - apiVersion: v1


apahim:
MON-2677 is about deploying OBO to the entire OSD/ROSA fleet, which can be achieved by using the label:

```
api.openshift.com/managed: "true"
```

Why are we targeting Management Clusters only?

jan--f:
> MON-2677 is about deploying OBO to the entire OSD/ROSA fleet, which can be achieved by using the label:
> 
> ```
> api.openshift.com/managed: "true"
> ```
> 
> Why are we targeting Management Clusters only?

My bad, fixed now

deploy/olm/syncselector-template.yaml
```diff
@@ -22,7 +22,8 @@ objects:
     spec:
       clusterDeploymentSelector:
         matchLabels:
-          api.openshift.com/addon-reference-addon: 'true'
+          ext-hypershift.openshift.io/cluster-type: management-cluster
+          ext-hypershift.openshift.io/cluster-type: service-cluster
```
apahim, line 26: We should use just:
```
api.openshift.com/managed: "true"
```
Nothing else, so it hits the entire fleet.

maorfr: https://issues.redhat.com/browse/SDA-6646?focusedCommentId=20924957&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20924957

maorfr: After reading this comment, +1 to what @apahim is suggesting

wanghaoran1988: My understanding is the two labels listed here is a "AND" operation, there is no cluster will has these two labels together.

maorfr: we've concluded over slack that @apahim's proposal is the way to go.

maorfr (approved):
/lgtm

sthaha (approved):

PR #208, opened 2022-10-27
fix: workaround CI failure by trying again

sthaha:
The workaround uses wait.PollImmediate to try to Patch again for at most 30 seconds

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -3,9 +3,14 @@ package reconciler
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -33,14 +36,66 @@ func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime
...
+
+		// failing to patch shouldn't be retried since it is a genuine error
+		if err := c.Patch(ctx, r.resource, client.Apply, client.ForceOwnership, client.FieldOwner(owner)); err != nil {
+			patchError = fmt.Errorf("%s/%s (%s): updater failed to patch: %w", ns, name, gvk.String(), err)
+			return done, nil
+		}
+
+		// Get requires a client.Object but we can't use r.resource since it may
+		// need to be reapplied, so deepcopy the resource and set the GVK
+		dup, ok := r.resource.DeepCopyObject().(client.Object)
```
sthaha, line 71: @simonpasquier @jan--f, Do you know of a better way that type asserting the copy? 
I have seem similar approach in controller-runtime code as well but if you know of a better way, I am all ears

PR #213, opened 2022-11-01
fix: broken release workflow

sthaha:
GitHub release workflow was broken by changes made in commit 4625e3e which was intended to fix deprecation message but the usage was wrong and didn't correctly fix all warnings. This commit aims to fix all warnings and deprecation messages.

### DIFF
--- a/.github/e2e-tests-olm/action.yaml
+++ b/.github/e2e-tests-olm/action.yaml
@@ -23,7 +23,7 @@ runs:
     - name: Use go cache
       uses: ./.github/go-cache
 
-    - uses: azure/setup-kubectl@v1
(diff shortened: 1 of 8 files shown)


sthaha:
Please note that engineerd/setup-kind is the only action that continues to use node 12. 
May be worth switching to https://github.com/helm/kind-action since it uses node 16 https://github.com/helm/kind-action/blob/main/action.yml#L38

sthaha:
Tested the workflow using - https://github.com/sthaha/observability-operator/actions/runs/3366617136/jobs/5583287072

tremes:
lgtm

jan--f (approved):
Thanks!

PR #215, opened 2022-11-01
fix: validate Prometheus replicas nummber

tremes:
This adds kubebuilder validation to Prometheus replicas attribute in the MonitoringStack API.

JIRA: https://issues.redhat.com/browse/MON-2838


jan--f (dismissed):
Nice, thanks!

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -156,6 +156,8 @@ type PrometheusConfig struct {
 	// Number of replicas/pods to deploy for a Prometheus deployment.
 	// +optional
 	// +kubebuilder:default=2
+	// +kubebuilder:validation:Required
```
sthaha, line 159: doesn't this conflict with the `//optional` above. 
`replicas` is optional and defaults to `2`. I think only `validation:Minimum` is what we need

tremes: Hmm maybe....let me try

tremes: It seems to work exactly the same with `kubebuilder:validation:Required` as well as without it. Do you want me to remove it?

jan--f: If you look at the CRD manifest, the `Required` doesn't actually do anything. That's likely because we have a default. So :+1: for dropping this annotation.

sthaha (approved):
Thanks for the patch @tremes 🙇

PR #220, opened 2022-11-10
build: publish catalog image with SHA

sthaha:
To enable continuous delivery of OBO to OSD/ROSA clusters via app-interface, the OLM catalog image must be tagged with a git short hash that represents the commit that was used for the creation of the catalog.

This patch adds a make target `catalog-tag-sha` that tags the catalog image with the SHA of the commit that CI bot makes when it updates the catalog.

In addition to that the PR also update the Makefile to reflect the FBC change that was pushed to olm-catalog branch but was missing in main.

JIRA: https://issues.redhat.com/browse/MON-2859

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ IMAGE_BASE ?= observability-operator
 
 VERSION ?= $(shell cat VERSION)
 OPERATOR_IMG = $(IMAGE_BASE):$(VERSION)
+OPERATOR_BUNDLE=observability-operator.v$(VERSION)
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 
 # running `make` builds the operator (default target)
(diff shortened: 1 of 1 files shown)


sthaha:
I have tried this on my fork 
* After merging the PR the [Release - job](https://github.com/sthaha/observability-operator/actions/runs/3434012700/jobs/5725059562) published [`d41ceef8` tagged catalog image](https://quay.io/repository/sthaha/observability-operator-catalog?tab=tags) 
* https://github.com/sthaha/observability-operator/commit/d41ceef83828783a71c909da7c6f18ca78d5aa99 is the commit made by the CI bot

JoaoBraveCoding (commented):
LGTM 👍 just thinking, if it wouldn't also help to have the version preceding the SHA. But maybe the app-interface folks are not interested in that

sthaha:
@JoaoBraveCoding
>  if it wouldn't also help to have the version preceding the SHA

I thought the same however ruled that out because it wouldn't help us trace from the image the commit responsible for create the image.  I feel the current implementation is better since it helps us trace from the image (`tag`)  the bundle and the catalog used to publish the image.  So reproducing the same build is straight forward. 

If you look at the commts in olm-catalog branch, for every release you will find 2 additional commits on top of the `main` 

1. `main` : the commit that just got merged 
2.  `merge olm-catalog`: this merges olm-catalog branch (with FBC) to create the catalog
 `... after this is merged, we run all the `make images` and `catalog`` 
3. `ci(bot): update catalog image`: this commit is made after the bundles and the catalog are updated. 

Currently we are tagging the `3rd` commit.  If we tagged the first commit, then given an image `foobar` we can only guess what the actual catalog and the bundle is.

JoaoBraveCoding:
Ahh OK I think I get it thanks for the clarification!

JoaoBraveCoding (approved):
lgtm

PR #223, opened 2022-11-17
feat: add resourceDiscovery status condition

jan--f:
feat: add resourceDiscovery status condition

    The ResourceSelector field in a MonitoringStack CR can be set to nil.
    This is passed on to the Prometheus CR and can be used to disable all
    service discovery and thus most scraping (we add prometheus and
    alertmanager scrape jobs via additionalScrapeConfigs).
    When this is configured we now set a new status condition to alert users
    to the fact that no service discovery is configured.

Fixes: https://issues.redhat.com/browse/MON-2864

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/types.go
+++ b/pkg/apis/monitoring/v1alpha1/types.go
@@ -58,8 +58,11 @@ type MonitoringStackSpec struct {
 	LogLevel LogLevel `json:"logLevel,omitempty"`
 
 	// Label selector for Monitoring Stack Resources.
+	// Set to the empty LabelSelector ({}) to monitoring everything.
+	// Set to null to disable service discovery.
 	// +optional
-	ResourceSelector *metav1.LabelSelector `json:"resourceSelector,omitempty"`
+	// +nullable
+	ResourceSelector *metav1.LabelSelector `json:"resourceSelector"`
 
 	// Namespace selector for Monitoring Stack Resources.
 	// If left empty the Monitoring Stack will only match resources in the namespace it was created in.
@@ -108,8 +111,9 @@ const (
 	ConditionFalse   ConditionStatus = "False"
 	ConditionUnknown ConditionStatus = "Unknown"
 
-	ReconciledCondition ConditionType = "Reconciled"
-	AvailableCondition  ConditionType = "Available"
+	ReconciledCondition        ConditionType = "Reconciled"
+	AvailableCondition         ConditionType = "Available"
+	ResourceDiscoveryCondition ConditionType = "ResourceDiscovery"
 )
 
 type Condition struct {
--- a/pkg/controllers/monitoring/monitoring-stack/conditions.go
+++ b/pkg/controllers/monitoring/monitoring-stack/conditions.go
@@ -15,13 +15,17 @@ const (
 	PrometheusNotAvailable         = "PrometheusNotAvailable"
 	PrometheusNotReconciled        = "PrometheusNotReconciled"
 	PrometheusDegraded             = "PrometheusDegraded"
+	ResourceSelectorIsNil          = "ResourceSelectorNil"
 	CannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
 	AvailableMessage               = "Monitoring Stack is available"
 	SuccessfullyReconciledMessage  = "Monitoring Stack is successfully reconciled"
+	ResourceSelectorIsNilMessage   = "No resources will be discovered, ResourceSelector is nil"
+	ResourceDiscoveryOnMessage     = "Resource discovery is operational"
 	NoReason                       = "None"
 )
 
-func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, generation int64, recError error) []v1alpha1.Condition {
+func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, recError error) []v1alpha1.Condition {
+	msConditions := ms.Status.Conditions
 	if len(msConditions) == 0 {
 		return []v1alpha1.Condition{
 			{
@@ -36,8 +40,10 @@ func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, 
 				Reason:             NoReason,
 				LastTransitionTime: metav1.Now(),
 			},
+			updateResourceDiscovery(ms),
 		}
 	}
+	generation := ms.Generation
 	var updatedConditions []v1alpha1.Condition
 
 	for _, mc := range msConditions {
@@ -54,10 +60,42 @@ func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, 
 				reconciled.LastTransitionTime = metav1.Now()
 			}
 			updatedConditions = append(updatedConditions, reconciled)
+		case v1alpha1.ResourceDiscoveryCondition:
+			resourceDiscovery := updateResourceDiscovery(ms)
+			if !resourceDiscovery.Equal(mc) {
+				resourceDiscovery.LastTransitionTime = metav1.Now()
+			}
+			updatedConditions = append(updatedConditions, resourceDiscovery)
 		}
 	}
 
 	return updatedConditions
+}
+
+// updateResourceDiscovery updates the ResourceDiscoveryCondition based on the
+// ResourceSelector in the MonitorinStack spec. A ResourceSelector of nil causes
+// the condition to be false, any other value sets the condition to true
+func updateResourceDiscovery(ms *v1alpha1.MonitoringStack) v1alpha1.Condition {
+	if ms.Spec.ResourceSelector == nil {
+		return v1alpha1.Condition{
+			Type:               v1alpha1.ResourceDiscoveryCondition,
+			Status:             v1alpha1.ConditionFalse,
+			Reason:             ResourceSelectorIsNil,
+			Message:            ResourceSelectorIsNilMessage,
+			LastTransitionTime: metav1.Now(),
+			ObservedGeneration: ms.Generation,
+		}
+	} else {
+		return v1alpha1.Condition{
+			Type:               v1alpha1.ResourceDiscoveryCondition,
+			Status:             v1alpha1.ConditionTrue,
+			Reason:             NoReason,
+			Message:            ResourceDiscoveryOnMessage,
+			LastTransitionTime: metav1.Now(),
+			ObservedGeneration: ms.Generation,
+		}
+	}
+
 }
 
 // updateAvailable gets existing "Available" condition and updates its parameters
--- a/pkg/controllers/monitoring/monitoring-stack/conditions.go
+++ b/pkg/controllers/monitoring/monitoring-stack/conditions.go
@@ -25,51 +25,20 @@ const (
 )
 
 func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, recError error) []v1alpha1.Condition {
-	msConditions := ms.Status.Conditions
-	if len(msConditions) == 0 {
-		return []v1alpha1.Condition{
-			{
-				Type:               v1alpha1.AvailableCondition,
-				Status:             v1alpha1.ConditionUnknown,
-				Reason:             NoReason,
-				LastTransitionTime: metav1.Now(),
-			},
-			{
-				Type:               v1alpha1.ReconciledCondition,
-				Status:             v1alpha1.ConditionUnknown,
-				Reason:             NoReason,
-				LastTransitionTime: metav1.Now(),
-			},
-			updateResourceDiscovery(ms),
+	return []v1alpha1.Condition{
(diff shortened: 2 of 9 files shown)


JoaoBraveCoding (dismissed):
/lgtm

sthaha (commented):
Nice! lgtm

I do feel that we need a test that asserts that explicit `nil` is preserved and not specifying anything sets `{}` 

To actually cover the usecase we discussed, I think we need a e2e test that 
* sets resourceSelector
* resets to nil
* validates selectors in Prometheus reflect the same 
* 

WDYT @simonpasquier ?

jan--f:
i'm not sure OBO is the right place for these tests. Prometheus-operator would be a better place for this, wouldn't it?
[A test for `ServiceMonitorSelector = nil`](https://github.com/prometheus-operator/prometheus-operator/blob/66f07d0267e224391d16e0ff20d771ddd0b61404/test/e2e/prometheus_test.go#L809-L822) already exists in the prometheus-operator repo. If a check for "No generated scrape jobs in Prometheus" makes sense in an e2e test, I'd argue over there is the right place.

sthaha:
> * sets resourceSelector
> * resets to nil
> * validates selectors in Prometheus reflect the same



> i'm not sure OBO is the right place for these tests. Prometheus-operator would be a better place for this, wouldn't it? [A test for `ServiceMonitorSelector = nil`](https://github.com/prometheus-operator/prometheus-operator/blob/66f07d0267e224391d16e0ff20d771ddd0b61404/test/e2e/prometheus_test.go#L809-L822) already exists in the prometheus-operator repo.

I am not sure I agree ... The suggestion was to only validate the functionality provided by ObO .. i.e. setting `resourceSelector` of the **MonitoringStack** from `{}` (default) to `nil` does indeed update/generate `Prometheus` (which ObO generates) that has the `<resource>Selector` set to nil. 
We are not validating the functionality of the PO rather only validating the functionality provided by Obo. 


 > If a check for "No generated scrape jobs in Prometheus" makes sense in an e2e test, I'd argue over there is the right place.
This I agree is beyond the scope of the ObO tests but that isn't what I was requesting.

jan--f:
> The suggestion was to only validate the functionality provided by ObO .. i.e. setting resourceSelector of the MonitoringStack from {} (default) to nil does indeed update/generate Prometheus (which ObO generates) that has the <resource>Selector set to nil.

Ahh I see, yeah thats indeed worth testing here. I'll add that.

simonpasquier (commented):
late to this but yes I agree with @sthaha that end-to-end tests for the following are valuable:
* By default, ObO provisions underlying objects with the empty selector.
* `spec.resourceSelector: nil` is supported and propagates to the underlying objects.

jan--f:
Update: I have a test for this but I'm seeing some unexpected behavior which I'd like to look into.
It seems like `+kubebuilder:default={}` on `ResourceSelector` doesn't have the desired effect and when a stack with a `null` selector is created, thanos side car doesn't start.

sthaha:
> It seems like +kubebuilder:default={} on ResourceSelector doesn't have the desired effect and when a stack with a null selector is created,

I suspected `nil` will be set back to `{}` but this is even weirder :(

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -55,8 +55,10 @@ type MonitoringStackSpec struct {
 	LogLevel LogLevel `json:"logLevel,omitempty"`
 
 	// Label selector for Monitoring Stack Resources.
-	// +optional
-	ResourceSelector *metav1.LabelSelector `json:"resourceSelector,omitempty"`
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default={}
```
simonpasquier, line 59: `+kubebuilder:default` doesn't seem to play well with pointer fields: the default value doesn't appear in the generated YAML manifest of the CRD :thinking:

simonpasquier: I suspect that kubebuilder's defaulting works well with value fields but not so much with pointers.

JoaoBraveCoding (commented):
In general looks good! Just two comments but nothing blocking 👍

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -15,13 +15,17 @@ const (
...
+	ResourceSelectorIsNil          = "ResourceSelectorNil"
 	CannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
 	AvailableMessage               = "Monitoring Stack is available"
 	SuccessfullyReconciledMessage  = "Monitoring Stack is successfully reconciled"
+	ResourceSelectorIsNilMessage   = "No resources will be discovered, ResourceSelector is nil"
+	ResourceDiscoveryOnMessage     = "Resource discovery is operational"
 	NoReason                       = "None"
 )
 
-func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, generation int64, recError error) []v1alpha1.Condition {
```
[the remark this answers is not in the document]
jan--f: This is unfortunately necessary since the resource discovery condition is set based on a field of `MonitoringStack`. 
See https://github.com/rhobs/observability-operator/pull/223/files#diff-3c84e774a1e9ef2eafc525a973f4b4a273410b6b242fabda42e514c524328a38R64
And since the conditions are part of that anyway I simply changed the argument, instead of passing both the conditions list and the monitoring stack.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -54,12 +60,44 @@ func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus,
...
+			}
+			updatedConditions = append(updatedConditions, resourceDiscovery)
 		}
 	}
 
 	return updatedConditions
 }
 
+// updateResouceDiscovery updates the ResouceDiscoveryCondition based on the
+// ResrouceSelector in the MonitorinStack spec. A ResourceSelector of nil causes
```
tremes, line 76: Nit: typo in `ResrouceSelector`

jan--f: thanks, nice catch. For some reason typing `resource` is the bane of my existence.

pkg/controllers/monitoring/monitoring-stack/conditions.go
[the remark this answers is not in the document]
JoaoBraveCoding: Ahh yes, missed that detail!

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -13,152 +13,220 @@ func TestUpdateConditions(t *testing.T) {
 	transitionTime := metav1.Now()
 
 	tt := []struct {
-		name                 string
-		originalMSConditions []v1alpha1.Condition
-		prometheus           monv1.Prometheus
-		generation           int64
-		recError             error
-		expectedResults      []v1alpha1.Condition
+		name             string
```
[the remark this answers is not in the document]
JoaoBraveCoding: OK after that explanation and with the argument change, I think it makes sense to split!

JoaoBraveCoding (approved):
lgtm

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -58,8 +58,11 @@ type MonitoringStackSpec struct {
 	LogLevel LogLevel `json:"logLevel,omitempty"`
 
 	// Label selector for Monitoring Stack Resources.
+	// Set to the empty LabelSelector to monitoring everything.
```
sthaha, line 61:
```suggestion
	// Set to an empty LabelSelector ({}) to monitoring everything.
```

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -15,13 +15,17 @@ const (
...
 	AvailableMessage               = "Monitoring Stack is available"
 	SuccessfullyReconciledMessage  = "Monitoring Stack is successfully reconciled"
+	ResourceSelectorIsNilMessage   = "No resources will be discovered, ResourceSelector is nil"
+	ResourceDiscoveryOnMessage     = "Resource discovery is operational"
 	NoReason                       = "None"
 )
 
-func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus, generation int64, recError error) []v1alpha1.Condition {
+func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, recError error) []v1alpha1.Condition {
+	msConditions := ms.Status.Conditions
```
sthaha, line 28: nit: could we move this close to where it is used? and perhaps we may not even need this.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -54,12 +60,44 @@ func updateConditions(msConditions []v1alpha1.Condition, prom monv1.Prometheus,
 				reconciled.LastTransitionTime = metav1.Now()
 			}
 			updatedConditions = append(updatedConditions, reconciled)
+		case v1alpha1.ResourceDiscoveryCondition:
```
sthaha, line 63: doesn't this mean, that a `ms` that is already deployed prior to this feature  will  never  have this condition set? Instead shouldn't we set all 3 conditions irrespective of them them having the condition?

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -122,6 +128,23 @@ func emptyStackCreatesPrometheus(t *testing.T) {
...
+	f.AssertResourceEventuallyExists(ms.Name, ms.Namespace, &monv1.Prometheus{})(t)
+
+	updatedMS := &stack.MonitoringStack{}
+	f.GetResourceWithRetry(t, ms.Name, ms.Namespace, updatedMS)
+	updatedMS.Spec.ResourceSelector = nil
+	err = f.K8sClient.Update(context.Background(), updatedMS)
+	assert.NilError(t, err, "failed to patch monitoring stack with nil resource selector")
+
+	prometheus := monv1.Prometheus{}
+	f.GetResourceWithRetry(t, updatedMS.Name, updatedMS.Namespace, &prometheus)
```
sthaha, line 144: this may yield failure depending on the timing because `GetResourceWithRetry` may give us the old Prometheus ... we need to assert the generation has changed whenever we modify an exisiting stack.

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -58,8 +58,11 @@ type MonitoringStackSpec struct {
...
+	// Set to null to disable service discovery.
 	// +optional
-	ResourceSelector *metav1.LabelSelector `json:"resourceSelector,omitempty"`
+	// +nullable
```
sthaha, line 64: Shouldn't we set the default to have service discovery enabled? i.e. - `{}` ?

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -626,10 +649,17 @@ func msNamespaceSelector(labels map[string]string) stackModifier {
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: stack.GroupVersion.String(),
+			Kind:       "MonitoringStack",
+		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: e2eTestNamespace,
 		},
+		Spec: stack.MonitoringStackSpec{
+			ResourceSelector: &metav1.LabelSelector{},
```
sthaha, line 661: Isn't it better to use a `stackModifier` to explicitly set the ResourceSelector since the default is nil?

jan--f: Given that the majority of tests expects the empty selector and this is a convenience function I went with this.
Causes less code changes to existing code too.

pkg/controllers/monitoring/monitoring-stack/conditions.go
jan--f: Its used in the very next line.

pkg/apis/monitoring/v1alpha1/types.go
jan--f: But isn't there just the one empty LabelSelector? Put :+1: on the `({})` bit.

pkg/controllers/monitoring/monitoring-stack/conditions.go
jan--f: Hm yeah good point.

pkg/apis/monitoring/v1alpha1/types.go
jan--f: Setting a default annotation here has no effect on the generated manifest. I haven't looked into the depth of the OpenAPI code but I suspect `default` and `nullable` doesn't go well together.

sthaha (approved):
lgtm ♥️

PR #224, opened 2022-11-21
doc: release workflow

sthaha:


docs/design/assets/release.uml
```diff
@@ -0,0 +1,69 @@
...
+== candiate release prep ==
+
+== candiate release  ==
+Upstream -> CI: start candidate release (olm-candidate.yaml)
+CI -> Publish: candidate release (olm-publish.yaml)
+CI -> Publish: run olm-publish (channels: development, candidate)
+    activate Publish #FFBBBB
+    Publish -> Publish: login to quay
+    Publish -> Publish: merge olm-catalog branch
+    Publish -> Publish: build and
```
JoaoBraveCoding, line 42: nit: Not sure if the "and" was on purpose but in the UML I think it might be better like this, WDYT?
```suggestion
    Publish -> Publish: build operator image
```

docs/design/assets/release.uml
```diff
@@ -0,0 +1,69 @@
...
+CI -> Upstream: git tag release and git push tag
+CI -> Upstream: publish pre release to GH
+deactivate CI
+== candiate release prep ==
+
+== candiate release  ==
+Upstream -> CI: start candidate release (olm-candidate.yaml)
+CI -> Publish: candidate release (olm-publish.yaml)
+CI -> Publish: run olm-publish (channels: development, candidate)
+    activate Publish #FFBBBB
```
JoaoBraveCoding, line 39: Non-blocking question the difference in colour between this and the  #DarkSalmon on line 62 is supposed to differentiate the candidate and stable versions, correct?

JoaoBraveCoding (commented):
Small suggestion but apart from that lgtm 👍

docs/design/assets/release.uml
sthaha: Thats correct :-) and that it doesn't show the entire flow again

JoaoBraveCoding (commented):
The image also has to be updated to reflect the changes done to the UML code.

That reminds me that we should probably have a pointer or mention on how to do that.

Afterwards I'm happy to leave the lgtm :)

docs/design/assets/release.uml
sthaha: Done , and also noticed that the `candidate` was spelt wrong.

sthaha:
> The image also has to be updated to reflect the changes done to the UML code.
> 
> That reminds me that we should probably have a pointer or mention on how to do that.
> 
> Afterwards I'm happy to leave the lgtm :)

Done, and also noticed that the candidate was spelt wrong.

JoaoBraveCoding (approved):
/lgtm

docs/design/assets/release.uml
```diff
@@ -0,0 +1,72 @@
+@startuml
+'NOTE: use http://sujoyu.github.io/plantuml-previewer to render the Sequence diagram
+
+title Release Workflow
+autonumber
+actor RM order 10
```
jan--f, line 6: Any objection against being explicit and calling this actor `Release Manager`?

jan--f (commented):
Thanks this is a very nice addition. I have two somewhat general questions:

1. Should we be more explicit in the text description in the diagram and does that work for a UML diagram? Especially where users interact with the system I think it might be nice to name PRs and branches that actual people will interact with (even if those names will include the inevitable placeholders). For example step two could `Open PR "release: vX.X.X"` which we could then reference again in step 5 `approve PR "release vX.X.X"`.
2. Should we generate the png in a make file step and maybe even check that it corresponds to the uml file in a CI step?

docs/design/assets/release.uml
JoaoBraveCoding: Not from my side, however I'm not sure how it will look in the generated image through

sthaha:
> Thanks this is a very nice addition. I have two somewhat general questions:

> Should we be more explicit in the text description in the diagram and does that work for a UML diagram? Especially where users interact with the system I think it might be nice to name PRs and branches that actual people will interact with (even if those names will include the inevitable placeholders).

RM imho is acceptable since it is a well know abbreviation 

> For example step two could Open PR "release: vX.X.X" which we could then reference again in step 5 approve PR "release vX.X.X".

yes, this does improve readability we can say something like " `Below is a workflow for releasing 1.2.3` ". I do feel giving a concrete example is better 

> Should we generate the png in a make file step and maybe even check that it corresponds to the uml file in a CI step?


I am not sure if we will get good ROI on that :) . This would require installation (along with maintenance of installation steps) of another tool and since we don't expect changes to happen often (once in a month) in this flow it may not be worth the effort.  When it does, using the online editors to regenerate the uml is easier. - WDYT?

jan--f:
> I am not sure if we will get good ROI on that :) . This would require installation (along with maintenance of installation steps) of another tool and since we don't expect changes to happen often (once in a month) in this flow it may not be worth the effort. When it does, using the online editors to regenerate the uml is easier. - WDYT?

I still think we should be able to generate all artifacts from the repo. PlantUML is a java project so the installation is arguable quite easy (download JAR, run it). It does require java though. With online generators my concern is that these go away (and new ones appear) so that ultimately a contributor who wants to add or change something needs to jump though extra hoops.
If I'd be a user and consider contributing a fix, this would be the moment where I think "I guess this bug stays".

Maybe if PlantUML is to expensive to maintain in our tool chain, its not the right tool?

In any case I don't think this should block this PR. I could live with an issue to change or add this.

sthaha:
> In any case I don't think this should block this PR. I could live with an issue to change or add this.

Thank you :), Merging 


> 
> I still think we should be able to generate all artifacts from the repo. PlantUML is a java project so the installation is arguable quite easy (download JAR, run it). It does require java though. With online generators my concern is that these go away (and new ones appear) so that ultimately a contributor who wants to add or change something needs to jump though extra hoops. If I'd be a user and consider contributing a fix, this would be the moment where I think "I guess this bug stays".
> 
> Maybe if PlantUML is to expensive to maintain in our tool chain, its not the right tool?
> 

I have a different point of view ...  PlantUML is only a means to an end i.e. we need a sequence diagram.  plantuml makes it "easier" to modify the sequence diagram.  Say the tool I used was powerpoint or even hand drawn, would we take a similar approach to this?

jan--f:
> Say the tool I used was powerpoint or even hand drawn, would we take a similar approach to this?

I think we'd have a different discussion in this case and I'd argue strongly against any of these two options.
I just think there is value in automating the up-to-date keeping of documentation artifacts. Happy to follow up in https://github.com/rhobs/observability-operator/issues/228 though.

jan--f (approved):

PR #225, opened 2022-11-21
fix: number of Prometheus replicas can be 0

tremes:
This updates my original PR cbb95f32509ecde346b70eb30bd558483f162e2f to allow 0 number of replicas for the Prometheus config.

### DIFF
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -102,6 +102,9 @@ func TestMonitoringStackController(t *testing.T) {
 	}, {
 		name:     "Verify multi-namespace support",
 		scenario: namespaceSelectorTest,
+	}, {
+		name:     "Verify ability to scale down Prometheus",
+		scenario: prometheusScaleDown,
 	}}
 
 	for _, tc := range ts {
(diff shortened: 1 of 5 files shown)


simonpasquier (dismissed):

simonpasquier:
> then one still has to scale down the Prometheus operator (to scale Prometheus to custom number of replicas - including 0)

it shouldn't be needed. As soon as OBO set `spec.replicas`  to 0 for the Prometheus object, the Prometheus operator will scale the number of Prometheus pods to zero.

I'll defer the final approval to @sthaha and @jan--f :)

sthaha:
@tremes I agree with Simon, that we shouldn't have to create artificial limits for the minimum allowed. IMHO, we should permit all values that PO permits this keeps the interfaces consistent. 
Could we please add a  test for the scenario that Simon mentioned that setting 0 does indeed scale down the pods. E.g. 

* set replica to 1
* validate pod count = 1
* set replica to 0
* validate pod count = 0

sthaha (changes requested):
This looks good to me! 
Requesting change only for a test to be added that assert setting 0 does indeed scale down the pods.

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -517,6 +518,33 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+
+	key := types.NamespacedName{Name: ms.Name, Namespace: ms.Namespace}
+	err = f.K8sClient.Get(context.Background(), key, ms)
+	assert.NilError(t, err, "failed to get a monitoring stack")
+
+	numOfRep = 0
+	ms.Spec.PrometheusConfig.Replicas = &numOfRep
+	err = f.K8sClient.Update(context.Background(), ms)
+	assert.NilError(t, err, "failed to update a monitoring stack")
+	f.GetResourceWithRetry(t, ms.Name, ms.Namespace, &prom)
```
sthaha, line 544: shouldn't we retry if the status takes time to update and then fail?

simonpasquier: I assume that we want at least that generation == observedGeneration (for the MonitoringStack object) before retrieving the Prometheus object?

jan--f (approved):
This lgtm, will leave it to @sthaha to merge though.

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -517,6 +518,45 @@ func assertAlertmanagerReceivesAlerts(t *testing.T) {
...
+	err := f.K8sClient.Create(context.Background(), ms)
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	prom := monv1.Prometheus{}
+	f.GetResourceWithRetry(t, ms.Name, ms.Namespace, &prom)
+
+	assert.Equal(t, prom.Status.Replicas, int32(1))
+
+	key := types.NamespacedName{Name: ms.Name, Namespace: ms.Namespace}
+	err = f.K8sClient.Get(context.Background(), key, ms)
```
sthaha, line 537: nit: it is always safer to use a new blank object.

sthaha (approved):
lgtm, thank you for the patch @tremes 🙇

PR #227, opened 2022-11-30
Package operator artifacts

jan--f:
This PR adds artifacts and make targets to build a package-operator package image (https://github.com/package-operator/package-operator).
The package is setup to install everything into the `observability-operator` namespace for now.
Additionally this setup currently excludes two sets of resources:

- The rules for all operants are not packaged due to https://github.com/package-operator/package-operator/issues/101.
- The admission-webhook related resources are not packaged until we figure out to add TLS certificates. Currently this is handled by OLM.
- No ghub actions yet to build and push the image. I might add that before this merges.

Fixes: https://issues.redhat.com/browse/MON-2831


Makefile
```diff
@@ -211,6 +222,25 @@ catalog-push: catalog-tag-sha ## Push a catalog image.
 	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) $(CATALOG_IMG_LATEST)
 	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) $(CATALOG_IMG_SHA)
 
+## package-operator package
+
+# The image tag given to the resulting package image
+PACKAGE_IMG_BASE ?= $(IMAGE_BASE)-package
+PACKAGE_IMG ?= $(PACKAGE_IMG_BASE):$(VERSION)
```
sthaha, line 229: should we allow overriding this?
 
```suggestion
PACKAGE_IMG = $(PACKAGE_IMG_BASE):$(VERSION)
```

sthaha (dismissed):
Nice that this PR does not introduce any changes to bundle, making review super easy!!

Makefile
```diff
@@ -87,14 +87,25 @@ generate-crds: $(CONTROLLER_GEN) generate-prom-op-crds
 .PHONY: generate-kustomize
 generate-kustomize: $(KUSTOMIZE)
 	cd deploy/operator && \
-		$(KUSTOMIZE) edit set image observability-operator=*:$(VERSION)
+		$(KUSTOMIZE) edit set image observability-operator=$(OPERATOR_IMG)
```
sthaha, line 90: With this change we should be able to remove the `set image` below

Makefile
```diff
@@ -87,14 +87,25 @@ generate-crds: $(CONTROLLER_GEN) generate-prom-op-crds
...
+
+.PHONY: generate-package-resources
+generate-package-resources: $(KUSTOMIZE) generate-kustomize
+	cd deploy/package-operator && \
+		rm -r package/crds package/dependencies package/operator ;\
```
sthaha, line 95: nit 
```suggestion
		rm -rf package/crds package/dependencies package/operator ;\
```

deploy/package-operator/operator/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: observability-operator
+images:
+- name: observability-operator
+  newName: local-registry:30000/observability-operator/observability-operator
+  newTag: 0.0.16
```
sthaha, line 8: is this right? 🤔 
Also how is the version/tag updated ?

sthaha: Should also `make package` and `git diff --exit`  in our CI checks to ensure that packages are up to date?

sthaha (commented):
I have approved so that this doesn't block the release but please do consider addressing the comments.

deploy/package-operator/operator/kustomization.yaml
jan--f: > is this right? thinking Also how is the version/tag updated ?

The tag is updated via `kustomize set image` iiuc.

> Should we also make package and git diff --exit in our CI checks to ensure that packages are up to date?

We don't actually store any package artifacts. The resources are generated with kustomize and an image is created from that. So no need to check if any generated files have changed.

slashpai (approved):

PR #231, opened 2022-12-05
build: generate and publish package in ghub actions

jan--f:
This adds github actions to generate and publish the package image to quay.

Versioning and publishing policy follows what we do for the OLM artifacts with one difference:

Since package-operator does not have a notion of channels, this currently only pushes packages for development versions (with the `$tag-$date` tagging scheme and for releases (i.e. not for prereleases).

If we want packages for pre-releases, we need to discuss an addition to the tagging policy.

https://issues.redhat.com/browse/MON-2892

### DIFF
--- a/.github/package-operator-publish/action.yaml
+++ b/.github/package-operator-publish/action.yaml
@@ -1,0 +1,44 @@
+name: 'Publish package-operator package'
+description: 'Publishes the operator as a package-operator package'
+inputs:
+  quay_login:
+    description: "Quay login"
+    required: true
+  quay_token:
+    description: "Quay token"
+    required: true
+  go-version:
+    description: "go version"
+    required: true
+runs:
+  using: composite
+  steps:
+  - uses: actions/checkout@v3
+    with:
+      fetch-depth: 0
+
+  - name: Setup Go environment
+    uses: actions/setup-go@v3
+    with:
+      go-version: ${{ inputs.go-version }}
+
+  - name: Use go cache
+    uses: ./.github/go-cache
+
+  - name: Install tools
+    uses: ./.github/tools-cache
+
+  - name: Registry Login
+    uses: docker/login-action@v2
+    with:
+      registry: quay.io
+      username: ${{ inputs.quay_login }}
+      password: ${{ inputs.quay_token }}
+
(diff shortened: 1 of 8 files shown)


jan--f:
side note: the github action linter seems to only run against newly added actions. The check in this PR complaint about a few things, that are present in existing actions.

.github/package-operator-publish/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+      username: ${{ inputs.quay_login }}
+      password: ${{ inputs.quay_token }}
+
+  - name: Generate yaml files
+    shell: bash
+    run: make generate
+
+  - name: Generate Package image
+    shell: bash
+    run: make package
```
sthaha, line 44: does it make sense to have `generate` as a dependency of `package` make target given that we require generate for making up to date `package` ?

sthaha (dismissed):
Looks good to me. 
May want to consider addressing the comment.

.github/package-operator-publish/action.yaml
jan--f: Yeah good point.

jan--f:
I added an additional commit that builds a second package. In this package the operators mount a secret and use the kubeconfig in that secret.
There might be a more elegant way of handing that, but for now this should do what we want.
https://issues.redhat.com/browse/MON-2893 has more information.

deploy/package-operator/operator-kubeconfig/patches/kubeconfig-volume.yml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+    app.kubernetes.io/part-of: observability-operator
+    name: observability-operator
+spec:
+  template:
+    spec:
+      volumes:
+        - name: kubeconfig
+          secret:
+            defaultMode: 420
```
sthaha, line 13: how about 400 instead of 420 ? doesn't `2` mean allow allow write to group? `rwx`

sthaha (changes requested):
Minor correction in Makefile. Otherwise looks good!

sthaha (approved):

PR #233, opened 2022-12-06
chore: improve kind setup UX

sthaha:
Previously, when running hack/kind/setup.sh, the script ran its validations such as if the tools are installed, local-registry is entry is added to /etc/hosts etc in the middle of the run instead of the start. This required the cluster to be deleted again and again until all validations pass.

This commit improve the experience
  * running validations in the starting of the script
  * ensuring binaries in <project-root>/tmp/bin takes precedence all the time (as opposed to only when kubeclt was missing)
  * Better emojis 🔔 ✅ 🛑

```
❯ ./hack/kind/setup.sh

  🔆🔆🔆  Validating Prerequisites  🔆🔆🔆
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 🛑 operator-sdk not found - did you run 'make tools'?

 🛑 /etc/hosts does not contain local-registry entry
 🔔 No local-registry entry in hosts; run:
    ❯ echo "127.0.0.1	local-registry" | sudo tee -a /etc/hosts

 ✋ fix errors above and rerun the script
──────────────────── ⛔️⛔️⛔️ ────────────────────────
```

### DIFF
--- a/hack/kind/setup.sh
+++ b/hack/kind/setup.sh
@@ -6,25 +6,78 @@ declare -r SCRIPT_DIR=$(cd $(dirname "$SCRIPT_PATH") && pwd)
 declare -r PROJECT_ROOT_DIR="$SCRIPT_DIR/../../"
 declare -r OP_NAME="obs-operator"
 
+# use tools installed to tmp/bin
+export PATH="$PROJECT_ROOT_DIR/tmp/bin/:$PATH"
+
 ### create a separate config file for kind clusters
 export KUBECONFIG=~/.kube/kind/obs-operator
 
-err() {
-  echo -e "ERROR: $@" >&2
+header(){
+  local title="🔆🔆🔆  $*  🔆🔆🔆 "
+
+  local len=40
+  if [[ ${#title} -gt $len ]]; then
+    len=${#title}
+  fi
+
+  echo -e "\n\n  \033[1m${title}\033[0m"
+  echo -n "━━━━━"
+  printf '━%.0s' $(seq "$len")
+  echo "━━━━━"
 }
 
-info() {
-  echo -e "INFO: $@"
+info(){
+  echo " 🔔 $*"
 }
 
-die() {
-  echo -e "FATAL: $@" >&2
+ok(){
+  echo " ✅ $*"
+}
+
+err(){
+  echo " 🛑 $*"
+}
+
+die(){
+  echo -e "\n ✋ $* "
+  echo -e "──────────────────── ⛔️⛔️⛔️ ────────────────────────\n"
   exit 1
+}
+
+line(){
+  local len="$1"; shift
+
+  echo -n "────"
+  printf '─%.0s' $(seq "$len")
+  echo "────────"
+}
+
+validate_prerequisites(){
+  header "Validating Prerequisites"
+
+  local fail=0
+
+  [[ -x $PROJECT_ROOT_DIR/tmp/bin/operator-sdk  ]] || {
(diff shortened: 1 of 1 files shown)


hack/kind/setup.sh
```diff
@@ -6,50 +6,97 @@ declare -r SCRIPT_DIR=$(cd $(dirname "$SCRIPT_PATH") && pwd)
...
+  [[ -x $PROJECT_ROOT_DIR/tmp/bin/operator-sdk  ]] || {
+    err "operator-sdk not found - did you run 'make tools'?"
+    fail=1
+  }
+
+  grep -q local-registry /etc/hosts || {
+    err "/etc/hosts does not contain local-registry entry"
+
+    info "No local-registry entry in hosts; run:"
+    echo -e "    ❯ echo \"127.0.0.1\tlocal-registry\" | sudo tee -a /etc/hosts"
```
slashpai, line 69: Should the script add the entry instead of asking user?

hack/kind/setup.sh
```diff
@@ -61,11 +108,15 @@ run_registry() {
     http://local-registry:30000 || {
     die "Failed to reach local-registry running"
   }
-  echo -e "      ---------------------------------- \n"
+  line 50
 
 }
 
 setup_cluster() {
+  header "Setting up Cluster"
```
slashpai, line 116: can we add a check if cluster already exists use same cluster instead of failing. This will help if script was failed at some point and we need to re-run script

```
  🔆🔆🔆  Setting up Cluster  🔆🔆🔆
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
ERROR: failed to create cluster: node(s) already exist for a cluster with the name "obs-operator"
```

hack/kind/setup.sh
```diff
@@ -78,17 +129,20 @@ setup_cluster() {
 }
 
 install_kubectl() {
-    if ! command -v kubectl &> /dev/null; then
+  header "Installing kubectl"
+
+  if ! command -v kubectl &> /dev/null; then
     info "kubectl not found, attempting to install"
     mkdir -p tmp/bin
     curl -o tmp/bin/kubectl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
slashpai, line 137: should we consider mac users too? ;) check device type and install based on type of device

slashpai (commented):
A few comments after trying the updated script

hack/kind/setup.sh
sthaha: sure, perhaps in a separate patch.. 
Patches welcome :p

JoaoBraveCoding (approved):
Looks good from my side 👍 Great work!

JoaoBraveCoding:
🤦 didn't see that auto-merge was enabled...

hack/kind/setup.sh
slashpai: I can create a patch for this no worries 😎

slashpai:
> 🤦 didn't see that auto-merge was enabled...

no problem it looks good to me as well just had a few comments but I agree with Sunil's point :)

sthaha:
I should unlearn the habit of enabling auto-merge!  I do that to ensure the commit message when merging looks good! 
The good thing is that it keep discussions short 😂 

Please feel free to comment and I am happy to address in subsequent PRs if any.

PR #234, opened 2022-12-07
feat: upgrade PO to 0.61.0-rhobs1

sthaha:


simonpasquier (commented):
The failure might be related to the fact that the admission webhook's service inside the pod is now exposed on port 443 rather than 8443?

sthaha:
@simonpasquier 

> The failure might be related to the fact that the admission webhook's service inside the pod is now exposed on port 443 rather than 8443?

Thank you for the pointer. I had noticed it. The issue was that I was modifying only the prometheusrule validating webhook to  adapt to the changes to admission-webhook but not the alertmanager-config validating webhook and the service generated by OLM resulted in conflict. 

I have fixed that now by 
* removing the handcoded service
* adding a comment to the future maintainers to modify both validating-webhooks and not just one
*

jan--f (dismissed):
:+1: Thanks!

deploy/dependencies/admission-webhook/prometheus-rule-validating-webhook.yaml
```diff
@@ -12,18 +12,18 @@ webhooks:
...
     service:
-      # NOTE: the name and namespace will be replaced by OLM
+      # NOTE: when changing the service, ensure the same changes are applied
+      # to alertmanager-config-validating-webhook as well
       name: obo-prometheus-operator-admission-webhook
       namespace: operators
       path: /admission-prometheusrules/validate
-      port: 8443
-  failurePolicy: Ignore
+  failurePolicy: Fail
```
simonpasquier, line 20: I wouldn't change the failure policy as part of this PR. 'Fail' means that applying PrometheusRule will be rejected when the admission webhook service is unavailable or unreachable. We haven't found a good solution for this in CMO for instance.

deploy/dependencies/admission-webhook/alertmanager-config-validating-webhook.yaml
```diff
@@ -12,18 +12,18 @@ webhooks:
...
-      # NOTE: the name and namespace will be replaced by OLM
+      # NOTE: when changing the service, ensure the same changes are applied
+      # to prometheus-rule-validating-webhook
       name: obo-prometheus-operator-admission-webhook
       namespace: operators
       path: /admission-alertmanagerconfigs/validate
-      port: 8443
-  failurePolicy: Ignore
   name: alertmanagerconfigs.monitoring.rhobs
+  failurePolicy: Fail
```
sthaha, line 21:
```suggestion
  failurePolicy: Ignore
```

deploy/dependencies/admission-webhook/prometheus-rule-validating-webhook.yaml
sthaha, line 20:
```suggestion
  failurePolicy: Ignore
```

simonpasquier (approved):

PR #237, opened 2022-12-15
build: remove custom cache and use setup-go@v3 caching

jan--f:
Otherwise the two compete and we get tons of errors like
Error: /usr/bin/tar: ../../../go/pkg/mod/fybrik.io/crdoc@v0.5.2/pkg/functions/functions.go: Cannot open: File exists
E.g.: https://github.com/rhobs/observability-operator/actions/runs/3704737122/jobs/6278551


simonpasquier (approved):
lgtm

PR #238, opened 2022-12-16
build: create branch for olm-publish action

jan--f:
Problem: In the make target `catalog-tag-sha` we want to use main's sha sum to tag the catalog image. The olm-publish action however adds commits to the local main branch before that happens, resulting in a sha tag that doesn't correspond to main.

Solution: The olm-publish action creates a branch to add commits in.

### DIFF
--- a/.github/olm-publish/action.yaml
+++ b/.github/olm-publish/action.yaml
@@ -13,10 +13,6 @@ inputs:
 runs:
   using: composite
   steps:
-  - uses: actions/checkout@v3
-    with:
-      fetch-depth: 0
-
   - name: Setup Go environment
     uses: actions/setup-go@v3
     with:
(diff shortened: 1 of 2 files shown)


jan--f:
This PR imposes the assumption that CI should not change the local main branch.

An alternative solution would be to specifically get the sha of `origin/main`, though that assumes that the remote if called `origin`.

Not sure what the better approach is, happy to hear others opinions.

JoaoBraveCoding (commented):
So IIUC here the commits done by `olm-publish` actions would go to a different branch but this is not a problem because we are also not interested in them, we just want the action itself to occur, correct?

If so with what I know I think I'm happy with this approach 👍 

Is there any other downside I'm not seeing?

jan--f:
The current workflow is:

- Checkout `main`
- Merge `olm-catalog`
- Update catalog artifacts and commit
- Push the result to remote `olm-catalog` branch

The change proposed here is the do all steps after "Checkout `main`" in its own branch, so that local `main` reflects remote `main` at all times.

JoaoBraveCoding:
Got it! Then I had understood it incorrectly. Sounds good to me 👍

JoaoBraveCoding (approved):
lgtm

Issue #212, opened 2022-10-31
Add shellcheck lint


sthaha:
Add shellcheck linter to CI job so that scripts are kept in good condition

PR #239, opened 2022-12-22
build: add make lint-shell and add to PR checks

jan--f:
This add a `shellcheck` make target and PR test. Additionall this fixes existing shellcheck issues.

Fixes: https://github.com/rhobs/observability-operator/issues/212

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ all: operator
 ## Development
 
 .PHONY: lint
-lint: lint-golang lint-jsonnet
+lint: lint-golang lint-jsonnet lint-shell
 
 .PHONY: lint-golang
 lint-golang: $(GOLANGCI_LINT)
@@ -29,6 +29,10 @@ lint-jsonnet: $(JSONNET_LINT) jsonnet-vendor
 		-o -name '*.libsonnet' -print \
 		-o -name '*.jsonnet' -print \
 	| xargs -n 1 -- $(JSONNET_LINT) -J $(JSONNET_VENDOR)
+
+.PHONY: lint-shell
+lint-shell:
+	find -name "*.sh" -print0 | xargs --null $(SHELLCHECK)
 
 .PHONY: fmt-jsonnet
 fmt-jsonnet: $(JSONNETFMT) jsonnet-vendor
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -50,10 +50,13 @@ JB_VERSION = v0.4.0
 ## NOTE: gojsontoyaml does not have any releases, so we use a fake version starting with v0.0.1
 # thus to upgrade/invalidate the github cache, increment the value
 GOJSONTOYAML = $(TOOLS_DIR)/gojsontoyaml
-GOJSONTOYAML_VERSRION = 0.0.1
+GOJSONTOYAML_VERSION = 0.0.1
 
 JSONNET_VENDOR = jsonnet/vendor
 JSONNETFMT_ARGS = -n 2 --max-blank-lines 2 --string-style s --comment-style s
+
+SHELLCHECK = $(TOOLS_DIR)/shellcheck
+SHELLCHECK_VERSION = 0.9.0
 
 $(TOOLS_DIR):
 	@mkdir -p $(TOOLS_DIR)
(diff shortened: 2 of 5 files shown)


Makefile.tools
```diff
@@ -50,11 +50,14 @@ JB_VERSION = v0.4.0
...
 # thus to upgrade/invalidate the github cache, increment the value
 GOJSONTOYAML = $(TOOLS_DIR)/gojsontoyaml
-GOJSONTOYAML_VERSRION = 0.0.1
+GOJSONTOYAML_VERSION = 0.0.1
 
 JSONNET_VENDOR = jsonnet/vendor
 JSONNETFMT_ARGS = -n 2 --max-blank-lines 2 --string-style s --comment-style s
 
+SHELLCHECK = $(TOOLS_DIR)/shellcheck
+SHELLCHECK_VERSION = 0.9.0
```
JoaoBraveCoding, line 59: Not quite sure why we set a version here if we always curl for the latest stable release 🤔 Is the goal that `make shellcheck` fails when a new version comes out and this way we know we have to update this env var?

JoaoBraveCoding (commented):
looks good! just one minor nit

Makefile.tools
jan--f: This is indeed a little weird. We use the tool versions for two purposes:

1. Some tools download in specific versions indeed.
2. We use the versions in the github actions to signal when to tool install needs to be redone. The versions are piped into a file and that files hash is used to tag the tools directory. This tools dir is reused until a version changes. That changes the file, which changes the hash, which triggers an action to reinstall the tools.

And yes when a new version is released the install target for `shellcheck` will fail and let us know to update the version.

hack/kind/setup.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 set -e -u -o pipefail
 
-declare -r SCRIPT_PATH=$(readlink -f "$0")
-declare -r SCRIPT_DIR=$(cd $(dirname "$SCRIPT_PATH") && pwd)
+declare -r SCRIPT_PATH
+SCRIPT_PATH=$(readlink -f "$0")
+declare -r SCRIPT_DIR
```
sthaha, line 6: This should cause the script to fail since we are modifying the variable that is declared readonly. I think it is safe to turn off this lint than assign and then mark the variable readonly. 

See: https://github.com/koalaman/shellcheck/issues/900

Makefile.tools
```diff
@@ -185,6 +188,19 @@ $(GOJSONTOYAML) gojsontoyaml: $(TOOLS_DIR)
 		GOBIN=$(TOOLS_DIR) go install github.com/brancz/gojsontoyaml@latest ;\
 	}
 
+.PHONY: shellcheck
+$(SHELLCHECK) shellcheck: $(TOOLS_DIR)
+	@{ \
+		set -ex ;\
+		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
+		cd $$(mktemp -d) ;\
+		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz";\
```
sthaha, line 197: How about we use the same version as the one declared above to be consistent ? 

`https://github.com/koalaman/shellcheck/releases/download/$(VERSION)/shellcheck-v$(VERSION).linux.x86_64.tar.xz`

Makefile.tools
sthaha: I think we can use the pinned version to construct the url 

`https://github.com/koalaman/shellcheck/releases/download/v$(SHELLCHECK_VERSION)/shellcheck-v$(SHELLCHECK_VERSION).linux.x86_64.tar.xz`

Makefile
```diff
@@ -30,6 +30,10 @@ lint-jsonnet: $(JSONNET_LINT) jsonnet-vendor
 		-o -name '*.jsonnet' -print \
 	| xargs -n 1 -- $(JSONNET_LINT) -J $(JSONNET_VENDOR)
 
+.PHONY: lint-shell
+lint-shell:
+	find -name "*.sh" -print0 | xargs --null shellcheck
```
sthaha, line 35: Shouldn't we use $(SHELLCHECK) instead of `shellcheck` to make use of the shellcheck installed to tmp/bin?

sthaha (changes requested):
I think this patch breaks hack/kind/setup.sh.  Could you please confirm that the script runs fine?

hack/kind/setup.sh
jan--f: Right, good catch. I kinda assumed `shellcheck` would catch this but :shrug:

Makefile.tools
jan--f: Sure we could but we don't really depend on a specific `shellcheck` version. The version tracking is really just to support the ghub action tool caching.
This way we'll automatically pull in a new version when its released and we only need to update the version. If we pin it we might end up using an outdated version for a long time when we don't really need to.
The same already happens for other tools (`oc`, `jb` and several others)...in some cases even without an version check. I'd rather propose we add the version check to the other tools and only pin to a particular version if required.

sthaha: For the record I am not against this patch from merging since it already adds value. 

However, I feel we should pin the version for consistency sake and not needing to upgrade the tools at random.  I  do agree there is a downside to pinning version but the advantage is that we will have consistent behaviour across all environments (dev, ci); i.e. .. if it fails on CI, it would fail on dev as well. 

> This way we'll automatically pull in a new version when its released and we only need to update the version.

Not really the case, since tools will only be updated if the .github/tools file change

More importantly, unrelated changes shouldn't break CI .. e.g. given the uncertainty that the tool version can change at anytime (whenever tools cache is invalidated), we may have a situation where you update a tool to fix an issue but would also have to fix unrelated shell script. 

> The same already happens for other tools (oc, jb and several others)...in some cases even without an version check. 

I consider this  more as an advantage to pinning the version, i.e. we focus on addressing the issues that matter and CI doesn't break because `oc` or `jb` has a new release mode.

sthaha (approved):
lgtm

Makefile.tools
jan--f: I created #241 to follow up on this.

PR #243, opened 2023-01-08
build: add option to pass SHA to tag catalog

jan--f:
Problem: In bc3b0d202f0ea36c04a20fbcd44a51d2504d2aed the git rev-parse command was changed to tag catalog images with the `main` commit on which this catalog image is build. The argument to rev-parse was changed from `HEAD` to `main`. In some workflows however `main` might not be a valid local revision.
To reporduce:
```
cd $(mktemp -d)
git init observability-operator
git remote add origin https://github.com/rhobs/observability-operator
cd observability-operator
git remote add origin https://github.com/rhobs/observability-operator
git fetch --prune origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
git checkout --force refs/tags/v0.0.18
git rev-parse main
```

Solution: Add a new Makefile variable `RELEASH_SHA`. If its not defined we use `origin/main`, which should provide backwards compatibility. In the github action workflows we now set this explicitly to `GITHUB_SHA`, which corresponds to the commit that triggered a workflow.

### DIFF
--- a/.github/workflows/olm-candidate.yaml
+++ b/.github/workflows/olm-candidate.yaml
@@ -33,6 +33,7 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
           CHANNELS: candidate,development
           DEFAULT_CHANNEL: candidate
+          RELEASE_SHA: ${GITHUB_SHA}
         with:
           quay_login: ${{ secrets.QUAY_LOGIN }}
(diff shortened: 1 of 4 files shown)


jan--f:
This should fix the recent failures of https://github.com/rhobs/observability-operator/actions/workflows/olm-stable.yaml

simonpasquier (approved):

PR #253, opened 2023-01-23
Several deployment resource fixes

jan--f:
Originally the motivation was to fix the package-operator build, but this ended up being a bit larger, fixing multiple issues. I tried to make reviewing easier by splitting it up in sensible commits.

1. Only the OLM bundle needs the minimal prometheus-operator CRDs, so move if there.
2. Generate and track the forked p-o CRD manifests in `deploy/crds/kubernetes/`
3. Before this PR the OLM bundle would include `monitoring.coreos.com` prometheus rules and a service monitor. This required some contortions for testing (our test scripts would pre-seed CRDs). I have changed that here to include only the `monitoring.rhobs` rules, as we don't need the service monitor, OBO self-monitors through an additional scrape config.
I'm open to shipping the `monitoring.coreos.com` rules and service monitor, but it needs to be in a way that doesn't prevent the bundle from installing, of those CRD manifests are not present in the cluster.
4. Don't pre-seed CRD manifests for out e2e tests, since that practice can hide issues with out bundle build process.


deploy/olm/kustomization.yaml
```diff
@@ -2,17 +2,32 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+# FATA[0020] Failed to run bundle
+# : create catalog
+# : error creating registry pod: error building registry pod definition
+# : configMap error: error updating ConfigMap
+# : error creating ConfigMap
+# : ConfigMap "operator-sdk-run-bundle-config" is invalid
+# : []: Too long: must have at most 1048576 bytes
+- https://raw.githubusercontent.com/rhobs/obo-prometheus-operator/v0.62.0-rhobs1/example/stripped-down-crds/all.yaml
 - ../crds/common
+# NOTE: swap the follwoing line for the two above  when the bug above is solved
```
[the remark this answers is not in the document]
jan--f: I just moved this line, so not 100% sure, but I believe https://github.com/operator-framework/operator-lifecycle-manager/issues/1523 is what this refers to.

sthaha:
@jan--f I lack some context about the issue. Could you please help shed some light into it? 

> Before this PR the OLM bundle would include monitoring.coreos.com prometheus rules and a service monitor. This required some contortions for testing (our test scripts would pre-seed CRDs). I have changed that here to include only the monitoring.rhobs rules, as we don't need the service monitor, OBO self-monitors through an additional scrape config.

How can OBO Self Monitor? ObO doesn't deploy a stack when installed. So, shouldn't OBO rely on platform monitoring to monitor itself (i.e. the operator)?

deploy/crds/kubernetes/monitoring.rhobs_alertmanagerconfigs.yaml
```diff
@@ -5,17 +5,15 @@ metadata:
...
   creationTimestamp: null
-  name: alertmanagerconfigs.monitoring.coreos.com
+  name: alertmanagerconfigs.monitoring.rhobs
 spec:
-  group: monitoring.coreos.com
+  group: monitoring.rhobs
   names:
     categories:
-    - prometheus-operator
+    - rhobs-prometheus-operator
```
tremes, line 13: Not sure how much the categories are important, but I think we usually call it `obo-prometheus-operator` or?

tremes: Actually it seems we use/mix both names...

jan--f: This comes from https://github.com/rhobs/obo-prometheus-operator/ and is a result of our rename script. This might be something to track down there, less so in this PR.

deploy/crds/kubernetes/kustomization.yaml
sthaha, line 1: I think we should rename `kubernetes` dir to `prometheus-operator` or `rhobs-po` since the intent of the `kubernetes` dir was to keep all CRDs / resources that aren't in vanilla k8s cluster but are present in openshift.

deploy/crds/kubernetes/kustomization.yaml
```diff
@@ -2,12 +2,14 @@
...
- - ../common
- - monitoring.coreos.com_alertmanagerconfigs.yaml
- - monitoring.coreos.com_alertmanagers.yaml
- - monitoring.coreos.com_podmonitors.yaml
- - monitoring.coreos.com_probes.yaml
- - monitoring.coreos.com_prometheuses.yaml
- - monitoring.coreos.com_prometheusrules.yaml
- - monitoring.coreos.com_servicemonitors.yaml
- - monitoring.coreos.com_thanosrulers.yaml
+  - ../common
```
sthaha, line 5: I think it is easier to understand if the structure was something like 
```
crds
├── kustomization.yaml
├── operator
│   └── kustomization.yaml
└── rhobs-po
    └── kustomization.yaml

```
wdyt?

jan--f: :+1:  I was wondering about the naming. I didn't change things due to lack of context.

deploy/monitoring/observability-operator-rules.yaml
```diff
@@ -1,9 +1,5 @@
 ---
-# NOTE: For PrometheusRule ObO uses platform monitoring in OpenShift
-# instead of the forked PO's CRDs. As the ObO bundle ships the forked PO's CRD,
-# these won't be available at the time OLM installs the bundle
-
-apiVersion: monitoring.coreos.com/v1
```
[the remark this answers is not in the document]
jan--f: Not sure I follow...
The bundle contains the `monitoring.rhobs` CRDs. Shouldn't that suffice? Fwiw the e2e tests pass locally for me. Not sure why the tests fail in CI.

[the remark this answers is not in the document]
sthaha: [AFAIK, OLM installs can't refer to operator's own crds (see failure in CI). ](https://github.com/rhobs/observability-operator/actions/runs/4446674080/jobs/7807274076?pr=253#step:4:555) . I.E if an operator add a CRD Foobar, then its csv can't create resource `xyz` of kind `Foobar` because when the operator installs the bundle, it tries to create `xyz` and fails due to missing `Foobar` crd. 

The reason this passes on your machine could be because you already have these CRDs added. 

We should see if there is a workaround for this. This was also one of the historic reason why we relied on incluster monitoring for ObO. If ObO fails, then nothing will give us alerts.

PR #257, opened 2023-03-01
improve namespace and resource selector docs

lihongyan1:
fix description of `namespaceSelector` and `resourceSelector` to unify the empty and null

### DIFF
--- a/deploy/crds/common/monitoring.rhobs_monitoringstacks.yaml
+++ b/deploy/crds/common/monitoring.rhobs_monitoringstacks.yaml
@@ -56,9 +56,10 @@ spec:
                 - error
                 type: string
               namespaceSelector:
-                description: Namespace selector for Monitoring Stack Resources. If
-                  left empty the Monitoring Stack will only match resources in the
-                  namespace it was created in.
+                description: 'Namespace selector for Monitoring Stack Resources. To
+                  monitor everything, set to empty map selector. E.g. namespaceSelector:
+                  {}. To monitor resources in the namespace where Monitoring Stack
+                  was created in, set to null. E.g. namespaceSelector:.'
                 properties:
                   matchExpressions:
                     description: matchExpressions is a list of label selector requirements.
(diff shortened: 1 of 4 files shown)


deploy/crds/common/monitoring.rhobs_monitoringstacks.yaml
```diff
@@ -56,9 +56,10 @@ spec:
                 - error
                 type: string
               namespaceSelector:
-                description: Namespace selector for Monitoring Stack Resources. If
-                  left empty the Monitoring Stack will only match resources in the
-                  namespace it was created in.
+                description: 'Namespace selector for Monitoring Stack Resources. Set
```
lihongyan1, line 59: @sthaha Not sure why the `single quote` is added after description

sthaha: I don't think this matters as long as the string is valid. (and the test pass)

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -58,14 +58,15 @@ type MonitoringStackSpec struct {
 	LogLevel LogLevel `json:"logLevel,omitempty"`
 
 	// Label selector for Monitoring Stack Resources.
-	// Set to the empty LabelSelector ({}) to monitoring everything.
-	// Set to null to disable service discovery.
+	// Set to the empty label selector (resourceSelector: {}) to monitoring everything.
+	// Set to null (resourceSelector:) to disable service discovery.
```
sthaha, line 62:
```suggestion
	// To monitor everything, set to a empty map selector. E.g. resourceSelector: {}
	// To disable service discovery, set to null. E.g. resourceSelector:
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -58,14 +58,15 @@ type MonitoringStackSpec struct {
...
+	// To monitor everything, set to an empty map selector. E.g. resourceSelector: {}
+	// To disable service discovery, set to null. E.g. resourceSelector:
 	// +optional
 	// +nullable
 	ResourceSelector *metav1.LabelSelector `json:"resourceSelector"`
 
 	// Namespace selector for Monitoring Stack Resources.
-	// If left empty the Monitoring Stack will only match resources in the namespace it was created in.
+	// Set to null (namespaceSelector:) the Monitoring Stack will only monitor resources in the namespace it was created in.
+	// Set to the empty namespace selector (namespaceSelector: {}) means monitor everything.
```
sthaha, line 69: Could you please change this as well to match the style above? 
E.g. 
```
// To only monitor resources in the same namespace as the stack, set this to `null`. E.g. namespaceSelector: 
// To monitor resources in all namespaces, set the selector to empty map ({}). E.g. namespaceSelector: {}

sthaha (approved):

PR #258, opened 2023-03-02
feat: extend OBO conditions to read & propagate Alertmanager conditions

tremes:
This extends existing `conditions.go` to read and cover Alertmanager status if the Alertmanager is deployed in the monitoring stack.

### DIFF
--- a/pkg/controllers/monitoring/monitoring-stack/conditions.go
+++ b/pkg/controllers/monitoring/monitoring-stack/conditions.go
@@ -9,36 +9,45 @@ import (
 )
 
 const (
-	AvailableReason                = "MonitoringStackAvailable"
-	ReconciledReason               = "MonitoringStackReconciled"
-	FailedToReconcileReason        = "FailedToReconcile"
-	PrometheusNotAvailable         = "PrometheusNotAvailable"
-	PrometheusNotReconciled        = "PrometheusNotReconciled"
-	PrometheusDegraded             = "PrometheusDegraded"
-	ResourceSelectorIsNil          = "ResourceSelectorNil"
-	CannotReadPrometheusConditions = "Cannot read Prometheus status conditions"
-	AvailableMessage               = "Monitoring Stack is available"
-	SuccessfullyReconciledMessage  = "Monitoring Stack is successfully reconciled"
-	ResourceSelectorIsNilMessage   = "No resources will be discovered, ResourceSelector is nil"
-	ResourceDiscoveryOnMessage     = "Resource discovery is operational"
-	NoReason                       = "None"
+	AlertmanagerNotAvailable         = "AlertmanagerNotAvailable"
+	AlertmanagerDegraded             = "AlertmanagerDegraded"
+	AlertmanagerNotReconciled        = "AlertmanagerNotReconciled"
+	AvailableReason                  = "MonitoringStackAvailable"
+	ReconciledReason                 = "MonitoringStackReconciled"
+	FailedToReconcileReason          = "FailedToReconcile"
+	PrometheusNotAvailable           = "PrometheusNotAvailable"
+	PrometheusNotReconciled          = "PrometheusNotReconciled"
+	PrometheusDegraded               = "PrometheusDegraded"
+	ResourceSelectorIsNil            = "ResourceSelectorNil"
+	CannotReadPrometheusConditions   = "Cannot read Prometheus status conditions"
+	CannotReadAlertmanagerConditions = "Cannot read Alertmanager status conditions"
+	AvailableMessage                 = "Monitoring Stack is available"
+	SuccessfullyReconciledMessage    = "Monitoring Stack is successfully reconciled"
+	ResourceSelectorIsNilMessage     = "No resources will be discovered, ResourceSelector is nil"
+	ResourceDiscoveryOnMessage       = "Resource discovery is operational"
+	NoReason                         = "None"
 )
 
-func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, recError error) []v1alpha1.Condition {
+func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, alertManager *monv1.Alertmanager, recError error) []v1alpha1.Condition {
 	return []v1alpha1.Condition{
 		updateResourceDiscovery(ms),
-		updateAvailable(ms.Status.Conditions, prom, ms.Generation),
-		updateReconciled(ms.Status.Conditions, prom, ms.Generation, recError),
+		updateAvailable(ms.Status.Conditions, prom, alertManager, ms.Generation),
+		updateReconciled(ms.Status.Conditions, prom, alertManager, ms.Generation, recError),
 	}
 }
 
-func getMSCondition(conditions []v1alpha1.Condition, t v1alpha1.ConditionType) (v1alpha1.Condition, error) {
+func getMSCondition(conditions []v1alpha1.Condition, t v1alpha1.ConditionType) v1alpha1.Condition {
 	for _, c := range conditions {
 		if c.Type == t {
-			return c, nil
+			return c
 		}
 	}
-	return v1alpha1.Condition{}, fmt.Errorf("condition type %v not found", t)
+	return v1alpha1.Condition{
+		Type:               t,
+		Status:             v1alpha1.ConditionUnknown,
+		Reason:             NoReason,
+		LastTransitionTime: metav1.Now(),
+	}
 }
 
 // updateResourceDiscovery updates the ResourceDiscoveryCondition based on the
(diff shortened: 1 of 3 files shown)


tremes:
I am not sure this is the best approach to do that, but I don't have anything better so far.....

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -9,36 +9,45 @@ import (
...
+	CannotReadAlertmanagerConditions = "Cannot read Alertmanager status conditions"
+	AvailableMessage                 = "Monitoring Stack is available"
+	SuccessfullyReconciledMessage    = "Monitoring Stack is successfully reconciled"
+	ResourceSelectorIsNilMessage     = "No resources will be discovered, ResourceSelector is nil"
+	ResourceDiscoveryOnMessage       = "Resource discovery is operational"
+	NoReason                         = "None"
 )
 
-func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, recError error) []v1alpha1.Condition {
+func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, alertManager *monv1.Alertmanager, recError error) []v1alpha1.Condition {
```
sthaha, line 31: Lets keep the convention the same as `monv1.Prometheus` argument

```suggestion
func updateConditions(ms *v1alpha1.MonitoringStack, prom monv1.Prometheus, alertmanager monv1.Alertmanager, recError error) []v1alpha1.Condition {
```

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
```diff
@@ -13,6 +13,7 @@ func TestUpdateAvailable(t *testing.T) {
 	tt := []struct {
 		name               string
 		prometheus         monv1.Prometheus
+		alertmanager       *monv1.Alertmanager
```
sthaha, line 16:
```suggestion
		alertmanager       monv1.Alertmanager
```

tremes: Hi @sthaha! Any reason on why you resolved this conversation? :) I actually think it's a good point...it would make the tests more clear IMO

sthaha: I thought you might have wanted to set the variable to nil. I also wanted to do a proper review of the whole patch after checking it out locally.

sthaha:
@simonpasquier @tremes  Does it make sense to make the stack as partially available if **only** the Alertmanager is down?

tremes:
@sthaha I don't think so. User can choose to disable the Alertmanager and why he/she should have partially available stack when she/he decided so?

pkg/controllers/monitoring/monitoring-stack/conditions_test.go
tremes: Yes....but I am thinking that it shouldn't be nil in every test...so I can update

sthaha:
@tremes, If the user that disabled AM, then there is no question of partial availability since the conditions are based on what prometheus is but in cases where both are requested, and Prometheus deployment works but some other fails, then the stack is still partially available.

tremes:
@sthaha Ahh OK...didn't get it. Sorry. Yes in this case it makes more sense. Let me think about it, but I am open to suggestions :)

pkg/controllers/monitoring/monitoring-stack/conditions.go
JoaoBraveCoding, line 31: nit: I think that here both prom and alertmanager should be pointers. Eventually, ObO will support other monitoring O11y signals so there might a future where Prometheus might also be disabled. I'm also happy with them both becoming references as long as we stay consistent.

JoaoBraveCoding (commented):
After a long sabbatical not reviewing PRs for this repo I want to start helping more <3

pkg/controllers/monitoring/monitoring-stack/conditions.go
sthaha: This was my initial comment as well which I deleted because this usage has a  semantic meaning. i.e. Prometheus isn't an optional component.

pkg/controllers/monitoring/monitoring-stack/conditions.go
```diff
@@ -9,36 +9,45 @@ import (
...
 	return []v1alpha1.Condition{
 		updateResourceDiscovery(ms),
-		updateAvailable(ms.Status.Conditions, prom, ms.Generation),
-		updateReconciled(ms.Status.Conditions, prom, ms.Generation, recError),
+		updateAvailable(ms.Status.Conditions, prom, alertManager, ms.Generation),
+		updateReconciled(ms.Status.Conditions, prom, alertManager, ms.Generation, recError),
```
sthaha, line 35: I feel if possible we should improve this pattern so something on the lines of 

```
conditions := ms.Status.Conditions
conditions = updateResourceDiscovery(conditions, ms)
conditions = updateAlertManagerCondition(conditions, alertmanager, ms.Generation)
conditions = updatePrometheusCondition(conditions, promethues, ms.Generation) 
conditions = updateReconcilationError(conditions, recError)

return conditions
```

I think we can generalise this further by using a condition-setter interface but the  idea is that every step takes as input the previous conditions and decides to update if the severity is higher than existing and return back the updated condition.

WDYT?

pkg/controllers/monitoring/monitoring-stack/conditions.go
JoaoBraveCoding: I see, I had seen your comment but misunderstood the intent behind it 🤦‍♂️ and now that I think about it I'm wrong because if users provide a MonitoringStack they should at least want Prometheus right 🤔

pkg/controllers/monitoring/monitoring-stack/conditions.go
JoaoBraveCoding: What do we mean by "update if the severity is higher than existing" that's the only part that I'm concerned about 👀 How would we codify that?

sthaha: Say Monitoring stack has 3 components (or 5 some time in future) 

- prometheus 
- alertmanager (optional)
-  thanos (optional)

We need to update the condition of the stack by reading each of the component's status and update 
* Available
* Reconciled

Currently we pass updateAvailable(prometheus, am, thanos) where the func is somewhat like 
```
func updateAvailable(pm, am, t) {
  if am {
      updateAMAvailable(...)
   }
  if t {
     updateThanosAvailable(...)
  }
  updatePrometheusAvailable
}
```
Say AM is a Degraded, thanos is Unavailable, and Prometheus is Degraded.  What should the status of Stack be? 
I see 2 issues with this. The function takes the responsibility of computing the `available` condition based on all 3 components and it is unclear what comes out of it if you look at the code above. Adding a new component means modifying this (already) huge function. 

Instead of this, I think if you followed the pattern above of passing previous conditions to the next in chain, then you would get something like 
```
conditions := emptyConditions()
conditions = updateResourceDiscovery(conditions, ms) 
conditions = updateThanosCondition(conditions, thanos, ms.Generation) 
conditions = updateAlertManagerCondition(conditions, alertmanager, ms.Generation) 
conditions = updatePrometheusCondition(conditions, promethues, ms.Generation) 
conditions = updateReconcilationError(conditions, recError)
```
Each component updates its condition based on previous components conditions (passed in as input) and the precedence here is that `Prometheus > AlertManager > Thanos`. Each of these methods would look at the input `condition` and overwrite it based on `NotReconciled < Degraded < Unavailable`. 

I will modify this PR a bit to demonstrate what I mean, further.

tremes: Thanks for thorough description @sthaha. I agree in general but I think it would be good to define all the new states and coresponding conditions first or if everyone is OK with the precedence `Prometheus > AlertManager > Thanos` then fine, but what we will do when there's a new component? Will it be always at the end of the "chain"?

sthaha: @tremes , indeed! I think as such this pr adds value. The refactor (which I am trying out) can follow later. 
Lets go ahead with shipping the feature since I don't expect anything to change due to refactor.

sthaha: > Will it be always at the end of the "chain"?

Yes, the last one has the say.  If an `updateX(..)` needs to know the condition set by `updateY` before, then those need to be explictly passed in as `updateX(..., conditionsSetByUpdateY)` and that makes the dependencies explicit.

sthaha (commented):
Looks good to me.
But I will let @simonpasquier take another look before merging.

JoaoBraveCoding (approved):
lgtm, deferring to @simonpasquier  for the final merge

simonpasquier (commented):
I'm not sure if OBO should  only report the first "failed" object or if it should combine both of them?
For instance  if Prometheus reports Available=False and Alertmanager Available=Degraded, the MonitoringStack's Available condition would be False with reason AlertmanagerDegraded.
One way would be to join the different reasons (e.g. `AlertmanagerDegradedPrometheusUnavailable`)

tremes:
This is not sufficient and ideal solution. Closing. This will need to be reworked.

PR #261, opened 2023-03-27
Fix to log events when run e2e test with ns not default value

lihongyan1:
Run e2e test with ns parameter "openshift-obo", events will not be logged


sthaha (commented):
lgtm! Thanks for the patch @lihongyan1

sthaha:
@lihongyan1 CI fails because of the commit message format.  How about 
```
fix: use OPERATORS_NS instead of hardcoded namespace
```

lihongyan1:
@sthaha All the CI pass now, please approve when convenient.

test/run-e2e.sh
```diff
@@ -300,7 +300,7 @@ update_cluster_mon_crds() {
   # CRDs are absent and in that case, try creating and fail if that fails
 
   kubectl replace -k deploy/crds/kubernetes ||
-    kubectl create -k deploy/crds/kubernetes || return 1
+    kubectl create -k deploy/crds/kubernetes || return 0
```
sthaha, line 303: why is this change required? and seems unrelated to the topic?

sthaha, line 303: why is this change required? and seems unrelated to the topic?

sthaha (changes requested):

sthaha (approved):
Thanks for the patch @lihongyan1 🙇

PR #262, opened 2023-03-29
ci: package osd e2e tests

sthaha:
Package tests to execution in OSD pipelines. 
See: https://github.com/openshift/osde2e-example-test-harness

Related issue: https://issues.redhat.com/browse/MON-3125

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ RELEASE_SHA ?= $(shell git rev-parse origin/main)
 OPERATOR_IMG = $(IMAGE_BASE):$(VERSION)
 OPERATOR_BUNDLE=observability-operator.v$(VERSION)
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
+OSD_E2E_TEST_HARNESS_IMG=$(IMAGE_BASE)-test-harness:$(VERSION)
+OSD_E2E_TEST_HARNESS_IMG_LATEST=$(IMAGE_BASE)-test-harness:latest
 
 # running `make` builds the operator (default target)
 all: operator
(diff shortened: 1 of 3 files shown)


test/Dockerfile
simonpasquier, line 1: I see that https://github.com/openshift/osde2e-example-test-harness/blob/main/Dockerfile uses multi-stage builds: the first stage build the test binary and the second and final stage produces an image that contains ony the test binary. Can we do the same?

Makefile
```diff
@@ -11,6 +11,7 @@ RELEASE_SHA ?= $(shell git rev-parse origin/main)
 OPERATOR_IMG = $(IMAGE_BASE):$(VERSION)
 OPERATOR_BUNDLE=observability-operator.v$(VERSION)
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
+OSD_E2E_IMG=$(IMAGE_BASE)-osde2e:$(VERSION)
```
jan--f, line 14: I'd propose we inlcude `test-harness` in the image name. I [proposed `observability-operator-test-harness`](https://issues.redhat.com/browse/MON-3125?focusedId=22000201&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22000201), but for explicitness sake maybe even `observability-operator-osde2e-test-harness` makes sense?

sthaha: I felt it verbose but wanted to make `osd` explicit.  If we are to follow their nomenclature, I think we should call it `osde2e-observability-operator-test-harness` see:
```
          - name: ADDON_TEST_HARNESSES
            value: quay.io/rmundhe_oc/osde2e-example-test-harness // update to your test harness image
```

I will follow the same.

sthaha:
To test this locally, I have the kind setup and then 

```
❯ make osd-e2e-test-image
❯ podman run --rm -it --net=host -v $PWD:/src:Z --env 'KUBECONFIG=/src/obs-operator' localhost/observability-operator-test-harness:0.0.20
```
I am not sure if we should have a version? or just use `latest` for the image tag.  Thoughts @jan--f @simonpasquier ? 

Once the patch goes in, I will work on making the image available in quay.io (i.e. publish image just like all other images)

test/Dockerfile
sthaha: Thanks! indeed it is possible!  I assumed `go` didn't allow tests to be compiled into binaries but in hindsight I should have put a bit more thought into it :D

jan--f:
Re image tagging: the example test harness can use different channels, as detailed in https://github.com/openshift/osde2e-example-test-harness#parameters, specified via the `CHANNELS` env var. I'm not quite sure what they mean by "lets you specify the Cincinnati channel for version selection." though. will follow up in Slack.

So for now I'm happy with this, we might need to change our tagging approach later though.

jan--f (approved):
lgtm

PR #264, opened 2023-04-07
feat: scripts to run e2e tests against OCP

lihongyan1:


test/run-e2e-ocp.sh
```diff
@@ -0,0 +1,103 @@
...
+declare OPERATORS_NS
+OPERATORS_NS="openshift-operators"
+declare NO_INSTALL=false
+declare NO_UNINSTALL=false
+declare SHOW_USAGE=false
+
+install_obo_OCP(){
+  oc apply -f ./hack/olm/catalog-src.yaml
+  oc apply -f ./hack/olm/subscription.yaml
+}
```
sthaha, line 15: we should wait for the catalog-src to be ready before subscribing. 
We should also wait for the subscription to be complete before exiting the function.

test/run-e2e-ocp.sh
```diff
@@ -0,0 +1,103 @@
...
+    then
+      ret=0
+    else
+      echo "Failed to install OBO"
+      exit 1
+    fi
+    sleep 10
+    wait_time=$((wait_time+10))
+    oc -n openshift-operators rollout status deploy/obo-prometheus-operator || ret=1
+  done
```
sthaha, line 32: could you please follow the same coding convention as other scripts? 
```
while [[ ... ]] ; do 
  if [[ ...]]; then
  fi
done
```

test/run-e2e-ocp.sh
```diff
@@ -0,0 +1,103 @@
...
+  local ret=0
+  local -i wait_time=0
+  oc -n openshift-operators rollout status deploy/obo-prometheus-operator || ret=1
+  while [[ ret -eq 1 ]]; do
+    if [[ wait_time -lt 60 ]];
+    then
+      ret=0
+    else
+      echo "Failed to install OBO"
+      exit 1
```
sthaha, line 27: scripts should try to have a single exit point. So it is better to use `return` so that main can fail / handle it.

sthaha (approved):

Issue #266, opened 2023-04-11
fix `managedFields.manager` name in generated resources


vimalk78:
All Obo generated resources (e.g. `Prometheus`) have `fieldManager` set to same value of `ownerReference`

`Prometheus` managed fields example:

```yaml
managedFields:
- apiVersion: monitoring.rhobs/v1
  fieldsType: FieldsV1
  fieldsV1:
    f:metadata:
      f:labels:
        f:app.kubernetes.io/managed-by: {}
        f:app.kubernetes.io/name: {}
        f:app.kubernetes.io/part-of: {}
      f:ownerReferences:
        k:{"uid":"81da0d9a-61aa-4df3-affc-71015bcbde5a"}: {}
    f:spec:
      f:additionalScrapeConfigs: {}
      f:affinity:
        f:podAntiAffinity:
          f:requiredDuringSchedulingIgnoredDuringExecution: {}
      f:alerting:
        f:alertmanagers: {}
      f:arbitraryFSAccessThroughSMs: {}
      f:logLevel: {}
      f:podMetadata:
        f:labels:
          f:app.kubernetes.io/component: {}
          f:app.kubernetes.io/part-of: {}
      f:podMonitorSelector: {}
      f:replicas: {}
      f:resources:
        f:limits:
          f:cpu: {}
          f:memory: {}
        f:requests:
          f:cpu: {}
          f:memory: {}
      f:retention: {}
      f:ruleSelector: {}
      f:rules:
        f:alert: {}
      f:securityContext:
        f:fsGroup: {}
        f:runAsNonRoot: {}
        f:runAsUser: {}
      f:serviceAccountName: {}
      f:serviceMonitorSelector: {}
      f:thanos:
        f:baseImage: {}
        f:resources: {}
        f:version: {}
      f:tsdb: {}
  manager: obo-demo/sample-monitoring-stack    << should be ObservabililtyOperator
  operation: Apply
```

The resources where field manager needs to be updated are:
```
Secret/sample-monitoring-stack-prometheus-additional-scrape-configs
Service/sample-monitoring-stack-alertmanager
Service/sample-monitoring-stack-prometheus
Service/sample-monitoring-stack-thanos-sidecar
Alertmanager/sample-monitoring-stack
Prometheus/sample-monitoring-stack
PodDisruptionBudget/sample-monitoring-stack-alertmanager
PodDisruptionBudget/sample-monitoring-stack-prometheus
[...]

PR #267, opened 2023-04-11
fix: field manager name for generated resources

vimalk78:
Fixes the  field `manager` name of the generated resources

fixes: #266

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -34,8 +34,7 @@ func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime
 		}
 	}
 
-	owner := fmt.Sprintf("%s/%s", r.resourceOwner.GetNamespace(), r.resourceOwner.GetName())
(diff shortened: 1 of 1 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -34,8 +34,7 @@ func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime
 		}
 	}
 
-	owner := fmt.Sprintf("%s/%s", r.resourceOwner.GetNamespace(), r.resourceOwner.GetName())
-	if err := c.Patch(ctx, r.resource, client.Apply, client.ForceOwnership, client.FieldOwner(owner)); err != nil {
+	if err := c.Patch(ctx, r.resource, client.Apply, client.ForceOwnership, client.FieldOwner("ObservabilityOperator")); err != nil {
```
sthaha, line 37:
```suggestion
	if err := c.Patch(ctx, r.resource, client.Apply, client.ForceOwnership, client.FieldOwner("observability-operator")); err != nil {
```

vimalk78: updated

sthaha (approved):

PR #269, opened 2023-04-13
fix(test):fix clean of ocp is executed twice when INT e2e run

lihongyan1:
fix clean of ocp is executed twice when interrupt e2e run and some comments format

### DIFF
--- a/test/run-e2e-ocp.sh
+++ b/test/run-e2e-ocp.sh
@@ -12,7 +12,7 @@ declare -r PROJECT_ROOT
 source "$PROJECT_ROOT/test/lib/utils.bash"
 
 # NOTE: openshift-operators is the namespace used in subscription.yaml to install
-# obo, so this is harded coded for the test as well.
+# ObO, so this is harded coded for the test as well.
 declare -r OPERATORS_NS="openshift-operators"
 
 ### Configuration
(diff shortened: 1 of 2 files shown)


test/run-e2e-ocp.sh
```diff
@@ -22,25 +22,26 @@ declare SHOW_USAGE=false
 
 cleanup() {
 	# NOTE: clears trap INT
-	trap - INT
+	info "Cleaning up ocp..."
 
 	# skip cleanup if user requested help
 	$SHOW_USAGE && return 0
 	delete_obo
+	return 1
```
sthaha, line 30: why are we returning 1 ? `return 0` mean everything worked as normal. Any non-zero indicates error

test/run-e2e.sh
```diff
@@ -28,6 +28,7 @@ cleanup() {
 	# shell check  ignore word splitting when using jobs -p
 	# shellcheck disable=SC2046
 	[[ -z "$(jobs -p)" ]] || kill $(jobs -p) || true
+	return 1
```
sthaha, line 31: we should return 1 here as well.

test/run-e2e-ocp.sh
```diff
@@ -22,25 +22,26 @@ declare SHOW_USAGE=false
 
 cleanup() {
 	# NOTE: clears trap INT
-	trap - INT
+	info "Cleaning up ocp..."
```
sthaha, line 25: header instead of "info" :)

test/run-e2e-ocp.sh
lihongyan1: Usually we interrupt a program run when found sth abnormal

test/run-e2e.sh
lihongyan1: If we don't return 1 here, `run-e2e-ocp.sh` calls `run-e2e.sh` and suppose `run-e2e.sh` end normally and continue it's run and execute clean, the trap invoke clean function also

test/run-e2e-ocp.sh
lihongyan1: `cleanup()` of `run-e2e.sh` has `info`, suppose here should be consistent

sthaha:
@lihongyan1 , can we close this one since #271 got merged?

PR #271, opened 2023-04-13
fix: e2e test cleanup and exit code

sthaha:
Previously, the cleanup of ocp test will be run twice; i.e.
  1. One after the tests are run
  2. When the script exits

Additionally both test scripts failed to return the actual test run return code and only returned 1 on failure. This patch fixes both issues.

### DIFF
--- a/test/run-e2e-ocp.sh
+++ b/test/run-e2e-ocp.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e -u -o pipefail
 
-trap cleanup INT
+trap cleanup EXIT
 
 # NOTE: install ObO and run e2e against the installation
 
@@ -21,12 +21,11 @@ declare NO_UNINSTALL=false
 declare SHOW_USAGE=false
 
 cleanup() {
-	# NOTE: clears trap INT
-	trap - INT
-
 	# skip cleanup if user requested help
 	$SHOW_USAGE && return 0
-	delete_obo
+
+	delete_obo || true
+	return 0
 }
 
 install_obo() {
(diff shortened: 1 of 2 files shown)


test/run-e2e-ocp.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e -u -o pipefail
 
-trap cleanup INT
+trap cleanup EXIT
```
lihongyan1, line 4: Trap only EXIT, if I interrupt run not during `run-e2e.sh`, `cleanup` in `run-e2e-ocp.sh` will not be called.

test/run-e2e-ocp.sh
```diff
@@ -123,9 +122,9 @@ main() {
 	wait_for_operators_ready "$OPERATORS_NS"
 
 	local -i ret=0
-	./test/run-e2e.sh --no-deploy --ns "$OPERATORS_NS" || ret=1
-	delete_obo
+	./test/run-e2e.sh --no-deploy --ns "$OPERATORS_NS" || ret=$?
```
lihongyan1, line 125: suppose 
`./test/run-e2e.sh --no-deploy --ns "$OPERATORS_NS" || ret=1`
and
`./test/run-e2e.sh --no-deploy --ns "$OPERATORS_NS" || ret=$?`
have the same effect

test/run-e2e-ocp.sh
sthaha: It will be called on `exit` and INT will result in `exit`

test/run-e2e-ocp.sh
sthaha: no, the newer version gives you the exact return code instead of `1` all the time.

jan--f (approved):
lgtm

Issue #273, opened 2023-04-19
fix obo image version in CSV


vimalk78:
Obo CSV refers to observability operator image version

```
                image: observability-operator:0.0.17
```

https://github.com/rhobs/observability-operator/blob/main/bundle/manifests/observability-operator.clusterserviceversion.yaml#L519-L527

Need to generate correct version in CSV.

PR #274, opened 2023-04-19
fix: updated Obo image version in CSV

vimalk78:
fix: updated Obo image version in CSV
fix: increase timeout for installing OLM on kind cluster

fixes: #273

### DIFF
--- a/deploy/package-operator/operator/kustomization.yaml
+++ b/deploy/package-operator/operator/kustomization.yaml
@@ -3,7 +3,8 @@ kind: Kustomization
 namespace: observability-operator
 images:
 - name: observability-operator
-  newTag: 0.0.17
+  newName: observability-operator
+  newTag: 0.0.20
 commonAnnotations:
   package-operator.run/phase: operator
 resources:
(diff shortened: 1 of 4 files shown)


hack/kind/setup.sh
```diff
@@ -93,7 +93,7 @@ label_infra_node() {
 
 setup_olm() {
   header "Install OLM"
-  "$PROJECT_ROOT_DIR"/tmp/bin/operator-sdk olm install
```
sthaha, line 96: Unrelated change :), perhaps a separate PR 🙏

deploy/package-operator/operator/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Kustomization
 namespace: observability-operator
 images:
 - name: observability-operator
-  newTag: 0.0.17
```
sthaha, line 6: when the version gets bumped to 0.0.21, would we have to do this manually? If so, can we automate this somehow in Makefile?

vimalk78: updated

sthaha (approved):

PR #276, opened 2023-05-04
[OSD-16025] Add scrape interval parameter to the prometheus config

zmird-r:
- Needed for improving performance on hypershift management clusters

Ticket: [Increase OBO global scrape interval from 30s to 60s](https://issues.redhat.com/browse/OSD-16025)
Epic: [OBO Performance Improvements](https://issues.redhat.com/browse/SDE-2823)


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -184,6 +184,10 @@ type PrometheusConfig struct {
 	// Enable Prometheus to be used as a receiver for the Prometheus remote write protocol. Defaults to the value of `false`.
 	// +optional
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
+    // Scrape interval for prometheus.
```
simonpasquier, line 187:
```suggestion
    // Default interval between scrapes.
```

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -172,6 +172,7 @@ func newPrometheus(
 				RemoteWrite:               config.RemoteWrite,
 				ExternalLabels:            config.ExternalLabels,
 				EnableRemoteWriteReceiver: config.EnableRemoteWriteReceiver,
+				ScrapeInterval:            config.ScrapeInterval,
```
simonpasquier, line 175: If we make it a pointer then we need to test for nil.

zmird-r: Added a check for nil :)

simonpasquier:
we have special requirements about commit message format :) 
https://github.com/rhobs/observability-operator/blob/main/docs/developer.md#commit-message-convention

You need also to run `go fmt ./...` and `make generate`.

zmird-r:
Thank you for the suggestions Simon. I merged all the commits into one with the correct message format.

sthaha (approved):
Looks great! 
Thank you for the patch @zmird-r

PR #280, opened 2023-05-06
test(managedFields): added test case to assert Obo managed fields of Prometheus object

vimalk78:
Update this test case when a new Prometheus field is managed by Observability Operator.

### DIFF
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/rhobs/observability-operator/test/e2e/framework"
 
+	"golang.org/x/exp/slices"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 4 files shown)


test/e2e/monitoring_stack_controller_test.go
```diff
@@ -592,6 +595,128 @@ func prometheusScaleDown(t *testing.T) {
...
+	have := objmap["f:spec"]
+
+	var expected map[string]interface{}
+	_ = json.Unmarshal([]byte(OboManagedFieldsJson), &expected)
+
+	assert.DeepEqual(t, have, expected)
+}
+
+// Update this json when a new Prometheus field is set by MonitoringStack
+const OboManagedFieldsJson = `
```
sthaha, line 658:
```suggestion
const oboManagedFieldsJson = `
```

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -592,6 +595,128 @@ func prometheusScaleDown(t *testing.T) {
...
+	assert.NilError(t, err, "failed to create a monitoring stack")
+
+	prom := monv1.Prometheus{}
+	f.GetResourceWithRetry(t, ms.Name, ms.Namespace, &prom)
+
+	mf := prom.GetManagedFields()
+
+	var oboManagedFields *metav1.ManagedFieldsEntry
+	for _, e := range mf {
+		if e.Manager == "observability-operator" {
```
sthaha, line 635: replace "observabitliy-operator" with a const from pkg/controller ?

sthaha (approved):
Thanks @vimalk78

PR #293, opened 2023-05-25
feat: expose prometheus nodeselector and tolerations in monitoringstack CRD

bng0y:
Allows to isolate the Prometheus pods when necessary. 
Context: https://issues.redhat.com/browse/SDE-2823


simonpasquier (commented):
Maybe the fields should be located under `.spec` rather than `.spec.prometheusConfig` and apply to all child resources of the MonitoringStack resource?

@sthaha @jan--f WDYT?

sthaha:
> Maybe the fields should be located under `.spec` rather than `.spec.prometheusConfig` and apply to all child resources of the MonitoringStack resource?

@simonpasquier, I like the idea to have a global setting for all child resources, however should we also allow each child resource to have different setting to the global one?

sthaha:
> If this PR changes the Prometheus `managedFields`, please update [this test case](https://github.com/rhobs/observability-operator/blob/main/test/e2e/monitoring_stack_controller_test.go#L601C1-L656) to keep managed Fields info updated.

@vimalk78 shouldn't that test fail in that case 🤔

vimalk78:
> @vimalk78 shouldn't that test fail in that case 
it would, but the new parameters are optional, so unless explicitly set in the object the managedFields wont change.

simonpasquier:
@sthaha my gut feeling is that in most cases, you want the monitoring pods to run on the same set of nodes (e.g. infra) so having a top-level field is probably what makes the most sense. We can argue later if there's a need for individual settings for each component but I'd rather make the mainstream situation easy to configure.

sthaha:
@simonpasquier  +1 to having it global for now 

> We can argue later if there's a need for individual settings for each component but I'd rather make the mainstream situation easy to configure.

+1 to YAGNI :)

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -172,6 +172,8 @@ func newPrometheus(
 				RemoteWrite:               config.RemoteWrite,
 				ExternalLabels:            config.ExternalLabels,
 				EnableRemoteWriteReceiver: config.EnableRemoteWriteReceiver,
+				NodeSelector:              config.NodeSelector,
+				Tolerations:               config.Tolerations,
```
sthaha, line 176: @bng0y , Like @simonpasquier suggested, lets make this global and apply the same to all resources that ObO creates.

sthaha (changes requested):
Could you please consider implementing the suggestion.

sthaha:
@bng0y , It would be great to have this feature added to ObO. Do you need help implementing this?

PR #301, opened 2023-06-05
fix(doc): update release doc typo

lihongyan1:


docs/design/assets/release.uml
```diff
@@ -37,7 +37,7 @@ deactivate CI
 == candidate release  ==
 Upstream -> CI: start candidate release (olm-candidate.yaml)
 CI -> Publish: candidate release (olm-publish.yaml)
-CI -> Publish: run olm-publish (channels: development, candidate)
+CI -> Publish: run olm-publish (channel: candidate)
```
sthaha, line 40: @lihongyan1 , the original doc is right, when 
* code merges to main -> update development channel 
* create pre-release -> update development, candidate channels
* create release -> update development, candidate, stable channels 

so .. if someone has 
* subscribed to `development` channel they get all updates (merge to main, pre-release, release)
* subscribed to `candidate` channel, then get pre-releaes and stable release
* subscribed to `stable` channel they get only stable releases. 
*

sthaha (changes requested):
this isn't the workflow :(

sthaha (changes requested):
this isn't the workflow :(

lihongyan1:
> this isn't the workflow :(

thanks, will update

sthaha (approved):
Thanks for the doc fix, @lihongyan1 🙇

PR #304, opened 2023-06-26
Make catalogsource compatible with restricted SCC enforcement

NautiluX:
* Restricted SCC enforcement will be added with OCP 4.14
* Updating the catalogsource to allow the operator to get deployed
* Clusters that don't support the setting (<4.12) will ignore it

Jira: OSD-15621


sthaha:
@NautiluX Thanks a lot for the patch 🤗 . Could you please fix the commit message format and resubmit ?

sthaha (approved):
Thanks for the patch @NautiluX

PR #308, opened 2023-07-05
ppc64le/multi-arch support for observability-operator

jcho02:
Providing ppc64le (IBM Power) support towards observability-operator.
build.sh is a script that builds and pushes multi-arch supported images for the operator (operator, bundle, and catalog images)
build-and-publish.yaml is a github workflow action that runs build.sh 
Dockerfile and Makefile modified to build multi-arch operator and bundle image


.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  build-and-publish:
+    name: Build and Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
sthaha, line 12: could we please continue to use podman instead?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,45 @@
...
+        uses: docker/setup-buildx-action@v2
+      
+      - name: Install yq
+        env:
+          VERSION: v4.14.2
+          BINARY: yq_linux_amd64
+        run: |
+          wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O /usr/local/bin/yq
+          chmod +x /usr/local/bin/yq
+
```
sthaha, line 33: Our convention is that all CI builds should be reproducible locally as well, thus all necessary tools should installable using `make tools` target.  It may be worth understanding why some tools are needed (to reduce the maintenance effort)

Makefile
```diff
@@ -199,7 +199,7 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
 
 .PHONY: bundle-image
 bundle-image: bundle ## Build the bundle image.
-	$(CONTAINER_RUNTIME) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	docker buildx build --push --platform linux/amd64,linux/ppc64le -f bundle.Dockerfile -t quay.io/rhobs/observability-operator-bundle:0.0.23 .
```
sthaha, line 202: I think we should 
* use podman 
* should use $(BUNDLE_IMAGE)

build/Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20 as builder
```
sthaha, line 2: nit:
```suggestion
FROM --platform=$BUILD_PLATFORM golang:1.20 as builder
```

build/Dockerfile
```diff
@@ -1,6 +1,7 @@
...
 
+ARG TARGETOS TARGETARCH
```
sthaha, line 4: nit
```suggestion
ARG TARGET_OS TARGET_ARCH
```

build/Dockerfile
```diff
@@ -14,7 +15,7 @@ COPY cmd/operator/main.go cmd/operator/main.go
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager cmd/operator/main.go
+RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -o manager cmd/operator/main.go
```
sthaha, line 18:
```suggestion
RUN CGO_ENABLED=0 GOOS=$TARGET_OS GOARCH=$TARGET_ARCH go build -a -o manager cmd/operator/main.go
```

hack/build.sh
```diff
@@ -0,0 +1,89 @@
+# To push to your own registry, override the REGISTRY and NAMESPACE env vars,
```
sthaha, line 1:
```suggestion
#! /usr/bin/env bash

# To push to your own registry, override the REGISTRY and NAMESPACE env vars,
```

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+#
+# REQUIREMENTS:
+#  * a valid login session to a container registry.
+#  * `docker`
+#  * `jq`
+#  * `yq`
+#  * `opm`
+#  * `skopeo`
+
+set -e
```
sthaha, line 13: lets be a bit more strict about scripts :) 
```suggestion
set -eu -o pipefail
```

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+export OPERATOR_NAME='observability-operator'
+export REGISTRY=${REGISTRY:-'quay.io'}
+export NAMESPACE=${NAMESPACE:-'rhobs'}
+export TAG=${TAG:-'0.0.23'}
```
sthaha, line 18: I think it is best if the script explicitly accepted a TAG or use `VERSION` file
```
./hack/build.sh <TAG>
```

sthaha, line 18: Should these be exported? How about `declare -r` ?

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+export CSV_PATH=${CSV_PATH:-'bundle/manifests/observability-operator.clusterserviceversion.yaml'}
+export ANNOTATIONS_PATH=${ANNOTATIONS_PATH:-'bundle/metadata/annotations.yaml'}
+
+function cleanup {
```
sthaha, line 22: nit: could you please use the same convention used in other scripts in the project? 

```suggestion
cleanup() {
```

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# prints pre-formatted info output.
+function info {
+	echo "INFO $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
+
+# prints pre-formatted error output.
+function error {
+	>&2 echo "ERROR $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
```
sthaha, line 40: Could you please consider reusing the functions in https://github.com/rhobs/observability-operator/blob/main/test/lib/utils.bash 
Perhaps we should move the functions to hack/lib instead.

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+function digest() {
```
sthaha, line 42:
```suggestion
digest() {
```

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+	declare -n ret=$2
```
sthaha, line 43:
```suggestion
	local -n ret=$2
```

hack/build.sh
sthaha, line 1: Could you please restructure the script to follow the convention used elsewhere in the project 
E.g. 
* https://github.com/rhobs/observability-operator/blob/main/test/run-e2e.sh
* https://github.com/rhobs/observability-operator/blob/main/hack/kind/setup.sh

```
#!/usr/bin/env bash 
set -eu -o pipefail

# traps 


# exports <if needed> 
# avoid exports if possible and make use of 
   FOOBAR=X \ 
   BAR=Baz \
   yq eval -i ...

# Globals 
declare FOO_BAR=${FOOBAR:-"default"}

# Constants 
declare -r READONLY_VAR="something"


# functions 
foo() {
}


main() {
}

main "$@"
```

Keeping all actions in seperate functions makes it easier to reason with

hack/build.sh
```diff
@@ -0,0 +1,89 @@
...
+	IMAGE=$1
+	docker pull "${IMAGE}"
+	# shellcheck disable=SC2034
+	ret=$(docker inspect --format='{{index .RepoDigests 0}}' "${IMAGE}")
+}
+
+docker buildx build --push --platform "linux/amd64,linux/ppc64le" -f build/Dockerfile -t "${REGISTRY}/${NAMESPACE}/observability-operator:${TAG}" .
```
sthaha, line 50:
```suggestion
docker buildx build --push \
  --platform "linux/amd64,linux/ppc64le" \
   -f build/Dockerfile \
   -t "$REGISTRY/$NAMESPACE/$OPERATOR_NAME:${TAG}" .
```

sthaha, line 50: I am not sure if we should push at this stage.

build/Dockerfile
jcho02: Doesn't read $BUILD_PLATFORM as an argument

build/Dockerfile
jcho02: Doesnt read TARGET_OS and TARGET_ARCH as arguments when using `--platform linux/amd64,...`

hack/build.sh
jcho02: What is the difference between the two?

sthaha:
@jan--f , @simonpasquier, could you please take a look at this patch ?

jcho02:
Updated build commands to use `make tools` and podman. Operator image must use `docker buildx` however, Dockerfile doesn't read platform values from podman in multi-arch builds

jcho02:
Please review, operator builds and runs on Power platforms

jan--f:
Sorry for the latency, I'll review this next week.

.github/workflows/build-and-publish.yaml
danielmellado: +1 on podman usage.

Makefile
```diff
@@ -136,7 +136,7 @@ build:
 
 .PHONY: operator-image
 operator-image: generate
-	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
+	docker buildx build --push --platform linux/amd64,linux/ppc64le -f build/Dockerfile -t $(OPERATOR_IMG) .
```
danielmellado, line 139: please keep the $(CONTAINER_RUNTIME) var here

Makefile.tools
```diff
@@ -194,16 +197,40 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
 		set -ex ;\
 		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
 		cd $$(mktemp -d) ;\
-		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz";\
+		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/v0.9.0/shellcheck-v0.9.0.linux.x86_64.tar.xz" ;\
```
danielmellado, line 200: I'd avoid hardocoding anything here, specially given that you declared the SHELLCHECK_VERSION above, please do use it here as well, thanks!

Makefile.tools
```diff
@@ -194,16 +197,40 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
...
 		tar -xJf shellcheck-stable.tar.xz ;\
-		cp shellcheck-stable/shellcheck $(SHELLCHECK) ;\
+		cp shellcheck-v0.9.0/shellcheck $(SHELLCHECK) ;\
```
danielmellado, line 202: same here

danielmellado (changes requested):
Left a few comments over there, besides that it lgtm, thanks!

Makefile
jcho02: Can't build multi-arch images using podman

.github/workflows/build-and-publish.yaml
jcho02: Need to use docker for QEMU action as docker can successfully run multi-arch builds at once without the use of using manifests

jcho02:
@danielmellado multi-arch operator images fully build on Podman

danielmellado:
Thanks, let's see if CI is happy again ;)

hack/build.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# prints pre-formatted error output.
+error() {
+	>&2 echo "ERROR $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
+
+digest() {
+	local -n ret=$2
+	IMAGE=$1
+	podman pull "${IMAGE}"
```
danielmellado, line 47: is this specific to podman? could we use just the same CONTAINER_RUNTIME env var as above for this?

hack/build.sh
```diff
@@ -0,0 +1,122 @@
...
+error() {
+	>&2 echo "ERROR $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
+
+digest() {
+	local -n ret=$2
+	IMAGE=$1
+	${CONTAINER_RUNTIME} pull "${IMAGE}"
+	# shellcheck disable=SC2034
+	ret=$(podman inspect --format='{{index .RepoDigests 0}}' "${IMAGE}")
```
danielmellado, line 50: I guess we should also change the specific podman reference here on line 50 ;)

PR #309, opened 2023-07-06
feat: bump up PO to 0.66.0-rhobs1

sthaha:
Only for test


marioferh:
done here, https://github.com/rhobs/observability-operator/pull/319
Thanks Sunil

Issue #311, opened 2023-07-13
case "TestMonitoringStackController/Verify_multi-namespace_support" failed on OCP 4.14


lihongyan1:
```
./test/run-e2e-ocp.sh
......
=== RUN   TestMonitoringStackController/Verify_multi-namespace_support
    monitoring_stack_controller_test.go:850: assertion failed: error is not nil: failed to create demo app test-ns-1/prometheus: pods "prometheus" is forbidden: violates PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (container "prometheus" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "prometheus" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "prometheus" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "prometheus" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost"): test-ns-1: deploying demo app failed
--- FAIL: TestMonitoringStackController (382.73s)
```
% oc version
Client Version: 4.13.1
Kustomize Version: v4.5.7
Server Version: 4.14.0-0.nightly-2023-07-11-092038
Kubernetes Version: v1.27.3+af29f64

It's a automation code issue, will fix it following solution
https://access.redhat.com/solutions/7002730

lihongyan1:
After applying fix in solution https://access.redhat.com/solutions/7002730
Get the following error
```
% oc logs prometheus
......
ts=2023-07-14T02:48:02.307Z caller=query_logger.go:91 level=error component=activeQueryTracker msg="Error opening query log file" file=/prometheus/queries.active err="open /prometheus/queries.active: permission denied"
panic: Unable to create mmap-ed active query log

goroutine 1 [running]:
github.com/prometheus/prometheus/promql.NewActiveQueryTracker({0x7ffe3121adca, 0xb}, 0x14, {0x3aba480, 0xc000859540})
	/app/promql/query_logger.go:121 +0x3cd
main.main()
	/app/cmd/prometheus/main.go:605 +0x6773
```

PR #312, opened 2023-07-14
fix: test case multi-namespace_support

lihongyan1:
Fix issue #311 

Prometheus pod need more resources and has complicated configuration, replace it with prometheus example app pod

### DIFF
--- a/test/e2e/monitoring_stack_controller_test.go
+++ b/test/e2e/monitoring_stack_controller_test.go
@@ -22,6 +22,8 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"k8s.io/utils/pointer"
+
 	"github.com/google/go-cmp/cmp"
 
 	stack "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
@@ -861,7 +863,7 @@ func namespaceSelectorTest(t *testing.T) {
 
 	promClient := framework.NewPrometheusClient("http://localhost:9090")
 	if pollErr := wait.Poll(5*time.Second, 5*time.Minute, func() (bool, error) {
-		query := `prometheus_build_info{namespace=~"test-ns-.*"}`
+		query := `version{pod="prometheus-example-app",namespace=~"test-ns-.*"}`
 		result, err := promClient.Query(query)
 		if err != nil {
 			return false, nil
@@ -889,7 +891,7 @@ func deployDemoApp(t *testing.T, nsName string, nsLabels, resourceLabels map[str
 	}
 
 	// deploy a pod, service, service-monitor into that namespace
-	prom := newPrometheusPod(t, "prometheus", ns.Name)
+	prom := newPrometheusExampleAppPod(t, "prometheus-example-app", ns.Name)
 	if err := f.K8sClient.Create(context.Background(), prom); err != nil {
 		return fmt.Errorf("failed to create demo app %s/%s: %w", nsName, prom.Name, err)
 	}
@@ -902,15 +904,15 @@ func deployDemoApp(t *testing.T, nsName string, nsLabels, resourceLabels map[str
 	// these are prometheus ports
 	svc.Spec.Ports = []corev1.ServicePort{{
 		Name:       "metrics",
-		Port:       9090,
-		TargetPort: intstr.FromInt(9090),
+		Port:       8080,
+		TargetPort: intstr.FromInt(8080),
 	}}
 
 	if err := f.K8sClient.Create(context.Background(), svc); err != nil {
 		return fmt.Errorf("failed to create service for demo app %s/%s: %w", nsName, svc.Name, err)
 	}
 
-	svcMon := newServiceMonitor(t, ns.Name, "prometheus", resourceLabels, svcLabels, "metrics")
+	svcMon := newServiceMonitor(t, ns.Name, "prometheus-example-app", resourceLabels, svcLabels, "metrics")
 	if err := f.K8sClient.Create(context.Background(), svcMon); err != nil {
 		return fmt.Errorf("failed to create servicemonitor for demo service %s/%s: %w", nsName, svcMon.Name, err)
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/monitoring_stack_controller_test.go
```diff
@@ -973,6 +973,7 @@ func newService(t *testing.T, name, namespace string, labels, selector map[strin
 }
 
 func newPrometheusPod(t *testing.T, name, ns string) *corev1.Pod {
```
sthaha, line 975: could you please rename this to `newPrometheusExamplePod` now that the definition has changed?

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -983,16 +984,27 @@ func newPrometheusPod(t *testing.T, name, ns string) *corev1.Pod {
...
-				"app.kubernetes.io/version": "2.39.1",
+				"app.kubernetes.io/version": "multiarch-v0.1.0",
 			},
 		},
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{{
-				Name:  "prometheus",
-				Image: "quay.io/prometheus/prometheus:v2.39.1",
+				Name:  "prometheus-example-app",
+				Image: "quay.io/openshifttest/prometheus-example-app@sha256:382dc349f82d730b834515e402b48a9c7e2965d0efbc42388bd254f424f6193e",
```
sthaha, line 993: I wonder if we should instead rely on what upstream Prometheus Operator is using ?

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -983,16 +984,27 @@ func newPrometheusPod(t *testing.T, name, ns string) *corev1.Pod {
...
+				SecurityContext: &corev1.SecurityContext{
+					AllowPrivilegeEscalation: &bF,
```
sthaha, line 995: nit: consider `pointer.Bool` - https://pkg.go.dev/k8s.io/utils/pointer#Bool

sthaha, line 995: nit: consider `pointer.Bool` - https://pkg.go.dev/k8s.io/utils/pointer#Bool

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -861,7 +861,7 @@ func namespaceSelectorTest(t *testing.T) {
 
 	promClient := framework.NewPrometheusClient("http://localhost:9090")
 	if pollErr := wait.Poll(5*time.Second, 5*time.Minute, func() (bool, error) {
-		query := `prometheus_build_info{namespace=~"test-ns-.*"}`
+		query := `version{namespace=~"test-ns-.*"}`
```
sthaha, line 864: is there a different metric that we can reply on that is specific to `prometheus_example` ?

sthaha (changes requested):
Thank you for the patch @lihongyan1. The approach looks good to me. Some comments that needs addressing.

test/e2e/monitoring_stack_controller_test.go
lihongyan1: I would prefer to use this one, for I run the upstream test cases on OCP with all kinds of profiles share between QE team during ObO release, the image will be mirrored by default and the test cases can get the image and  run on disconnected or private OCP clusters, other wise, the case will failed on private and disconnected OCP cluster.

test/e2e/monitoring_stack_controller_test.go
lihongyan1: Yes, there are the following metrics, for detail you can refer [web site](https://github.com/brancz/prometheus-example-app)
```
version - of type gauge - containing the app version - as a constant metric value 1 and label version, representing this app version
http_requests_total - of type counter - representing the total numbere of incoming HTTP requests
http_request_duration_seconds - of type histogram, representing duration of all HTTP requests
http_request_duration_seconds_count- total count of all incoming HTTP requeests
http_request_duration_seconds_sum - total duration in seconds of all incoming HTTP requests
http_request_duration_seconds_bucket
```

lihongyan1: I know what you mean, I changed the metric as `version{pod="prometheus-example-app",namespace=~"test-ns-.*"}`, thanks for the comments

test/e2e/monitoring_stack_controller_test.go
jan--f: @lihongyan1 would you mind leaving a comment that describes this and a link to the upstream that is mirrored? Just for future us when someone asks why we don't use the upstream version :)

lihongyan1: Added comments

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -983,16 +985,29 @@ func newPrometheusPod(t *testing.T, name, ns string) *corev1.Pod {
...
+				"app.kubernetes.io/version": "multiarch-v0.4.1",
 			},
 		},
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{{
-				Name:  "prometheus",
-				Image: "quay.io/prometheus/prometheus:v2.39.1",
+				Name: "prometheus-example-app",
+				//To suppose multi-arch, image built directly from https://github.com/brancz/prometheus-example-app
+				//To support OCP disconnected cluster and private cluster, put image under quay.io/openshifttest to be mirrored by default
```
sthaha, line 995: How about .. 

```suggestion
				// The following image is a  rebuild  of  https://github.com/brancz/prometheus-example-app
				// with multi-arch enabled see: https://quay.io/repository/openshifttest/prometheus-example-app/manifest/sha256:382dc349f82d730b834515e402b48a9c7e2965d0efbc42388bd254f424f6193e 
				// In addition to the above, this image is also available in OCP disconnected cluster,
				// thus allowing the tests to be run in that environment.
```

sthaha (approved):
lgtm

PR #313, opened 2023-07-18
fix: enforce scheduling registry pod onto infra nodes

anispate:
This commit specifies .spec.grpcPodConfig.nodeSelector and .spec.grpdPodConfig.tolerations to schedule this operator's registry pod to infra nodes when possible.

[OSD-6629](https://issues.redhat.com//browse/OSD-6629)

### DIFF
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -55,6 +55,12 @@ objects:
             sourceType: grpc
             grpcPodConfig:
               securityContextConfig: restricted
+              nodeSelector:
+                node-role.kubernetes.io: infra
+              tolerations:
+              - effect: NoSchedule
+                key: node-role.kubernetes.io/infra
+                operator: Exists
         - apiVersion: operators.coreos.com/v1alpha2
           kind: OperatorGroup
           metadata:


hack/olm/catalog-src.yaml
```diff
@@ -16,6 +16,12 @@ spec:
   sourceType: grpc
   grpcPodConfig:
     securityContextConfig: restricted
+    nodeSelector:
+      node-role.kubernetes.io: infra
```
sthaha, line 20: @anispate , the PR says `prefer` instead of enforce. Is that really the case? I haven't tested this but my hunch is that this enforces Catalog pods to be scheduled only on infra nodes, doesn't it?

anispate: It was already tested before by us for other operators.

sthaha: If it enforces, like mentioned in the slack thread, we wouldn't want to enforce this on all users of the catalog. 
cc: @jan--f

mjlshen: Agreed, this configuration enforces that it can only be scheduled onto infra nodes. We likely need a separate CatalogSource for OSD/ROSA compared to the generic one

cblecker (approved):
This only updates the SelectorSyncSet used for managed deployments

deploy/olm/syncselector-template.yaml
sthaha, line 1: @jan--f are we okay with this change? I feel it is fine since it only modifies the syncselector-template.

jan--f (changes requested):
Sorry for blocking this, I think we should have an ack from the SD org.
For the deployment of the operator we were specifically asked to prefer (not require) infra nodes. Like in https://github.com/rhobs/observability-operator/pull/313.

Any reason we can't do the same for the CatalogSource?

jan--f:
2023-07-25-155005_852x72

The commit needs this name change :)

apahim:
The reason SD previously reccomended no "prefer" instead of "enfonce" was because we observed some cases where infra nodes didn't have capacity or were not available to take the workloads, but those were isolated cases.

The current understanding is that we are fine with enforcing the scheduling of infra bits to infra nodes, as SREP will receive an alert saying "InfraNodesNeedResizingSRE" and resize the nodes when needed.

This PR looks good to me.

jan--f (approved):
Thanks @apahim!
This lgtm.

sthaha:
@jan--f before merging, should the `prefer` in commit message be changed to `enforce` ?

PR #321, opened 2023-08-02
fix: add node tolerations to deployments

simonpasquier:
While the observability-operator, obo-prometheus-operator and obo-prometheus-operator-admission-webhook deployments are configured to be scheduled preferably on infra nodes, they lacked the required toleration that would enable them being scheduled properly.

This change fixes the issue by adding the expected toleration.

https://issues.redhat.com/browse/MON-3299


mjlshen (commented):
/lgtm

sthaha (approved):
/lgtm

PR #335, opened 2023-08-14
fix: use framework default timeout in ns tests

danielmellado:
This PR makes it so DefaultTimeout is used accordingly among the
different tests and assertions.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>


jan--f (approved):
Lgtm, though a little surprised we used such a long timeout before. If it works it works.

Will give other time to check it out and merge it tomorrow.

simonpasquier (approved):

PR #336, opened 2023-08-14
fix: update github workflow to use node>=16

danielmellado:
GitHub workflow is giving off a warning about the node version used for
some of the actions (engineerd/setup-kind) being deprecated. This commit
changes the plugin to helm/kind-action to avoid that.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/.github/e2e-tests-olm/action.yaml
+++ b/.github/e2e-tests-olm/action.yaml
@@ -25,10 +25,10 @@ runs:
     - uses: azure/setup-kubectl@v3
 
     - name: Start Kind
-      uses: engineerd/setup-kind@v0.5.0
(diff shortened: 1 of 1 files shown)


simonpasquier:
Doing the same in prometheus-operator: https://github.com/prometheus-operator/prometheus-operator/pull/5817

danielmellado:
> maybe we can switch to https://github.com/helm/kind-action which seems to see more updates?

Yes, I also submitted a fix to engineerd/setup-kind (https://github.com/engineerd/setup-kind/pull/247)  but doesn't seem to have any traction. I'll modify and update this PR to the helm action, thanks!

simonpasquier (approved):

PR #350, opened 2023-09-01
ci: use go version in go.mod file

jan--f:
And drop use of .github/env.


danielmellado (approved):
Great! Seems pretty straightforward and I also prefer to use go.mod directly now that it's available!

sthaha (approved):
/lgtm

PR #354, opened 2023-09-11
MON-2708: remove ServiceMonitor observability-operator from bundle

raptorsun:


simonpasquier:
this fails the commit message linter. See https://github.com/rhobs/observability-operator/blob/main/docs/developer.md#commit-message-convention

PR #361, opened 2023-09-18
chore: update operator-sdk v1.31.0

marioferh:
Operator-sdk needs this update: https://github.com/operator-framework/operator-sdk/pull/6408 in order to fix this: 
https://github.com/rhobs/observability-operator/blob/b6b11bb4f5e5a53041e555169abd71d7c1543ec4/deploy/dependencies/kustomization.yaml#L6-L21


jan--f (dismissed):
The commit linter actually looks at the commit message, not the PR message. I'd recommend squashin bot into one commit and adding `chore: ` to the title.

The generate job looks like it needs a change after updating `operator-sdk`. The `created-at` annotation makes this test fail, since it gets regenerated. Maybe we can prevent operator-sdk to add that?

danielmellado:
Same comment as @jan--f , please do a squash of both commits and fix violations. I'd recommend triggering the linter locally before pushing, thanks!

danielmellado (changes requested):
Small nit already explained

danielmellado:
> The commit linter actually looks at the commit message, not the PR message. I'd recommend squashin bot into one commit and adding `chore: ` to the title.
> 
> The generate job looks like it needs a change after updating `operator-sdk`. The `created-at` annotation makes this test fail, since it gets regenerated. Maybe we can prevent operator-sdk to add that?

https://github.com/operator-framework/operator-sdk/issues/6285 Seems related to this, we'll probably need to actively ignore that.

jan--f:
https://github.com/operator-framework/operator-sdk/issues/6285#issuecomment-1532150678 might be solution.

marioferh:
> [operator-framework/operator-sdk#6285 (comment)](https://github.com/operator-framework/operator-sdk/issues/6285#issuecomment-1532150678) might be solution.

thanks, done

jan--f:
/lgtm
Thanks!

danielmellado (approved):
/lgtm

PR #366, opened 2023-09-28
fix: add description field to CSV base

marioferh:
Add description field to CSV base of every owned CRD to generate correct description when run bundle generate.

### DIFF
--- a/bundle/manifests/observability-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/observability-operator.clusterserviceversion.yaml
@@ -42,7 +42,7 @@ metadata:
     categories: Monitoring
     certified: "false"
     containerImage: observability-operator:0.0.25
-    createdAt: "2023-09-28T13:10:33Z"
+    createdAt: "2023-10-04T08:52:37Z"
     description: A Go based Kubernetes operator to setup and manage highly available
       Monitoring Stack using Prometheus, Alertmanager and Thanos Querier.
     operators.operatorframework.io/builder: operator-sdk-v1.31.0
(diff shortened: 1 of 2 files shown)


bundle/manifests/observability-operator.clusterserviceversion.yaml
```diff
@@ -59,40 +59,68 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
-    - kind: AlertmanagerConfig
+    - description: AlertmanagerConfig defines a namespaced AlertmanagerConfig to be
```
machine424, line 62: Did you take a look at the UI rendering? I think it should be shorter to fit into the box, like in Gabriel's example

machine424:
The descriptions are not truncated and the boxes don't look text-heavy.

/lgtm

bundle/manifests/observability-operator.clusterserviceversion.yaml
marioferh: It is ok, thanks!

sthaha: image

sthaha: You can use  https://operatorhub.io/preview to preview the bundle/csv.yaml 
I think we can improve the  `AlertmanagerConfig` a bit ..  (ask chatgpt :D )

sthaha: BTW, are we now supporting all these API?

marioferh: In a cluster this is the result:
image

marioferh: `"AlertmanagerConfig is used to define configurations for Prometheus Alertmanager, specifying how alerts are handled and routed for notifications."`

Something like this?

sthaha: @simonpasquier would be the best judge of this :) .. BTW, for the record, this was only a suggestion and shouldn't block the PR from merging.

simonpasquier: small suggestion compared to @marioferh 

> ... specifying how alerts are grouped and notified to external systems.

marioferh: @simonpasquier https://github.com/prometheus-operator/prometheus-operator/pull/5955

jan--f:
Thanks! I guess there is nothing but keeping these up-to-date manually is there?

PR #370, opened 2023-10-12
build: tidy up remaining references to env andd go_version

jan--f:
In ab90d1ea181c356504c8d96f7a500e589f6d94e8 we moved to relying on the go version in go.mod. I forgot to fix a couple of spots.


simonpasquier (approved):
:+1:

PR #386, opened 2023-11-09
MON-3468: Add cli args to specify image pullspecs and versions for operants

### LINKED ISSUE #250: Avoid hardcoded versions for Thanos components
The observability operator hard-codes the versions used for the Thanos sidecar and querier. The current value is v0.24.0 which is rather old.

For the Thanos sidecar, we could leave the fieds unset and the Prometheus operator would use whatever is the latest Thanos version when the operator was released. For the querier, we could rely on [pkg/operator/defaults.go](https://github.com/prometheus-operator/prometheus-operator/blob/e4e27052f57040f073c6c1e4aedaecaaec77d170/pkg/operator/defaults.go#L28) from `github.com/rhobs/obo-prometheus-operator`.

https://github.com/rhobs/observability-operator/blob/c2ecb858856e68d1cf2f376b0b875410ef74d8ed/pkg/controllers/monitoring/monitoring-stack/components.go#L179-L182

https://github.com/rhobs/observability-operator/blob/c2ecb858856e68d1cf2f376b0b875410ef74d8ed/pkg/controllers/monitoring/thanos-querier/components.go#L71

jan--f:
Fixes #250


pkg/controllers/monitoring/thanos-querier/controller.go
```diff
@@ -41,6 +42,13 @@ type resourceManager struct {
 	client.Client
 	scheme *runtime.Scheme
 	logger logr.Logger
+	thanos map[string]string
+}
+
+// Options allows for controller options to be set
```
simonpasquier, line 48: (suggestion) should we have a config struct instantiated by main.go rather than map[string]string variables which are less explicit IMHO?

````go
type OperandConfiguration struct {
    Image string
    Version string
}

type OperatorConfiguration struct {
    Prometheus OperandConfiguration
    Alertmanager OperandConfiguration
    Thanos OperandConfiguration
}
````

**Comment by jan--f** [comment]
Updated with `prometheus/prometheus@v0.48.0-rc.2`.

File: pkg/controllers/monitoring/thanos-querier/controller.go (hunk 1 above)
Comment (jan--f) [reply pkg/controllers/monitoring/thanos-querier/controller.go:48]: Yeah I agree, that would improve readability. I did play around with that too and I didn't go that way because I wasn't sure how to get the key-value style argument parsing paired up with a struct without going through some function involving `reflect`. 

I missed that I didn't add the changes for the `versions` cli arg, added now.

Do you know how I can parse two repeatable cli args into a single struct? Otherwise I wonder if its worth going through the trouble to convert `map[string]string` to a struct.

File: cmd/operator/main.go [hunk 2]
```diff
@@ -39,6 +82,10 @@ func main() {
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
+	images := kvarg{}
+	flag.Var(&images, "images", "Images to use for containers managed by the observability-operator.")
+	versions := kvarg{}
+	flag.Var(&versions, "versions", "Version of containers managed by the observability-operator.")
```
simonpasquier, line 88: Do we need to define the operand versions or should we rely on the hardcoded default versions from github.com/rhobs/obo-prometheus-operator?

jan--f: I was going to argue for the p-o default versions. In the Thanos Querier controller I already set that explicitly https://github.com/rhobs/observability-operator/pull/386/files#diff-62df723ef5f27edbbfaa7feeb8a0612efaa9ef8d91828004287a17f77f0c48aeR74. In `Prometheus` objects the version is not set if empty to let p-o do its defaulting.

pkg/controllers/monitoring/thanos-querier/controller.go
danielmellado: Do we expect this to grow at some point? If so, +1 on the struct but just for image/version I think it may not be worth the change.

test/e2e/framework/assertions.go
```diff
@@ -252,12 +223,10 @@ func (f *Framework) GetOperatorMetrics(t *testing.T) []byte {
 // AssertNoReconcileErrors asserts that there are no reconcilation errors
 func (f *Framework) AssertNoReconcileErrors(t *testing.T) {
 	metrics := f.GetOperatorMetrics(t)
-	assertPromQL(t, metrics,
-		`controller_runtime_reconcile_errors_total`,
+	assertSamples(t, metrics,
 		map[string]float64{
-			`{__name__="controller_runtime_reconcile_errors_total", controller="grafana-operator"}`: 0,
```
danielmellado, line 258: I assume we're removing this as a part of the change as grafana-operator is no longer around in p-o

cmd/operator/main.go
danielmellado: Defaulting would be latest or does p-o do any kind of logic already?

simonpasquier (commented):
FYI I've stumbled across https://pkg.go.dev/k8s.io/component-base@v0.28.3/cli/flag#MapStringString which can translate `key=val` CLI args into map[string]string.

jan--f:
> FYI I've stumbled across https://pkg.go.dev/k8s.io/component-base@v0.28.3/cli/flag#MapStringString which can translate key=val CLI args into map[string]string.

Nice, thank you! Added.

cmd/operator/main.go
```diff
@@ -32,13 +33,17 @@ func main() {
...
+		images          k8sflag.MapStringString
+		versions        k8sflag.MapStringString
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
+	flag.Var(&images, "images", "Images to use for containers managed by the observability-operator.")
```
sthaha, line 45: I think it may help in the longer run to make use of `releatedImages` in cluster-service-version. see: https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs 

See also: https://github.com/sustainable-computing-io/kepler-operator/pull/242/files#diff-3492afe4e3cd6b74bb7970b8a215ba867f91984594da0cd558f5fe8dd03bfc6d where kepler-operator uses relatedImages to initialise the `kepler` image to be used.

vkmc: Using relatedImages seem the preferred approach. What are the current limitations that prevent us from doing so instead?

cmd/operator/main.go
vkmc: IMHO it makes more sense for p-o and thanos do their defaults

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -22,7 +23,14 @@ const AdditionalScrapeConfigsSelfScrapeKey = "self-scrape-config"
 const PrometheusUserFSGroupID = int64(65534)
 const AlertmanagerUserFSGroupID = int64(65535)
 
-func stackComponentReconcilers(ms *stack.MonitoringStack, instanceSelectorKey string, instanceSelectorValue string) []reconciler.Reconciler {
+func stackComponentReconcilers(
+	ms *stack.MonitoringStack,
+	instanceSelectorKey string,
+	instanceSelectorValue string,
+	thanos tqctrl.ThanosConfiguration,
```
vkmc, line 30: NIT thanosCfg, prometheusCfg and alertmanagerCfg might be more descriptive

vkmc (commented):
Looks good, just some minor comments, thanks Jan!

cmd/operator/main.go
jan--f: We will use `relatedImages`. This PR here is just the plumbing for that. Getting an env var vs. explicit passing via cli flags is imho an implementation detail.
We will still use the `relatedImages` CSV field. The main reason this isn't part of the PR for now is that our internal build tooling atomates setting `relatedImages` but afaiu only if its not already present in the CSV. We can always add that to the CSV generation here at a later stage though.

pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	stack "github.com/rhobs/observability-operator/pkg/apis/monitoring/v1alpha1"
+	tqctrl "github.com/rhobs/observability-operator/pkg/controllers/monitoring/thanos-querier"
```
simonpasquier, line 40: Importing `pkg/controllers/monitoring/thanos-querier` feels a bit weird to me, especially since Thanos Q probably doesn't need the version information. I'd decouple the 2 packages even if it means that code duplication.

pkg/controllers/monitoring/thanos-querier/controller.go
```diff
@@ -41,6 +42,30 @@ type resourceManager struct {
...
+	Image   string
+	Version string
+}
+
+func (t ThanosConfiguration) GetImageRef() string {
+	image := "quay.io/thanos/thanos"
+	if t.Image != "" {
+		image = t.Image
+	}
+	if t.Version == "" {
```
simonpasquier, line 58: I'd expect the --image arg to be a fully specified image name?

cmd/operator/main.go
simonpasquier, line 45: ?
```suggestion
	flag.Var(&images, "images", "Images to use for containers managed by the cluster-observability-operator.")
```

jan--f: hmm good point, but I'd say maybe we should go with `managed by the  operator`? This way we don't get weird doc strings that add confusion around upstream and downstream.
And it should be pretty clear which operator is meant.

pkg/controllers/monitoring/thanos-querier/controller.go
jan--f: Right now the operator accepts `--images` and `--versions`. The motivation is when a user wants to rely on the default location but with a specific version.
Though I guess in relation to OLM's `relatedImages` this is not a use case. I'll change this accordingly

pkg/controllers/monitoring/monitoring-stack/controller.go
jan--f: I used the thanos-querier type already in `pkg/operator/operator.go`. I'll add a commit where `monitoringstack` has its own type.

cmd/operator/main.go
simonpasquier: :+1:

pkg/controllers/monitoring/monitoring-stack/components.go
jan--f: :+1: will address that after we agree on the general structure of this change.

pkg/controllers/monitoring/thanos-querier/controller.go
jan--f: :+1: I was actually taking inspiration from p-o initially :laughing:

cmd/operator/main.go
```diff
@@ -32,13 +33,15 @@ func main() {
...
 		healthProbeAddr string
+		images          k8sflag.MapStringString
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
+	flag.Var(&images, "images", "Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0")
```
simonpasquier, line 44: We'd need to enumerate which keys are supported (e.g. "prometheus", "alertmanager" and "thanos") and the operator should bail out if a key isn't in the list.

pkg/operator/operator.go
```diff
@@ -26,23 +26,51 @@ type Operator struct {
...
+type OperatorConfiguration struct {
+	MetricsAddr     string
+	HealthProbeAddr string
+	Prometheus      stackctrl.PrometheusConfiguration
+	Alertmanager    stackctrl.AlertmanagerConfiguration
+	ThanosSidecar   stackctrl.ThanosConfiguration
+	ThanosQuerier   tqctrl.ThanosConfiguration
+}
+
+func NewOperatorConfiguration(metricsAddr string, healthProbeAddr string, images map[string]string) OperatorConfiguration {
```
simonpasquier, line 43: (suggestion) once a constructor has more than 2/3 arguments, I like to use functional options. It would also avoid passing the map[string]string directly and it would be more explicit IMHO.

```suggestion
func WithPrometheusImage(img string) func(*OperatorConfiguration) {
...
}

func WithThanosImage(img string) func(*OperatorConfiguration) {
...
}

func NewOperatorConfiguration(options ...Option) OperatorConfiguration {
```

pkg/operator/operator.go
```diff
@@ -26,23 +26,51 @@ type Operator struct {
...
 	})
 	if err != nil {
 		return nil, fmt.Errorf("unable to create manager: %w", err)
 	}
 
-	if err := stackctrl.RegisterWithManager(mgr, stackctrl.Options{InstanceSelector: instanceSelector}); err != nil {
+	if err := stackctrl.RegisterWithManager(mgr, stackctrl.Options{
+		InstanceSelector: instanceSelector,
+		Prometheus:       cfg.Prometheus,
+		Alertmanager:     cfg.Alertmanager}); err != nil {
```
simonpasquier, line 69: missing ThanosSidecar?

pkg/operator/operator.go
```diff
@@ -26,23 +26,51 @@ type Operator struct {
 	manager manager.Manager
 }
 
-func New(metricsAddr, healthProbeAddr string) (*Operator, error) {
+type OperandConfiguration struct {
```
simonpasquier, line 29: is it used?

pkg/controllers/monitoring/thanos-querier/controller.go
```diff
@@ -41,6 +42,24 @@ type resourceManager struct {
...
 	logger logr.Logger
+	thanos ThanosConfiguration
+}
+
+type ThanosConfiguration struct {
+	Image string
+}
+
+func (t ThanosConfiguration) GetImageRef() string {
+	image := "quay.io/thanos/thanos" + obopo.DefaultThanosVersion
```
simonpasquier, line 53: I'd rather delegate the default value to the pkg/operator package.

pkg/operator/operator.go
jan--f: Nice idea!

pkg/operator/operator.go
jan--f: nope, I missed that.

jan--f:
I'll rebase this once we agree on the code since there is another go dependency update in the pipeline.

danielmellado:
hmmmm e2e tests are complaining about ```mv: cannot stat 'tmp/e2e': No such file or directory``` let's see

cmd/operator/main.go
```diff
@@ -18,15 +18,34 @@ package main
...
+	k8sflag "k8s.io/component-base/cli/flag"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+// validateImages merges the passed images with the defaults and checks if any
+// unknown image names are passed. If an unknown image is found, this raises an
+// error.
+func validateImages(images *k8sflag.MapStringString) (map[string]string, error) {
+	res := operator.DefaultImages
```
simonpasquier, line 36: (nit) I would move DefaultImages to the main package because there's no use of it in pkg/operator

cmd/operator/main.go
```diff
@@ -35,10 +54,13 @@ func main() {
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
+	m := make(map[string]string)
+	images := k8sflag.NewMapStringString(&m)
```
simonpasquier, line 58: (nit)
```suggestion
	images := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
```

pkg/operator/operator.go
```diff
@@ -21,28 +23,105 @@ const instanceSelector = "app.kubernetes.io/managed-by=observability-operator"
...
+	HealthProbeAddr string
+	Prometheus      stackctrl.PrometheusConfiguration
+	Alertmanager    stackctrl.AlertmanagerConfiguration
+	ThanosSidecar   stackctrl.ThanosConfiguration
+	ThanosQuerier   tqctrl.ThanosConfiguration
+}
+
+func WithPrometheusImage(image string) func(*OperatorConfiguration) {
+	return func(oc *OperatorConfiguration) {
+		oc.Prometheus = stackctrl.PrometheusConfiguration{Image: image}
```
simonpasquier, line 61: in case we extend the Prometheus config in the future
```suggestion
		oc.Prometheus.Image = image
```

cmd/operator/main.go
danielmellado: +1 to that, as it would be more clear, but totally a nit

pkg/operator/operator.go
```diff
@@ -21,28 +23,105 @@ const instanceSelector = "app.kubernetes.io/managed-by=observability-operator"
 
 const ObservabilityOperatorName = "observability-operator"
 
+// The default values we use. Prometheus and Alertmanager are handled by
+// prometheus-operator. For thanos we use the default version from
+// prometheus-operator.
+var DefaultImages = map[string]string{
+	"prometheus":   "",
+	"alertmanager": "",
+	"thanos":       "quay.io/thanos/thanos:" + obopo.DefaultThanosVersion,
```
danielmellado, line 32: nit: Could we make `quay.io/thanos/thanos`a var? I know we're not using it anymore around but still.

pkg/operator/operator.go
danielmellado: Should we go and adapt this same approach for the other components as well? Not sure how likable for us to extend their configuration, though

pkg/operator/operator.go
jan--f: How about we use `obopo.DefaultThanosImage` :D

cmd/operator/main.go
```diff
@@ -18,15 +18,56 @@ package main
...
+	}
+	return imgs
+}
+
+// validateImages merges the passed images with the defaults and checks if any
+// unknown image names are passed. If an unknown image is found, this raises an
+// error.
+func validateImages(images *k8sflag.MapStringString) (map[string]string, error) {
+	res := defaultImages
+	if !images.Empty() {
```
simonpasquier, line 59: (nit) fast path if no custom images are provided
```suggestion
	if images.Empty() {
		return res
	}
```

cmd/operator/main.go
```diff
@@ -18,15 +18,56 @@ package main
...
+		imgs := *images.Map
+		for k, v := range imgs {
+			if _, ok := res[k]; !ok {
+				return nil, fmt.Errorf(fmt.Sprintf("image %v is unknown", k))
```
simonpasquier, line 63:
```suggestion
				return nil, fmt.Errorf("image %v is unknown", k)
```

danielmellado (approved):
lgtm

PR #413, opened 2024-01-30
ppc64le/multi-arch support for observability-operator

jcho02:
Rebased from close [#308](https://github.com/rhobs/observability-operator/pull/308)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,11 +136,16 @@ build:
 
 .PHONY: operator-image
 operator-image: generate
-	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile --no-cache --platform linux/amd64 -t $(OPERATOR_IMG)-amd64 --build-arg TAGRTEARCH=amd64 .
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile --no-cache --platform linux/ppc64le -t $(OPERATOR_IMG)-ppc64le --build-arg TARGETARCH=ppc64le .
+	$(CONTAINER_RUNTIME) push ${OPERATOR_IMG}-amd64
+	$(CONTAINER_RUNTIME) push ${OPERATOR_IMG}-ppc64le
+	$(CONTAINER_RUNTIME) manifest create $(OPERATOR_IMG) 
+	$(CONTAINER_RUNTIME) manifest add $(OPERATOR_IMG) $(OPERATOR_IMG)-amd64 $(OPERATOR_IMG)-ppc64le
 
 .PHONY: operator-push
 operator-push:
-	$(CONTAINER_RUNTIME) push $(PUSH_OPTIONS) ${OPERATOR_IMG}
+	$(CONTAINER_RUNTIME) manifest push --all $(PUSH_OPTIONS) ${OPERATOR_IMG}
 
 .PHONY: osd-e2e-test-image
 osd-e2e-test-image: tools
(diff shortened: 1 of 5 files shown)


jcho02:
@danielmellado

Makefile
```diff
@@ -221,13 +227,13 @@ CATALOG_IMG_LATEST ?= $(IMAGE_BASE)-catalog:latest
 # operator package manager tool, 'opm'.
 .PHONY: catalog-image
 catalog-image: $(OPM)
-	$(OPM) render $(BUNDLE_IMG) \
-		--output=yaml  >> olm/observability-operator-index/index.yaml
-	./olm/update-channels.sh $(CHANNELS) $(OPERATOR_BUNDLE)
-	$(OPM) validate ./olm/observability-operator-index
+	$(OPM) render quay.io/jcho0/observability-operator-catalog:0.0.23 \
```
jan--f, line 230: This seem like a debug change?

Makefile.tools
```diff
@@ -14,7 +14,7 @@ KUSTOMIZE=$(TOOLS_DIR)/kustomize
 KUSTOMIZE_VERSION= v4.5.2
 
 OPERATOR_SDK = $(TOOLS_DIR)/operator-sdk
-OPERATOR_SDK_VERSION = v1.31.0
+OPERATOR_SDK_VERSION = v1.24.0
```
jan--f, line 17: Does this need a version downgrade?

Makefile
```diff
@@ -195,19 +200,20 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
...
-	git diff --quiet -I'^    createdAt: ' bundle && git checkout bundle || true
+	$(OPERATOR_SDK) bundle validate ./bundle
 
 .PHONY: bundle-image
 bundle-image: bundle ## Build the bundle image.
-	$(CONTAINER_RUNTIME) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	$(CONTAINER_RUNTIME) build --no-cache -f bundle.Dockerfile --platform linux/amd64 -t $(BUNDLE_IMG)-amd64 .
+	$(CONTAINER_RUNTIME) build --no-cache -f bundle.Dockerfile --platform linux/ppc64le -t $(BUNDLE_IMG)-ppc64le .
+	$(CONTAINER_RUNTIME) push $(BUNDLE_IMG)-amd64
+	$(CONTAINER_RUNTIME) push $(BUNDLE_IMG)-ppc64le
```
jan--f, line 210: I'd prefer the image push stays in the push target. Is this needed to build the manifest?

Makefile
```diff
@@ -136,11 +136,16 @@ build:
 
 .PHONY: operator-image
 operator-image: generate
-	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile --no-cache --platform linux/amd64 -t $(OPERATOR_IMG)-amd64 --build-arg TAGRTEARCH=amd64 .
+	$(CONTAINER_RUNTIME) build -f build/Dockerfile --no-cache --platform linux/ppc64le -t $(OPERATOR_IMG)-ppc64le --build-arg TARGETARCH=ppc64le .
+	$(CONTAINER_RUNTIME) push ${OPERATOR_IMG}-amd64
+	$(CONTAINER_RUNTIME) push ${OPERATOR_IMG}-ppc64le
```
jan--f, line 142: Same question here as for the bundle image. Do we have to push images in the build step? If not strictly necessary I'd prefer to keep build and push separate.

Makefile.tools
```diff
@@ -194,16 +197,40 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
...
+	}
+
 .PHONY: jsonnet-tools
 jsonnet-tools: jsonnet jsonnetfmt jsonnet-lint jb gojsontoyaml
 
+.PHONY: jq
+jq: 
+	@{ \
+		set -ex ;\
+		sudo apt-get install jq ;\
```
jan--f, line 224: We keep the tool installs distro agnostic. and apt install won't work on my machine. Let's either install this like the other tools or add a gate that checks for presence and errors out if it's not in the PATH

hack/build.sh
jan--f, line 1: Is this actually used anywhere? If not the `skopeo` dependency can also be dropped iiuc.

Makefile
jcho02: Yes if we're using Podman, we can't build and push in one command for Podman multi-arch builds like Docker

hack/build.sh
jcho02: Its a bash script if users want to build all three operator, bundle, and catalog images in one command. I used it for dev purposes if you want to keep it, but its not necessary.

jan--f:
COO, the downstream version of this available in the Red Hat OLM catalog now supports multiple architectures. Since this is a bit outdated I'll close this. We can pick this back up if needed.

Issue #222, opened 2022-11-16
make release target is failing


tremes:
The error is:
```
bash: line 1: olm/observability-operator-index/index.yaml: No such file or directory
make: *** [Makefile:187: catalog-image] Error 1
```
There's no `olm` directory and there's probably nothing to create this directory. The [update-channels.sh](https://github.com/rhobs/observability-operator/blob/main/Makefile#L189) is also not available.

sthaha:
@tremes , after ObO moved to File Based Catalogs, the catalog is maintained in `olm-catalog` branch. So if you want to make a release (which shouldn't be really necessary for dev builds), the olm-catalog branch needs to be merged into your local branch. Check out  https://github.com/rhobs/observability-operator/blob/main/.github/olm-publish/action.yaml which does this in CI.

tremes:
@sthaha OK thanks.

sthaha:
@tremes Does this PR https://github.com/rhobs/observability-operator/pull/224 give you sufficient details? If not please feel free to comment.

sthaha:
@tremes perhaps we should remove the `release` make target. 
FYI: the recommended method to running bundle is now documented in the dev docs - see: https://github.com/rhobs/observability-operator/blob/main/docs/developer.md#run-operator-bundle---000-dev

PR #431, opened 2024-03-12
ci:publish bundle image with latest tag

lihongyan1:
Catalog image didn't build for a long time for issue https://github.com/rhobs/observability-operator/issues/222, it suggest to use bundle image to install operator, so publish bundle iamge with latest tag, so that refer the latest image easily during test


jan--f:
#432 should fix the issue.

lihongyan1:
> #432 should fix the issue.

#432 can't fix the issue, it doesn't make sense to test a release image in olm-catalog branch. #222 is not fixed and for every PR merged to main still has pipeline error, such as the following action, failed to publish olm development
https://github.com/rhobs/observability-operator/actions/runs/8254817309 
https://github.com/rhobs/observability-operator/actions/runs/8254817309/job/22580255690

lihongyan1:
https://github.com/rhobs/observability-operator/pull/432 should fix the issue. So I closed the pr

PR #432, opened 2024-03-12
fix: add channel marker back into index file

jan--f:
These were removed in 1875b9e99d14b7a84b59d0ab23ae939da27e6061 and are needed for release automation via olm/update-channels.sh.

Fixes: https://issues.redhat.com/browse/COO-17

### DIFF
--- a/.github/tools
+++ b/.github/tools
@@ -13,4 +13,4 @@ jsonnetfmt v0.17.0
 jsonnet-lint v0.17.0
 jb v0.4.0
 gojsontoyaml 0.0.1
-shellcheck 0.9.0
+shellcheck 0.10.0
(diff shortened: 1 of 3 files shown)


jan--f:
The generate check is expected to fail since this merge is not into `main`.

.github/tools
```diff
@@ -13,4 +13,4 @@ jsonnetfmt v0.17.0
 jsonnet-lint v0.17.0
 jb v0.4.0
 gojsontoyaml 0.0.1
-shellcheck 0.9.0
+shellcheck 0.10.0
```
danielmellado, line 16: Do we need to update shellcheck for the marker to work? Not that it would do any harm but just curious about it.

jan--f: No that is unrelated. However we install shellcheck@latest and they updated recently. So tool install fails with us still referencing 0.9.0.
Fwiw I'll push the same change to main.

jan--f: @danielmellado anything holding this up?

danielmellado: /lgtm

danielmellado: all good, thanks for the explanation!

PR #434, opened 2024-03-12
feat: Add ObservabilityUI plugins API

jgbernalp:
This PR adds the Observability UI CRDs.

https://issues.redhat.com/browse/OU-357

Every ObservabilityUIPlugin, creates the following resources:

- Deployment to host a lightweight backend for the plugin assets
- Service to expose the deployment to the console
- Service account
- ConsolePlugin CR so the console knows in which service the plugin lives
- Optional: ClusterRole and ClusterRoleBinding for some plugins so their backend can watch configmaps in a namespace

And patches the console to add the plugin so is enabled in the web console


jan--f:
Generally this looks good, I might have requests on some details later.
One question we should discuss: currently obo is Openshift agnostic and can be installed on a vanilla k8s. We make use of that in development during e2e testing with a kind cluster. If at all possible I'd like to preserve that feature. Iiuc the dependencies with this PR are the `ConsolePlugin` kind and serving cert annotations in the service. 
Do you think there is a way to degrade gracefully if we're not on OCP, i.e. the `ConsolePlugin` is not present?
Does the `Service` require and https endpoint or could this be implemented differently? So far we leave any tls setups to the user via service-side-apply, though we certainly have considered encrypted by default before.

jgbernalp:
> Generally this looks good, I might have requests on some details later. One question we should discuss: currently obo is Openshift agnostic and can be installed on a vanilla k8s. We make use of that in development during e2e testing with a kind cluster. If at all possible I'd like to preserve that feature. Iiuc the dependencies with this PR are the `ConsolePlugin` kind and serving cert annotations in the service. Do you think there is a way to degrade gracefully if we're not on OCP, i.e. the `ConsolePlugin` is not present? Does the `Service` require and https endpoint or could this be implemented differently? So far we leave any tls setups to the user via service-side-apply, though we certainly have considered encrypted by default before.

Thanks for the initial check. I agree, we should verify whether the cluster supports the ConsolePlugin and, if it's not present, skip the reconciliation process. Imo since the service is relevant only within an OpenShift cluster where plugins are available, opting to skip reconciliation seems to be an adequate solution also for the serving cert annotation.

danielmellado:
I was also about to suggest to drop openshift-specific names like groupAPI but it just makes sense to completely fall back if on non-openshift cluster. +1!

periklis:
> Generally this looks good, I might have requests on some details later. One question we should discuss: currently obo is Openshift agnostic and can be installed on a vanilla k8s. We make use of that in development during e2e testing with a kind cluster. If at all possible I'd like to preserve that feature. Iiuc the dependencies with this PR are the `ConsolePlugin` kind and serving cert annotations in the service. Do you think there is a way to degrade gracefully if we're not on OCP, i.e. the `ConsolePlugin` is not present? Does the `Service` require and https endpoint or could this be implemented differently? So far we leave any tls setups to the user via service-side-apply, though we certainly have considered encrypted by default before.

Sharing my experience on how we isolate OpenShift features in Loki Operator the last two years:
1. First things first, having an operator registering CRDs in the client side scheme even if the CRDs do not existing on the APIserver side is not a harmful thing. It get's only harmful if you start a controller starting to sync/watch for such resources.
2. What worked good for us are two things. First we publish three almost bundles ([community](https://github.com/grafana/loki/tree/main/operator/bundle/community), [community-openshift](https://github.com/grafana/loki/tree/main/operator/bundle/community-openshift), [openshift](https://github.com/grafana/loki/tree/main/operator/bundle/openshift)). The bundles are identical except for:
    - **Container images**: community and community-openshift reference images on dockerhub based on alpine. Openshift is a bundle only populated at and by Red Hat using RHEL-based images and other productization bits (annotations, CVP, etc.)
    - **Feature gates**: This used to be the controller-runtime project config that migrated into the [operator code base](https://github.com/grafana/loki/blob/main/operator/internal/config/options.go#L20) because it is migrated (i.e. [community](https://github.com/grafana/loki/blob/main/operator/bundle/community/manifests/loki-operator-manager-config_v1_configmap.yaml#L15), [community-openshift](https://github.com/grafana/loki/blob/main/operator/bundle/community-openshift/manifests/loki-operator-manager-config_v1_configmap.yaml#L15), [openshift](https://github.com/grafana/loki/blob/main/operator/bundle/openshift/manifests/loki-operator-manager-config_v1_configmap.yaml#L17)). The feature gates include a separate OpenShift only [section](https://loki-operator.dev/docs/feature-gates.md/#config-loki-grafana-com-v1-OpenShiftFeatureGates) that handles features like service-ca, CCO, etc. Regardless these some of the generic feature gates are opened all together because we know we have support on OpenShift only OOTB (e.g. ServiceMonitors, TLS Endpoints in ServiceMonitors, Restricted Pod Security, Alerts)
2. Controlling the feature gates per bundle gives us the maximum control on how much we can automate on the target cluster. The Community OpenShift and the OpenShift bundles basically go to the maximum because we can safely assume that Prometheus, ServiceCA, etc. exist on the platform. On the other hand the Community bundle is the bare minimum that one can install from [here](https://operatorhub.io/operator/loki-operator) and the user can adapt the Feature Gates at will. The only exception to the rule is the dependency to Cert-Manager (and this is only a recommendation) for TLS certificates used for our Webhooks.
3. Based on feature gates we  manage individual runtime characterists:
    - **Starting controllers**: For example only on OpenShift the [dashboards-controller](https://github.com/grafana/loki/blob/main/operator/main.go#L112-L129)
    - **Resource Registration**: Based on the gates we register [watches/owns](https://github.com/grafana/loki/blob/main/operator/controllers/loki/lokistack_controller.go#L202-L240) handlers for each controller, e.g. we watch for the cluster Proxy (used for container proxy configuration) or APIServer (used for TLS Profiles) objects only in OpenShift.
    - **Extra Webhook Validators**: For example only on OpenShift we enable the [validations](https://github.com/grafana/loki/blob/main/operator/main.go#L148-L150) of alerting rules according to the official[ style guide](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#style-guide).
    - **Manage Resources**: On OpenShift we replace Ingress for [Route](https://github.com/grafana/loki/blob/main/operator/main.go#L89-L92) and add support for CCO's [CredentialRequest](https://github.com/grafana/loki/blob/main/operator/main.go#L89-L92) using in ARO and ROSA.
4. Anything we put in [manifests](https://github.com/grafana/loki/tree/main/operator/internal/manifests) is pure k8s from the beginning. This means we can test this on kind. Anything OpenShift specific is preserved in dedicated [subpackages](https://github.com/grafana/loki/tree/main/operator/internal/manifests/openshift) that is applied by patching the structs with [mergo](https://github.com/grafana/loki/blob/main/operator/internal/manifests/openshift/configure.go#L54-L90).

simonpasquier:
Reading quickly through the discussion, I think that the Loki approach is lean and flexible (thanks @periklis for the details!). FWIW he upstream Prometheus operator does something different: it introspects CRDs + RBAC permissions to adjust the behavior at runtime (e.g. conditional start of some controllers) but the situation isn't the same since we don't control the deployment model (can be custom installation, Helm, OLM, ...) and have to deal with users not updating everything when they upgrade to a new operator version.

One question would be regarding HyperShift: do you foresee (or have you experienced) a situation where the operator depends on a feature that exists in classic OCP but not in HyperShift?

periklis:
> One question would be regarding HyperShift: do you foresee (or have you experienced) a situation where the operator depends on a feature that exists in classic OCP but not in HyperShift?

Not yet! But from experience with ARO/ROSA I can foresee the following workflow, assuming Service Delivery is managing the Loki Operator installation for customers:
- Open the Loki Operator feature gates as a separate API (e.g. similar to openshift/api's `FeatureGates` or via this one) to manage on runtime
- Or provide a mechanism for HIVE/HyperShift/whatever to populate a separate ConfigMap for our feature gates.

In general our support matrix for feature gates settings in Loki Operator has been static and in turn weakly tested to users manipulating them before installing the operator. We suspect with quite a confidence that some combinations will break the operator entirely. Therefore we give them a static but tested configmap upfront in the bundle as per later being treated immutable by OLM.

pkg/apis/observability-ui/v1alpha1/types.go
```diff
@@ -0,0 +1,120 @@
...
+	ConditionTrue    ConditionStatus = "True"
+	ConditionFalse   ConditionStatus = "False"
+	ConditionUnknown ConditionStatus = "Unknown"
+
+	ReconciledCondition        ConditionType = "Reconciled"
+	AvailableCondition         ConditionType = "Available"
+	ResourceDiscoveryCondition ConditionType = "ResourceDiscovery"
+)
+
+type Condition struct {
```
tremes, line 74: I am wondering whether it would make sense to share the `Condition` type between the APIs (same for the https://github.com/rhobs/observability-operator/pull/417). Hopefully, this could allow us to "standardize" the way status is defined and calculated (for each stack/plugin)....are there any downsides (of having one `Condition` type)?

docs/user-guides/observability-ui-plugins.md
```diff
@@ -0,0 +1,23 @@
+# Observability UI Plugins
+
+Using the Observability UI, you can install and manage plugins that extend the observability functionality of the OpenShift web console. Plugins are installed and managed by the Observability Operator.
+
+## Plugins
+
+- [dashboards](#dashboards): Add enhanced dashboards to the OpenShift web console. This plugin allows you to add other prometheus datasources present in the cluster, apart from the in-cluster one, to the default dashboards.
```
tremes, line 7: > This plugin allows you to add other prometheus datasources present in the cluster, apart from the in-cluster one, to the default dashboards.

nit: Prometheus should be probably with capital P, but maybe it would be good to explicitly distinguish the groups - `moitoring.rhobs` vs `monitoring.coreos.com`

pkg/apis/observability-ui/v1alpha1/types.go
```diff
@@ -0,0 +1,120 @@
...
+const (
+	TypeDashboards UIPluginType = "dashboards"
+)
+
+// Observabilityuipluginpec is the specification for desired state of ObservabilityUIPlugin.
+type ObservabilityUIPluginSpec struct {
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=dashboards
+	Type UIPluginType `json:"type"`
```
tremes, line 45: I was experimenting with `ObservabilityUIPlugin` creation in an OpenShift cluster and even though the `type` is displayed as required (in UI), I can create a plugin with an empty value. 

Perhaps the `// +kubebuilder:validation:Enum=dashboards` should be applied to `UIPluginType` type (example [here](https://github.com/openshift/api/blob/master/operator/v1/types.go#L94))

pkg/controllers/observability-ui/observability-ui-plugin/compatibility_matrix.go
```diff
@@ -0,0 +1,52 @@
...
+	{
+		PluginType:        obsui.TypeDashboards,
+		MinClusterVersion: "v4.11",
+		MaxClusterVersion: "",
+		ImageKey:          "ui-dashboards",
+		Features:          []string{},
+	},
+}
+
+func getImageKeyForPluginType(pluginType obsui.UIPluginType, clusterVersion string) string {
```
simonpasquier, line 29: (nit) the function could return (string, error) with error != nil if no key is found.

pkg/controllers/observability-ui/observability-ui-plugin/plugin_info_builder.go
```diff
@@ -0,0 +1,105 @@
...
+	image := pluginConf.Images[imageKey]
+
+	if image == "" {
+		return nil, fmt.Errorf("no image provided for plugin type %s with key %s", plugin.Spec.Type, imageKey)
+	}
+
+	name := "observability-ui-" + plugin.Name
+
+	switch plugin.Spec.Type {
+	case "dashboards":
```
simonpasquier, line 39: (nit)
```suggestion
	case obsui.TypeDashboards:
```

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	err := k8client.Get(context.TODO(), name, current)
+
+	return err == nil || !metaerrors.IsNoMatchError(err)
+}
+
+func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := rm.logger.WithValues("plugin", req.NamespacedName)
+
+	if !consolePluginCapabilityEnabled(rm.k8sClient, req.NamespacedName) {
+		logger.Info("Cluster console plugin not supported. Skipping observability UI plugin reconciliation")
```
simonpasquier, line 119: IIUC we would enter this branch in one of the following cases:
* COO is deployed on vanilla Kubernetes and the openshift feature gate is enabled. But we should already have failed the reconciler init since it tries to get the cluster version?
* COO is deployed on OpenShift but it has no permissions to get the ConsolePlugin objects.
* COO is deployed on OpenShift but the ConsolePlugin CRD doesn't exist (is it even possible? old version of OCP?).

I wonder if the check shouldn't be done only once at startup? Is there a case where it would return false then true?

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	logger := rm.logger.WithValues("plugin", req.NamespacedName)
+
+	plugin := obsui.ObservabilityUIPlugin{}
+
+	if err := rm.k8sClient.Get(ctx, req.NamespacedName, &plugin); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.V(3).Info("stack could not be found; may be marked for deletion")
+			return nil, nil
+		}
+		logger.Error(err, "failed to get ObservabilityUIPlugin")
```
simonpasquier, line 213: (nit) I think that the error will be logged twice (also by the reconciler).

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+// +kubebuilder:rbac:groups=operator.openshift.io,resources=consoles,verbs=get;patch;list;watch
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=get;list;watch;create;update;delete;patch
+
+// RBAC for reading cluster version
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
+
+func RegisterWithManager(mgr ctrl.Manager, opts Options) error {
+	logger := ctrl.Log.WithName("observability-ui")
+
+	clusterVersion, err := getClusterVersion(mgr.GetAPIReader())
```
simonpasquier, line 66: How does it work with cluster upgrades? We don't need to answer in this PR but maybe I expect that "something" watches the clusterversion object and reconciles the state when a change happens.

simonpasquier (commented):
Good work! I wonder if we couldn't split the PR into 2 for easier review:
1. Add support for OpenShift feature gate.
2. Implement the observability UI support.

pkg/controllers/observability-ui/observability-ui-plugin/components.go
```diff
@@ -0,0 +1,198 @@
...
+			APIVersion: appsv1.SchemeGroupVersion.String(),
+			Kind:       "Deployment",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      info.Name,
+			Namespace: namespace,
+			Labels:    componentLabels(info.Name),
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: func(i int32) *int32 { return &i }(1),
```
simonpasquier, line 88: (nit) we could use `ptr.To(int32(1))` instead

pkg/apis/observability-ui/v1alpha1/types.go
tremes: I think this is still problematic (enum required validation). Maybe the `ObservabilityUIPluginSpec` should not have `omitempty` for validation to happen and it should probably be `// +kubebuilder:validation:Enum=Dashboards`

Maybe we could also declare `Dashboards` as the default value, but I don't know what the future plans are.

pkg/apis/observability-ui/v1alpha1/types.go
```diff
@@ -0,0 +1,120 @@
...
+// ObservabilityUIPluginList contains a list of ObservabilityUIPlugin
+// +kubebuilder:resource
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type ObservabilityUIPluginList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ObservabilityUIPlugin `json:"items"`
+}
+
+// +kubebuilder:validation:Enum=dashboards
```
simonpasquier, line 34:
```suggestion
// +kubebuilder:validation:Enum=Dashboards
```

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jgbernalp: - I adjusted the code to exit early when the feature gate is enabled but running on vainilla k8s, so it effectively fails the reconciler
- Yes, this could be the case. We can adjust the log message
- Yes, versions before 4.11 don't have the plugin CRD

If we have a cluster upgrade then it can start on false and become true for a supported version.

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jgbernalp: Created a task to follow up on this: https://issues.redhat.com/browse/COO-106

pkg/apis/observability-ui/v1alpha1/types.go
```diff
@@ -0,0 +1,120 @@
+// +groupName=observabilityui.rhobs
+// +kubebuilder:rbac:groups=observabilityui.rhobs,resources=observabilityuiplugins,verbs=list;get;watch
```
jan--f, line 2: Group names were recently discussed in the Obsservability group context. Seems like we're going to align on `observability.openshift.io` without any further subdomains. So Maybe we should go with `observabilityuiplugins.observability.openshift.io` here?
See also https://redhat-internal.slack.com/archives/C02BY4191B6/p1712866056544739

simonpasquier: (nit) uiplugins.observability.openshift.io to avoid repetition?

Makefile
```diff
@@ -193,7 +193,7 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
 			--kustomize-dir=deploy/olm \
 			--package=observability-operator \
 		 	$(BUNDLE_METADATA_OPTS)
-	sed -e 's|<IMG_OBSERVABILITY_OPERATOR>|$(OPERATOR_IMG)|g' \
```
jan--f, line 196: This isn't really needed is it?

pkg/apis/observability-ui/v1alpha1/types.go
jan--f: @tremes wdyt about the current version here?

Makefile
jgbernalp: No. Is now fixed. I had trouble running the make bundle in my mac 🍏 as `sed` works differently here.

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jan--f: > If we have a cluster upgrade then it can start on false and become true for a supported version.

That currently requires a operator restart though doesn't it? I suppose this scenario is covered in https://issues.redhat.com/browse/COO-106?

@simonpasquier are you ok with this for now or does this need any changes still?

jan--f:
I think this looks good to go forward. I would like to see the two open threads get resolved before merging this though.

pkg/apis/observability-ui/v1alpha1/register.go
periklis, line 1: nit. By the rename from `observability-ui` to `uiplugin`, the folders need to be renamed accordingly.

pkg/controllers/observability-ui/observability-ui-plugin/compatibility_matrix.go
periklis, line 1: nit. By the rename from `observability-ui` to `uiplugin`, the folders need to be renamed accordingly.

pkg/controllers/observability-ui/observability-ui-plugin/components.go
periklis, line 1: I suggest to move all repetitive constant values like ports, names, replicas into constants.

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	generationChanged := builder.WithPredicates(predicate.GenerationChangedPredicate{})
+
+	ctrl, err := ctrl.NewControllerManagedBy(mgr).
+		For(&obsui.UIPlugin{}).
+		Owns(&appsv1.Deployment{}, generationChanged).
+		Owns(&v1.Service{}, generationChanged).
+		Owns(&v1.ServiceAccount{}, generationChanged).
+		Owns(&rbacv1.ClusterRole{}, generationChanged).
+		Owns(&rbacv1.ClusterRoleBinding{}, generationChanged).
+		Owns(&osv1alpha1.ConsolePlugin{}, generationChanged).
```
periklis, line 89: A namespaced resources such as UIPlugin are not allowed to have cluster-scoped depedents. This can lead into poisoning the k8s resource garbage collector like unwanted deletes of the UIPlugin resource. Source: https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return nil
+}
+
+func getClusterVersion(k8client client.Reader) (*configv1.ClusterVersion, error) {
+	clusterVersion := &configv1.ClusterVersion{}
+	key := client.ObjectKey{Name: "version"}
+	if err := k8client.Get(context.TODO(), key, clusterVersion); err != nil {
+		return nil, err
+	}
```
periklis, line 104: Double checking here. Is a NotFound error also worth to return and in turn reconcile once again? Usually NotFound errors are treated as non-queuing. WDYT?

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
```diff
@@ -0,0 +1,218 @@
...
+				Plugins: []string{
+					pluginInfo.ConsoleName,
+				},
+			},
+		}
+
+		if err := reconciler.NewMerger(cluster).Reconcile(ctx, rm.k8sClient, rm.scheme); err != nil {
+			return err
+		}
+	}
```
periklis, line 198:
```suggestion
	if slices.Contains(cluster.Spec.Plugins, pluginInfo.ConsoleName) {
	    return nil
	}
	
	// Register the plugin with the console
	cluster := &operatorv1.Console{
		TypeMeta: metav1.TypeMeta{
			APIVersion: operatorv1.GroupVersion.String(),
			Kind:       "Console",
		},
		ObjectMeta: metav1.ObjectMeta{
			Name: "cluster",
		},
		Spec: operatorv1.ConsoleSpec{
			OperatorSpec: operatorv1.OperatorSpec{
				ManagementState: operatorv1.Managed,
			},
			Plugins: []string{
				pluginInfo.ConsoleName,
			},
		},
	}

	if err := reconciler.NewMerger(cluster).Reconcile(ctx, rm.k8sClient, rm.scheme); err != nil {
			return err
	}

       return nil
```

pkg/controllers/observability-ui/observability-ui-plugin/plugin_info_builder.go
```diff
@@ -0,0 +1,106 @@
...
+	ClusterRole        *rbacv1.ClusterRole
+	ClusterRoleBinding *rbacv1.ClusterRoleBinding
+}
+
+func PluginInfoBuilder(plugin *obsui.UIPlugin, pluginConf UIPluginsConfiguration, clusterVersion string) (*UIPluginInfo, error) {
+	imageKey, err := getImageKeyForPluginType(plugin.Spec.Type, clusterVersion)
+
+	if err != nil {
+		return nil, err
+	}
```
periklis, line 29:
```suggestion
	imageKey, err := getImageKeyForPluginType(plugin.Spec.Type, clusterVersion)
	if err != nil {
		return nil, err
	}
```

pkg/controllers/observability-ui/observability-ui-plugin/plugin_info_builder.go
```diff
@@ -0,0 +1,106 @@
...
+
+	image := pluginConf.Images[imageKey]
+
+	if image == "" {
+		return nil, fmt.Errorf("no image provided for plugin type %s with key %s", plugin.Spec.Type, imageKey)
+	}
```
periklis, line 35:
```suggestion
	image := pluginConf.Images[imageKey]
	if image == "" {
		return nil, fmt.Errorf("no image provided for plugin type %s with key %s", plugin.Spec.Type, imageKey)
	}
```

periklis (changes requested):

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jgbernalp: This function gets called before the controller is created so IIUC is not part of the reconciliation. Meaning, if getting the cluster version fails, the controller does not register with the manager.

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jan--f: I would expect `Owns` to do the right thing here. First and foremost this sets up a watch. And since this controller generates the resource using `Owns` should be the right approach. If this sets an OwnerReference I'd consider this a bug in controller runtime tbh. At least I can't find anything in their documentation that calls this out.

docs/user-guides/observability-ui-plugins.md
```diff
@@ -0,0 +1,23 @@
...
+The plugin will search for datasources as ConfigMaps in the `console-dashboards` namespace with the `console.openshift.io/dashboard-datasource: 'true'` label. The namespace `console-dashboards` is required, more details on how to create a datasource ConfigMap can be found in the [console-dashboards-plugin](https://github.com/openshift/console-dashboards-plugin/blob/main/docs/add-datasource.md)
+
+To enable the console dashboards plugin, create a `UIPlugin` CR. The following example shows how to create a CR to enable the console dashboards plugin:
+
+```yaml
+apiVersion: observability.openshift.io/v1alpha1
+kind: UIPlugin
+metadata:
+  name: ui-dashboards
+  namespace: openshift-observability-ui
```
simonpasquier, line 20: Users can't create projects starting by "openshift-".
```suggestion
  namespace: observability-ui
```

pkg/controllers/uiplugin/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	if err := rm.registerPluginWithConsole(ctx, pluginInfo); err != nil {
+		return rm.updateStatus(ctx, req, plugin, err), err
+	}
+
+	return rm.updateStatus(ctx, req, plugin, nil), nil
+}
+
+func (rm resourceManager) updateStatus(ctx context.Context, req ctrl.Request, pl *uiv1alpha1.UIPlugin, recError error) ctrl.Result {
```
simonpasquier, line 177: I see the status subresource being constantly updated (e.g. `lastTransitionTime` increases over time). It also triggers the reconciliation loop ever and ever.

pkg/controllers/uiplugin/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	err := rm.k8sClient.Get(ctx, name, current)
+
+	return err == nil || !metaerrors.IsNoMatchError(err)
+}
+
+func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := rm.logger.WithValues("plugin", req.NamespacedName)
+
+	if !rm.consolePluginCapabilityEnabled(ctx, req.NamespacedName) {
+		logger.Info("Cluster console plugin not supported or not accesible. Skipping observability UI plugin reconciliation")
```
simonpasquier, line 127:
```suggestion
		logger.Info("Cluster console plugin not supported or not accessible. Skipping observability UI plugin reconciliation")
```

pkg/controllers/uiplugin/components.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func newConsolePlugin(info UIPluginInfo, namespace string) *osv1alpha1.ConsolePlugin {
+	return &osv1alpha1.ConsolePlugin{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: osv1alpha1.SchemeGroupVersion.String(),
+			Kind:       "ConsolePlugin",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: info.ConsoleName,
```
simonpasquier, line 67: IIUC we can't create more than 1 UIPlugin object per plugin type. E.g. if I create 2 UIPlugins with `type: Dashboards` then the controller will constantly update the same ConsolePlugin object.

Also ConsolePlugin is cluster-scoped while UIPlugin is namespace-scoped so we can't have owner ref from UIPlugin to ConsolePlugin. How do we deal with deletion? Should we have finalizers?

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
simonpasquier: I've checked on a live cluster and the controller doesn't seem to set the owner ref for the cluster role. That being said, deleting the UIPlugin object doesn't remove the cluster role (for the same reason).

deploy/operator/kustomization.yaml
```diff
@@ -38,6 +38,14 @@ patches:
     group: apps
     kind: Deployment
     version: v1
+- patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --images=ui-dashboards=quay.io/openshift-observability-ui/console-dashboards-plugin:v0.1.0
```
jan--f, line 44: not that it's any harm but this just passes the default value right doesn't it?

cmd/operator/main.go
```diff
@@ -84,6 +86,8 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
 	flag.Var(images, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", imagesUsed()))
+	flag.BoolVar(&openShiftEnabled, "openshift.enabled", true, "Enable OpenShift specific features such as Console Plugins.")
```
jan--f, line 89: Nur sure we want to enable the feature gate by default. But in order to not delay this further we can flip this and adjust the test setup in a follow up.

simonpasquier: Let's not forget about it then if we don't merge it with enabled=true by default :)

pkg/controllers/uiplugin/controller.go
```diff
@@ -0,0 +1,280 @@
...
+)
+
+// RBAC for managing UIPlugins
+// +kubebuilder:rbac:groups=observability.openshift.io,resources=uiplugins,verbs=get;list;watch;create;update;delete;patch
+// +kubebuilder:rbac:groups=observability.openshift.io,resources=uiplugins/status,verbs=get;update
+
+// RBAC for managing observability ui plugin objects
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=list;watch;create;update;delete;patch
+//+kubebuilder:rbac:groups="",resources=serviceaccounts;services;configmaps,verbs=get;list;watch;create;update;patch;delete
```
simonpasquier, line 63: Do we really need create/update/patch/delete permissions?
It can be adjusted as a follow-up and in fact, we probably want to review/tidy up the operator permissions after the PR merges.

deploy/operator/observability-operator-cluster-role.yaml
```diff
@@ -133,6 +167,42 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - observability.openshift.io
+  resources:
+  - uiplugins
+  verbs:
+  - create
```
simonpasquier, line 175: same remark here regarding permissions

docs/user-guides/observability-ui-plugins.md
```diff
@@ -0,0 +1,22 @@
...
+
+Using the Observability UI, you can install and manage plugins that extend the observability functionality of the OpenShift web console. Plugins are installed and managed by the Observability Operator.
+
+## Plugins
+
+- [dashboards](#dashboards): Add enhanced dashboards to the OpenShift web console. This plugin allows you to add other Prometheus datasources present in the cluster, apart from the in-cluster one, to the default dashboards.
+
+### Dashboards
+
+The plugin will search for datasources as ConfigMaps in the `openshift-config-managed` namespace with the `console.openshift.io/dashboard-datasource: 'true'` label. The namespace `openshift-config-managed` is required, more details on how to create a datasource ConfigMap can be found in the [console-dashboards-plugin](https://github.com/openshift/console-dashboards-plugin/blob/main/docs/add-datasource.md)
```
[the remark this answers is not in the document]
jgbernalp: The plugin will search only in the `openshift-config-managed` namespace. This can be adjusted in the future to be a configurable value. But now it follows the same namespace as the console dashboards.

deploy/operator/kustomization.yaml
jgbernalp: yes, I wanted to reflect the new image for the plugin in the generated files in case some replacement needs it for the bundle productization

cmd/operator/main.go
jgbernalp: I added a task for that [COO-120](https://issues.redhat.com/browse/COO-120) let me know if the details are correct.

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	name := "observability-ui-" + plugin.Name
+	namespace := pluginConf.ResourcesNamespace
+
+	switch plugin.Spec.Type {
+	case uiv1alpha1.TypeDashboards:
+		{
+			readerRoleName := plugin.Name + "-datasource-reader"
+			datasourcesNamespace := "openshift-config-managed"
```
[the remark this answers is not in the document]
jgbernalp: This can be configured in the plugin using the `--dashboards-namespace` flag but its default is `openshift-config-managed`. At this point there is a single deployment template used for any plugin. This might change when we add support for other plugins that need specific options.

pkg/controllers/uiplugin/controller.go
jgbernalp: I created a follow up task https://issues.redhat.com/browse/COO-121

deploy/operator/observability-operator-cluster-role.yaml
jgbernalp: I created a follow up task https://issues.redhat.com/browse/COO-121

pkg/controllers/observability-ui/observability-ui-plugin/controller.go
jgbernalp: I updated the UIPlugin to be a cluster scoped as it should be, and also changed the ClusterRole and ClusterRoleBinding to Role and RoleBinding. While doing so I realized the reconciler was only considering owner and resources with the same namespace. I wonder if [this](https://github.com/rhobs/observability-operator/pull/434/files#diff-fc2bc6354d4506ced4adfd45dda2737dc917c217e282041d9af5bce57e514a85R29) would be the right way to check for cluster scoped resources that own namespaced ones:

pkg/controllers/uiplugin/controller.go
jgbernalp: As discussed, this was caused by 2 instances trying to reconcile the same resources. I checked the kubebuilder validation, but wasn't able to find something that checks cross resources. It seems to run validation only on the current resource, so I'm inclined to write an admission webhook unless @jan--f knows a validation rule for this case.

jan--f: This is to validate that there is only a single UIplugin per cluster?

The trick for that is to add a validation to the yaml manifest (afaik this can't yet be added via a go comment, but we can use kustomize) that enforces a predefined name for the object and prevents any other object names.
See for example https://github.com/rhobs/observability-operator/compare/main...jan--f:observability-operator:config-crd#diff-9d14e1263c27af9a367892b64032baaa69bed686ae7e9c63cb5057fe6f3328eaR9-R26

jan--f: But as I said, I could see the being added in a later PR

deploy/operator/kustomization.yaml
jan--f: Got it. For the productuzed CSV we will need to change this, but this can only happen in CPaaS, since we use the digest pinned image reference. Feel free to remove this here.

pkg/reconciler/reconciler.go
```diff
@@ -25,7 +25,8 @@ type Updater struct {
 }
 
 func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
-	if r.resourceOwner.GetNamespace() == r.resource.GetNamespace() {
+	// If the resource owner is in the same namespace as the resource, or if the resource owner is cluster scoped set the owner reference.
+	if r.resourceOwner.GetNamespace() == r.resource.GetNamespace() || r.resourceOwner.GetNamespace() == "" {
```
simonpasquier, line 29: :+1:

simonpasquier:
I didn't test/review in depth the latest version of the PR but you've got my virtual `/lgtm` :)

danielmellado:
Besides the nit issue on the commit-lint IMHO this is ready to merge. @jgbernalp mind addressing that? Tnx!

pkg/controllers/uiplugin/controller.go
jgbernalp: Multiple UIPugins can exists in a cluster, but only one per type (Dashboards, Logging, DistributedTracing, etc). I added a task: https://issues.redhat.com/browse/COO-122

jan--f:
Works as expected, follow-up tasks for some minor nits are created and tracked. Good to go, thanks everyone!
/lgtm

jan--f (approved):

PR #447, opened 2024-03-26
PoC for general processing of status conditions for multiple operands

tremes:
This is just a proposal (maybe too many unstructured and type converting) for more general processing of status conditions for multiple operands. Only one shared `Condition` type is assumed (see https://github.com/rhobs/observability-operator/pull/434#discussion_r1524685405). 

The idea is based on the `Operand` type (see [here)](https://github.com/rhobs/observability-operator/compare/main...tremes:conditions-universal-qmark?expand=1#diff-31566b75036aa0a4df6f92d7800989389459ed2a837dfee57b483f31976af243R33), which is a wrapper for `client.Object`, but also introduces flags that tell what it affects in terms of status conditions. 

The disadvantage is that the condition is overwritten (see [here](https://github.com/rhobs/observability-operator/compare/main...tremes:conditions-universal-qmark?expand=1#diff-26efd3a8f6ca1b25a7118f612eafacaea09b60c0489b29a6df2993c346d8cf01R36), it also means that the order of the operands matters) and we could probably think about some more sophisticated "aggregating" approach. 

Question is how to calculate or aggregate the `observedGeneration` value when a stack has more than operand.


pkg/status/conditions.go
```diff
@@ -0,0 +1,241 @@
...
+func getConditionByType(conditions []v1alpha1.Condition, t v1alpha1.ConditionType) (v1alpha1.Condition, error) {
+	for _, c := range conditions {
+		if c.Type == t {
+			return c, nil
+		}
+	}
+	return v1alpha1.Condition{}, fmt.Errorf("ERROR: condition type %v not found", t)
+}
+
+func getConditionsFromObject(o client.Object) ([]v1alpha1.Condition, error) {
```
tremes, line 193: Perhaps we could avoid this function by using some interface like:
```
type StatusReporting interface {
	Conditions() []Condition
	Geeneration() int64
}
```
And require types like `MonitoringStack` to implement this interface.

tremes: @jan--f I was updating this PR this morning a bit (locally so far) and I think this ^^ can probably make few things more simple here. 
What do you think about this idea? The `Operand` type introduced here can then use this `StatusReporting` type instead of `client.Object` (but it's just theory. I haven't tried yet)

tremes: Hm actually the theory was not correct :) I guess it won't help much with the conditions reading from the operands.

tremes:
This needs updates and it would probably be good to split into more PRs.

PR #450, opened 2024-04-02
apis: provide option to enable otlp/http receiver

frzifus:
This PR adds an option to enable a featuregate on prometheus to accept data on the otlp/http ingestion path.

Is that something we can a here or should it be an option to accept multiple featuregates?

TODO:
- [x] call `make generte` (currently it panics)

cc @simonpasquier

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/types.go
+++ b/pkg/apis/monitoring/v1alpha1/types.go
@@ -184,6 +184,11 @@ type PrometheusConfig struct {
 	// Enable Prometheus to be used as a receiver for the Prometheus remote write protocol. Defaults to the value of `false`.
 	// +optional
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
+	// Enable Prometheus to accept OpenTelemetry Metrics via the otlp/http protocol.
+	// Defaults to the value of `false`.
+	// The resulting endpoint is /api/v1/otlp/v1/metrics.
+	// +optional
+	EnableOtlpHttpReceiver *bool `json:"enableOtlpHttpReceiver,omitempty"`
 	// Default interval between scrapes.
 	// +optional
 	ScrapeInterval *monv1.Duration `json:"scrapeInterval,omitempty"`
--- a/bundle/manifests/monitoring.rhobs_monitoringstacks.yaml
+++ b/bundle/manifests/monitoring.rhobs_monitoringstacks.yaml
@@ -111,6 +111,12 @@ spec:
                   replicas: 2
                 description: Define prometheus config
                 properties:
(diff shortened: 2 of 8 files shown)


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -184,6 +184,10 @@ type PrometheusConfig struct {
 	// Enable Prometheus to be used as a receiver for the Prometheus remote write protocol. Defaults to the value of `false`.
 	// +optional
 	EnableRemoteWriteReceiver bool `json:"enableRemoteWriteReceiver,omitempty"`
+	// Enable Prometheus to accept OpenTelemetry Metrics via the otlp/http protocol.
+	// Defaults to the value of `false`.
+	// +optional
+	EnableOtlpHttpReceiver bool `json:"enableOtlpHttpReceiver,omitempty"`
```
pavolloffay, line 190: IIRC the prometheus OTLP endpoint is under some URL prefix. 

It might be good to document it here.

pavolloffay (approved):

pkg/apis/monitoring/v1alpha1/types.go
frzifus: I can do. Its `/api/v1/otlp`...`/v1/metrics`

simonpasquier:
> Is that something we can a here or should it be an option to accept multiple featuregates?

Good question. My instinct is to say that we want a simple way for Obs-Operator users to toggle OTLP ingestion and while they can use server-side apply to customize `.spec.enableFeatures`, it's way more convenient to expose this in the MonitoringStack API.

The direct consequence though is that when OTLP is enabled, it's not possible for a user to add more values to `.spec.enableFeatures`. It should be easily fixable on prometheus-operator side though using `+listType:=set`.

simonpasquier:
re https://github.com/rhobs/observability-operator/pull/450#issuecomment-2034908676, i've submitted https://github.com/prometheus-operator/prometheus-operator/pull/6471

pkg/apis/monitoring/v1alpha1/types.go
simonpasquier, line 190: (suggestion) I'd make it a bool pointer so we can decide later to have it on by default (e.g. opt-in -> opt-out) if needed.
```suggestion
        // +optional
	EnableOtlpHttpReceiver *bool `json:"enableOtlpHttpReceiver,omitempty"`
```

frzifus: But it is already `+optional`? :)

simonpasquier:
/lgtm

@jan--f for the final approval

jan--f (approved):
/lgtm

Thanks everyone!

PR #454, opened 2024-04-05
chore: fix assertAlertmanagersAreResilientToDisruption()

simonpasquier:
I suspect that the test was flaky because it uses the address of the loop variable (see https://go.dev/blog/loopvar-preview).


simonpasquier:
After all I'm not sure that it's the bug because it failed on the same test. Trying an updated version where all evictions are executed serially then we check the errors.

lihongyan1:
The case failed again and even failed in the upstream ci, refer 
https://github.com/rhobs/observability-operator/actions/runs/8660157870/job/23747502573?pr=448

simonpasquier:
@lihongya indeed I've seen the test failed. I tried to simulate the eviction scenario manually but the PDB always kicked in to reject the last eviction. I know that @jan--f investigated too but I wonder if there's a flaw in the test or a bug in Kubernetes :thinking:

PR #460, opened 2024-04-17
COO-117: fix: remove invalid owner ref on cluster role

simonpasquier:
Cluster roles aren't namespace-scoped and can't have namespace-scoped owner references.

Closes [COO-117](https://issues.redhat.com//browse/COO-117)

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -25,7 +25,6 @@ type Updater struct {
 }
 
 func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime.Scheme) error {
-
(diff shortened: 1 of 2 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -39,6 +38,7 @@ func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime
 			r.resource.GetNamespace(), r.resource.GetName(),
 			r.resource.GetObjectKind().GroupVersionKind().String(), err)
 	}
+
```
danielmellado, line 41: Thanks also for this style clean up. I'll try to have that automatically applied for the whole project ;)

PR #475, opened 2024-05-08
fix: default goal in makefile and add goal for unit tests

xperimental:
Three changes, but all in the Makefile:

- Fix git command to ignore external diff tools
- Fix definition of default target
- Add a target for unit tests (optional, but I like being able to run tests with a simple command)

I can split this up into separate PRs, if you like.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,10 @@ all: operator
 
 ## Development
 
+.PHONY: test-unit
+test-unit:
+	go test -cover ./cmd/... ./pkg/...
+
 .PHONY: lint
 lint: lint-golang lint-jsonnet lint-shell
 
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ OSD_E2E_TEST_HARNESS_IMG=$(IMAGE_BASE)-test-harness:$(VERSION)
 OSD_E2E_TEST_HARNESS_IMG_LATEST=$(IMAGE_BASE)-test-harness:latest
 
 # running `make` builds the operator (default target)
-all: operator
+.DEFAULT_GOAL := operator
 
 ## Development
 


Makefile
```diff
@@ -15,10 +15,14 @@ OSD_E2E_TEST_HARNESS_IMG=$(IMAGE_BASE)-test-harness:$(VERSION)
 OSD_E2E_TEST_HARNESS_IMG_LATEST=$(IMAGE_BASE)-test-harness:latest
 
 # running `make` builds the operator (default target)
-all: operator
+.DEFAULT_GOAL := operator
```
jan--f, line 18: Afaiu the `.DEFaULT_GOAL` variable is intended for more complex use cases. The make docs say

> The .DEFAULT_GOAL variable allows you to discover the current default goal, restart the default goal selection algorithm by clearing its value, or to explicitly set the default goal.

I don't a benefit over just letting make pick the first target as the default.

Makefile
```diff
@@ -15,10 +15,14 @@ OSD_E2E_TEST_HARNESS_IMG=$(IMAGE_BASE)-test-harness:$(VERSION)
...
 
 ## Development
 
+.PHONY: test-unit
+test-unit:
+	go list ./... | grep -v test/e2e | xargs go test -cover
```
jan--f, line 24:
```suggestion
	go test -cover ./cmd/... ./pkg/...
```

Makefile
```diff
@@ -199,7 +203,7 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) generate
 		--select-optional name=operatorhub \
 		--optional-values=k8s-version=1.21 \
 		--select-optional suite=operatorframework
-	git diff --quiet -I'^    createdAt: ' bundle && git checkout bundle || true
+	git diff --no-ext-diff --quiet -I'^    createdAt: ' bundle && git checkout bundle || true
```
jan--f, line 206: Does this cause errors in any of our automations?

If not, I'm generally no a fan of local configurations/environments requiring changes in common tooling. If your difftool does not report an appropriate exit code, I'd suggest to start there.

xperimental: Honestly, this was mostly a quick fix to fix my own workflow and I understand that this _should_ usually not make it into the tooling, otherwise it'll get messy soon.

That said, now that I have looked through git docs and looked at how parameters are passed to the diff tool, I don't think `-I` is ever meant to interact properly with external tools. Both the `--quiet` and `-I` do not seem to be passed on to the external tool at all. So you could argue that `-I` is directly depending on the diff tool output and so you need to avoid running "unexpected tools".

I'm also open to just removing this parameter from the PR again, though.

Makefile
xperimental: The advantage of `.DEFAULT_GOAL` is that you explicitly set the default goal instead of relying on the structure of the Makefile to define the default goal. For example, in this Makefile we `include` another file above this declaration, which already defines some goals of which one becomes the default goal.

We can either move the goal we want as default further up in the file (above the `include`, in this case) or define what we want explicitly (which is what I opted for).

jan--f: Fair point, I didn't think of the import.

jan--f (approved):
/lgtm
/approve

PR #477, opened 2024-05-08
feat: support logging as ui-plugin

xperimental:
Implementation for [COO-147](https://issues.redhat.com/browse/COO-147).


cmd/operator/main.go
```diff
@@ -36,10 +36,11 @@ import (
...
 var defaultImages = map[string]string{
-	"prometheus":    "",
-	"alertmanager":  "",
-	"thanos":        obopo.DefaultThanosImage,
-	"ui-dashboards": "quay.io/openshift-observability-ui/console-dashboards-plugin:v0.1.0",
+	"prometheus":          "",
+	"alertmanager":        "",
+	"thanos":              obopo.DefaultThanosImage,
+	"ui-dashboards":       "quay.io/openshift-observability-ui/console-dashboards-plugin:v0.1.0",
+	"logging-view-plugin": "quay.io/openshift-logging/logging-view-plugin:6.0.0",
```
jgbernalp, line 43: just a nit, we are using the `ui-` prefix
```suggestion
	"ui-logging": "quay.io/openshift-logging/logging-view-plugin:6.0.0",
```

pkg/controllers/uiplugin/compatibility_matrix.go
```diff
@@ -25,29 +25,66 @@ var compatibilityMatrix = []CompatibilityEntry{
 		ImageKey:          "ui-dashboards",
 		Features:          []string{},
 	},
+	{
+		PluginType:        uiv1alpha1.TypeLogging,
+		MinClusterVersion: "v4.11",
+		MaxClusterVersion: "v4.11",
+		ImageKey:          "logging-view-plugin",
```
jgbernalp, line 32:
```suggestion
		ImageKey:          "ui-logging",
```

pkg/controllers/uiplugin/logging.go
```diff
@@ -0,0 +1,129 @@
...
+		ResourceNamespace: namespace,
+		Proxies: []osv1alpha1.ConsolePluginProxy{
+			{
+				Type:      "Service",
+				Alias:     "backend",
+				Authorize: true,
+				Service: osv1alpha1.ConsolePluginProxyServiceConfig{
+					Name:      fmt.Sprintf("%s-gateway-http", plugin.Spec.Logging.LokiStack.Name),
+					Namespace: "openshift-logging", // TODO decide if we want to support LokiStack in other namespaces
+					Port:      8080,
```
jgbernalp, line 55: Is the port fixed for a lokiStack?

bundle/manifests/observability-operator.clusterserviceversion.yaml
```diff
@@ -690,6 +698,7 @@ spec:
                 - --images=alertmanager=quay.io/prometheus/alertmanager:v0.26.0
                 - --images=prometheus=quay.io/prometheus/prometheus:v2.49.1
                 - --images=thanos=quay.io/thanos/thanos:v0.33.0
+                - --openshift.enabled=true
```
jan--f, line 701: Please don't add this here, this will break our e2e tests. We add this flag in the downstream builds

xperimental: Don't worry, I just did not know how this is done properly. It's a temporary change. It's in a commit with a "remove before merge" message :slightly_smiling_face:

pkg/controllers/uiplugin/logging.go
xperimental: Yes, the Loki Operator currently provides no way to customize the port.

cmd/operator/main.go
xperimental: Have updated the code here (and in the other places internal to the COO) accordingly, but the plugin (effectively the name of the ConsolePlugin resource) is still called `logging-view-plugin`, because for that name-change to work it would also need an update to the [`plugin-manifest.json`](https://github.com/openshift/logging-view-plugin/blob/main/config/plugin-manifest.json#L2).

xperimental:
Cleaned up the PR to only contain the changes needed for the logging-ui. Are we ok with the API of UIPlugin as-is? If yes, I'd remove the draft status.

JoaoBraveCoding (commented):
from my side lgtm (didn't manually test it)

JoaoBraveCoding (approved):
Tested on OpenShift and it works as expected 👍 However I had to manually add the "openshift.enabled"... I'm guessing that in the future there will be a version on OLM with this flag already enabled.

To test:
- Install loki-operator using the console UI
- Install cluster-logging operator using version in https://github.com/xperimental/cluster-logging-operator/tree/block-console-5.9
- Install observability-operator using this PR
- Edit CSV to add "--openshift.enabled"
- Create ClusterLogging resource
- Create ClusterLogForwarder resource
- Create LokiStack resource called "lokistack-dev"
- Create UIPlugin with type "logging" and pass the name "lokistack-dev"

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,50 @@ type UIPluginList struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Log Limit",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleLogLimit"}
+	LogsLimit int `json:"logsLimit,omitempty"`
+
+	// Timeout is the max duration before a query timeout
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
+	Timeout string `json:"timeout,omitempty"`
+}
```
periklis, line 66: The spec is missing the following two fields according to the [docs](https://github.com/openshift/logging-view-plugin?tab=readme-ov-file#configuration-values):
- `alertingRuleTenantLabelKey`
- `alertingRuleNamespaceLabelKey`

@jgbernalp are they omitted by design?

periklis (changes requested):
In general looks good to me besides some suggestions to improve handling and clarifying missing fields. We certainly need to add support for `nodeSelector` and `tolerations` to complete this work and stay in course with OpenShift Logging releases.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,50 @@ type UIPluginList struct {
...
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Log Limit",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleLogLimit"}
+	LogsLimit int `json:"logsLimit,omitempty"`
+
+	// Timeout is the max duration before a query timeout
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
+	Timeout string `json:"timeout,omitempty"`
```
simonpasquier, line 65: Do we even need to support a unit suffix (and assume seconds)? If yes, I suppose that it can't be omitted.

pkg/apis/uiplugin/v1alpha1/types.go
jgbernalp: They should not be omitted, they are seldom used. But they are still part of the allowed configuration.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,50 @@ type UIPluginList struct {
...
+	// It always references a LokiStack in the "openshift-logging" namespace.
+	//
+	// +kubebuilder:validation:Required
+	LokiStack corev1.LocalObjectReference `json:"lokiStack"`
+
+	// LogsLimit is the max number of entries returned for a query.
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Log Limit",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleLogLimit"}
+	LogsLimit int `json:"logsLimit,omitempty"`
```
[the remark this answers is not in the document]
xperimental: The readme is wrong. It's `logsLimit` on both [CLO](https://github.com/openshift/cluster-logging-operator/blob/release-5.9/api/logging/v1/clusterlogging_types.go#L172) and [logging-view-plugin](https://github.com/openshift/logging-view-plugin/blob/main/pkg/server/server.go#L35) side.

pkg/apis/uiplugin/v1alpha1/types.go
xperimental: The timeout here is usually in the order of minutes and I think it's easier (and more "visual") to configure `3m` instead of `180`. The default unit is seconds already.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,50 @@ type UIPluginList struct {
...
 type UIPluginSpec struct {
 	// Type defines the UI plugin.
 	// +required
 	// +kubebuilder:validation:Required
 	Type UIPluginType `json:"type"`
+
+	// Logging contains configuration for the logging console plugin.
+	//
+	// +kubebuilder:validation:Optional
+	Logging *LoggingConfig `json:"logging,omitempty"`
```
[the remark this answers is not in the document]
xperimental: Added a validation to the `UIPluginSpec`. Was this what you had in mind?

xperimental:
Some updates from the code-review comments. Not done with all the changes yet, though. Notably, the additional options for the logging-view-plugin and the `nodeSelector`/`tolerations` part is still missing.

pkg/apis/uiplugin/v1alpha1/types.go
xperimental: Currently this PR only includes the options that were present on the CLO API. I can add the other options as well, but then it might make sense to group them below another struct again. I'll take a stab at this tomorrow.

pkg/apis/uiplugin/v1alpha1/types.go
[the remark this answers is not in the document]
xperimental: It looks as if the condition I added does not actually work as intended. I'll have another look tomorrow. So far, we have only done these kind of validation in the operator (signaling the error using a Condition) and not directly in the API-server.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,68 @@ type UIPluginList struct {
...
+	// +kubebuilder:validation:Required
+	LokiStack LokiStackReference `json:"lokiStack"`
+
+	// LogsLimit is the max number of entries returned for a query.
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Log Limit",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleLogLimit"}
+	LogsLimit int `json:"logsLimit,omitempty"`
+
+	// Timeout is the maximum duration before a query timeout.
```
simonpasquier, line 59:
```suggestion
	// Timeout is the maximum duration before a query timeout.
	// The value is expected to be a sequence of digit numbers followed by an optional unit suffix which can 's' (seconds), 'm' (minutes), 'h' (hours) or 'd' (days).
	// If the suffix is omitted, it defaults to seconds.
```

pkg/apis/uiplugin/v1alpha1/types.go
simonpasquier: Ack. But hours and days seem a bit overkill?
Also I fail to see in https://github.com/openshift/logging-view-plugin/ where it assumes seconds if the unit's missing.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,68 @@ type UIPluginList struct {
...
 	TypeDashboards UIPluginType = "Dashboards"
+
+	// TypeLogging deploys the Logging View Plugin for OpenShift Console.
+	TypeLogging UIPluginType = "Logging"
 )
 
+// LoggingConfig contains options for configuring the logging console plugin.
+type LoggingConfig struct {
+	// LokiStack points to the LokiStack instance of which logs should be displayed.
+	// It always references a LokiStack in the "openshift-logging" namespace.
```
simonpasquier, line 48: AFAIU we can only have 1 logging UIPlugin object in the cluster but we could have multiple LokiStack objects. Is this correct?

pkg/apis/uiplugin/v1alpha1/types.go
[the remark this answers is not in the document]
simonpasquier: As I said above, I'm fine if we don't validate at the API layer but it should be clearly stated in the API documentation.

pkg/apis/uiplugin/v1alpha1/types.go
periklis: FYI we don't support multiple LokiStack objects through this UIPlugin in OCP. Although NetObserv saves partially data in their own LokiStack instance, they still use their own UI based on a custom console plugin

pkg/apis/uiplugin/v1alpha1/types.go
xperimental: I assumed it used seconds, because when I looked through the source briefly I spotted a function converting it to seconds for the internal representation for the JavaScript backend.

When actually testing this, it became more complex though. I have now added code that actually does what we advertise (choice of seconds and minutes, default of seconds) in the operator and formats the ConfigMap passed to the logging-view-plugin accordingly.

pkg/apis/uiplugin/v1alpha1/types.go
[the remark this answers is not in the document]
xperimental: Updated the validation rule and it now seems to work. Have a look.

xperimental (commented):
Updated the PR for the code-review feedback again.

While testing for the `timeout` in the ConfigMap I noticed that a changing ConfigMap also mandates a redeploy of the Deployment, so I added code for hashing the ConfigMap's contents so that restarts happen automatically.

---

One open question for me is, whether the options passed to the `logging-view-plugin` need another "level of indentation", especially if we plan to increase the number of options from the two that are currently available.

To illustrate my point: Currently the configuration would look like this:

```yaml
kind: UIPlugin
metadata:
  name: ui-logging
spec:
  type: Logging
  logging:
    lokiStack:
      name: lokistack-dev
    timeout: 3m
    logsLimit: 1000
```

And with another level of indentation/grouping it would look like this (think of `frontend` as a placeholder, I couldn't think of anything better so far):

```yaml
kind: UIPlugin
metadata:
  name: ui-logging
spec:
  type: Logging
  logging:
    lokiStack:
      name: lokistack-dev
    frontend:
      timeout: 3m
      logsLimit: 1000
```

pkg/controllers/uiplugin/components.go
```diff
@@ -166,6 +216,33 @@ func newDeployment(info UIPluginInfo, namespace string) *appsv1.Deployment {
 	return plugin
 }
 
+func computeConfigMapHash(cm *corev1.ConfigMap) string {
+	h := sha256.New()
```
simonpasquier, line 220: I'd rather use hash/fnv to make it clear that we need a non-crypto hashing function.

periklis (approved):

xperimental:
Two points of additional information:

- Added an implementation with "separate struct for frontend configuration" (see previous comment) in a [separate branch](https://github.com/xperimental/rhobs-observability-operator/compare/logging-ui...xperimental:rhobs-observability-operator:logging-ui-alternative-api).
- The 5.9 branch of Cluster Logging Operator has the "Cluster Observability Operator awareness" PR merged, so while there is no released version of that yet, [Joao's instructions](https://github.com/rhobs/observability-operator/pull/477#pullrequestreview-2066573688) are simplified to using the `release-5.9` branch of CLO for testing.

simonpasquier (commented):
Thanks for the perseverance! I'll take a deeper look tomorrow (with local testing) but I don't see any blocker.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,93 @@ type UIPluginList struct {
...
+	// Name of the LokiStack resource.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength:=1
+	Name string `json:"name"`
+}
+
 // UIPluginSpec is the specification for desired state of UIPlugin.
+//
+// +kubebuilder:validation:XValidation:rule="self.type != 'Logging' || has(self.logging)", message="Logging configuration is required if type is Logging"
```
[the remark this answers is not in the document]
xperimental: A testing framework for the api-server-side validations would be great, we'd probably also want that in other operators. So far I have only seen frameworks using a real Kubernetes API, so the validation becomes a "kind-of e2e test" which I would greatly dislike. I'd rather have a framework using the apiserver code to do validations as a unit-test, but I don't know if that already exists. I have not really looked yet, though. :slightly_smiling_face:

simonpasquier (commented):
2 final minor comments which can be addressed in a subsequent PR. Happy to merge this one after you've resolved the conflicts.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,92 @@ type UIPluginList struct {
...
+
+// LoggingConfig contains options for configuring the logging console plugin.
+type LoggingConfig struct {
+	// LokiStack points to the LokiStack instance of which logs should be displayed.
+	// It always references a LokiStack in the "openshift-logging" namespace.
+	//
+	// +kubebuilder:validation:Required
+	LokiStack LokiStackReference `json:"lokiStack"`
+
+	// LogsLimit is the max number of entries returned for a query.
```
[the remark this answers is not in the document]
xperimental: I think the `logging-view-plugin` defines its own default, which is used when no limit is specified (= "when it is 0"). If I am interpreting this correctly there's a default of [100 for normal queries](https://github.com/openshift/logging-view-plugin/blob/1b4040b9e423e00a7b35b3a840f132ffd6e73c0b/web/src/loki-client.ts#L114) and [200 for tail](https://github.com/openshift/logging-view-plugin/blob/1b4040b9e423e00a7b35b3a840f132ffd6e73c0b/web/src/loki-client.ts#L170).

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +32,92 @@ type UIPluginList struct {
...
+
+	// TypeLogging deploys the Logging View Plugin for OpenShift Console.
+	TypeLogging UIPluginType = "Logging"
 )
 
+// DeploymentConfig contains options allowing the customization of the deployment hosting the UI Plugin.
+type DeploymentConfig struct {
+	// Define a label-selector for nodes which the Pods should be scheduled on.
+	//
+	// When no selector is specified it will default to a value only selecting Linux nodes ("kubernetes.io/os=linux").
```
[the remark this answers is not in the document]
xperimental: The [code](https://github.com/xperimental/rhobs-observability-operator/blob/87f34746ed1dbc05b8d611a3c9e1ce181eb935a8/pkg/controllers/uiplugin/components.go#L255-L264) supports defaulting the node selector for both the case of having "no deployment configuration" and "not having `nodeSelector` set".

xperimental:
Rebase completed :white_check_mark: 

Any feedback on the question whether we should put the options for the frontend in a separate struct?

simonpasquier (approved):
/lgtm

PR #480, opened 2024-05-09
COO-169: feat: add troubleshooting panel and distributed tracing ui-plugin

PeterYurkovich:
[COO-64](https://issues.redhat.com/browse/COO-64)
[COO-169](https://issues.redhat.com/browse/COO-169)
[OU-378](https://issues.redhat.com/browse/OU-378)
[OU-314](https://issues.redhat.com/browse/OU-378)
[OU-423](https://issues.redhat.com/browse/OU-423)

This PR adds the troubleshooting panel console plugin and distributed tracing console plugin to the observability operator.


pkg/controllers/uiplugin/controller.go
```diff
@@ -251,7 +251,8 @@ func (rm resourceManager) registerPluginWithConsole(ctx context.Context, pluginI
 			OperatorSpec: operatorv1.OperatorSpec{
 				ManagementState: operatorv1.Managed,
 			},
-			Plugins: clusterPlugins,
+			Plugins:       clusterPlugins,
+			Customization: *cluster.Spec.Customization.DeepCopy(),
```
jgbernalp, line 255: I wonder if this is really needed, this is a last resource path we are using and we should not change other things unrelated to plugins

pkg/controllers/uiplugin/components.go
```diff
@@ -49,12 +52,52 @@ func newServiceAccount(info UIPluginInfo, namespace string) *corev1.ServiceAccou
 	}
 }
 
-func newRole(info UIPluginInfo) *rbacv1.Role {
-	return info.Role
+func newRole(info UIPluginInfo, roleName string) *rbacv1.Role {
+	if info.Role != nil {
+		return info.Role
+	}
+	return &rbacv1.Role{
```
jgbernalp, line 59: maybe we don't need this as we would be creating a role with no rules.

pkg/controllers/uiplugin/components.go
```diff
@@ -49,12 +52,52 @@ func newServiceAccount(info UIPluginInfo, namespace string) *corev1.ServiceAccou
...
+	}
 }
 
-func newRoleBinding(info UIPluginInfo) *rbacv1.RoleBinding {
-	return info.RoleBinding
+func newRoleBinding(info UIPluginInfo, roleName string, roleBindingName string, serviceAccountName string) *rbacv1.RoleBinding {
+	if info.RoleBinding != nil {
+		return info.RoleBinding
+	}
+	return &rbacv1.RoleBinding{
```
jgbernalp, line 80: same here.

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -102,6 +101,30 @@ func PluginInfoBuilder(plugin *uiv1alpha1.UIPlugin, pluginConf UIPluginsConfigur
 				},
 			}
 
+			return pluginInfo, nil
+		}
+	case uiv1alpha1.TypeTroubleshootingPanel:
+		{
```
jgbernalp, line 107: Maybe a better approach would be to follow [this](https://github.com/rhobs/observability-operator/pull/477/files#diff-8fe461b7b9b03ac1fddf5f028a5e339ab5fc6414582abee75fdff05fad6d77aa) pattern and extract this into its own file for clarity. We might also need to add support so the korrel8r endpoint and namespace can be configured

pkg/controllers/uiplugin/components.go
xperimental: I noticed that the code using `NewOptionalUpdater` does not work during my work on the logging-view-plugin as well, because it operates on wrong assumptions: the `NewOptionalUpdater` function does not work with a `nil` resource passed to it, so the current code will crash, when used with a plugin without a role or rolebinding.

This code here effectively creates an empty Role (or RoleBinding below) just so that the "optional updater" can remove the resource should it be present in the cluster.

I'm not sure if we need this complexity here though, because the role/rolebinding resources for the current set of plugins are either always or never created, depending on the plugin type. That's why I removed the `NewOptionalUpdater` in [my version of the fix](https://github.com/xperimental/rhobs-observability-operator/commit/73b2e65d79726c003f4ec86769db419a72a0e088) completely.

PeterYurkovich: I had taken a look at your PR and I think you have laid the structure out better than I had in some places. I'm in the process of adding the distributed tracing plugin to this PR as well, but in the process I'm going to be moving the definitions out of the builder and into separate files and making the Role and RoleBinding optional like you have. I'll try to keep the naming and formatting the same as yours to keep any conflicts to a minimum

pkg/controllers/uiplugin/controller.go
PeterYurkovich: I had found that the `spec.customization.customLogoFile.key` was being unset when the Patch method was called in [reconciler.go](https://github.com/rhobs/observability-operator/blob/main/pkg/reconciler/reconciler.go) for the Merger.Reconcile. This leads the Console Operator to get stuck in a reconciliation loop. This issue will pop up on all consoles using a custom image like ROSA. All this line does is ensure that the customization is available on both sides of the patch.

However, although this only happened with the customization and I checked that there wasn't some other item that was getting removed in cluster yaml, it's possible it could happen with other items. I'm not exactly certain how to go about debugging this, but I'd be happy to if someone could point me in the right direction. Otherwise I can remove it, but the UIPlugins won't deploy on ROSA and will cause the Console Operator to get stuck in a reconciliation loop.

pkg/controllers/uiplugin/components.go
xperimental: I was tempted to restructure some of the existing code during the implementation for the logging-view-plugin as well, but I refrained from it for now. Getting it to work was the first priority. But I might follow-up with some of the ideas in another PR, if I still remember them... :slightly_smiling_face:

pkg/controllers/uiplugin/plugin_info_builder.go
PeterYurkovich: I agree, the troubleshooting panel and distributed tracing have been move out into their own files. Korrel8r name and namespace support for the proxy has also been added

pkg/controllers/uiplugin/components.go
```diff
@@ -39,6 +39,26 @@ func pluginComponentReconcilers(plugin *uiv1alpha1.UIPlugin, pluginInfo UIPlugin
...
+
+	if len(pluginInfo.ClusterRoles) > 0 {
+		for _, role := range pluginInfo.ClusterRoles {
+			if role != nil {
+				components = append(components, reconciler.NewUpdater(role, plugin))
+			}
+		}
+	}
+
+	if len(pluginInfo.ClusterRoleBindings) > 0 {
```
tremes, line 54: This check is not needed. Same above.

PeterYurkovich: Removed in most recent commit

danielmellado:
This overall lgtm, however I'd like for @jgbernalp and @periklis to take a look at it as well.

cmd/operator/main.go
```diff
@@ -36,10 +36,12 @@ import (
...
-	"prometheus":    "",
-	"alertmanager":  "",
-	"thanos":        obopo.DefaultThanosImage,
-	"ui-dashboards": "quay.io/openshift-observability-ui/console-dashboards-plugin:v0.1.0",
+	"prometheus":               "",
+	"alertmanager":             "",
+	"thanos":                   obopo.DefaultThanosImage,
+	"ui-dashboards":            "quay.io/openshift-observability-ui/console-dashboards-plugin:v0.1.0",
+	"ui-troubleshooting-panel": "quay.io/openshift-observability-ui/troubleshooting-panel-console-plugin:v0.1.0",
+	"ui-distributed-tracing":   "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v0.1.0",
```
tremes, line 44: I am getting the following for this image:
```
Failed to pull image "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v0.1.0": reading manifest v0.1.0 in quay.io/openshift-observability-ui/distributed-tracing-console-plugin: unauthorized: access to the requested resource is not authorized
```

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,73 @@ type UIPluginList struct {
...
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// DistributedTracingConfig contains options for configuring the Distributed Tracing plugin
+type DistributedTracingConfig struct {
+	// Timeout is the max duration before a query timeout
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
```
tremes, line 81: I am not very good in regular expressions, but this allows me to create e.g:
```
spec:
  distributedTracing:
    timeout: '5'
```
so perhaps `^([0-9]+)([smhd]{1})$` ?

cmd/operator/main.go
jgbernalp: I fixed the permissions in quay

danielmellado:
@tremes @simonpasquier this lgtm now but I'd like another set of eyes, holding until then!

docs/user-guides/observability-ui-plugins.md
```diff
@@ -20,3 +22,33 @@ metadata:
 spec:
   type: Dashboards
 ```
+
+### Troubleshooting Panel
+
+The plugin will connect to a Korrel8r instance named `korrel8r` in the `korrel8r` namespace. A "Troubleshooting Panel" button is added to the alerts page, which will convert the current alert into a Korrel8r query, then retrieve related neighbors and display them in a topology view.
```
periklis, line 28: This is not accurate but can be changed with #497 by @shwetaap

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,79 @@ type UIPluginList struct {
...
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([sm]{1})$"
+	Timeout string `json:"timeout,omitempty"`
+	// korrel8r defines the Korrel8r instance that the troubleshooting panel plugin will connect to
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Korrel8r Instance"
+	Korrel8r TroubleshootingPanelKorrel8rConfig `json:"korrel8r,omitempty"`
```
periklis, line 61: Again not accurate when #497 lands because we can eliminate this field and the struct below. (cc @shwetaap )

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,79 @@ type UIPluginList struct {
...
+	// Timeout is the maximum duration before a query timeout.
+	//
+	// The value is expected to be a sequence of digits followed by a unit suffix, which can be 's' (seconds)
+	// or 'm' (minutes).
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([sm]{1})$"
+	Timeout string `json:"timeout,omitempty"`
+}
```
periklis, line 89: How is the distributed tracing panel connecting to a TempoStack instance? Do we need something like @xperimental created for the logging-ui plugin [here](https://github.com/rhobs/observability-operator/pull/477/files#diff-16e5f13859d1d0daf88ef24d58309d2a62d1768a556ca142b9482722e50a68d5R67-R69), i.e:

```go
// DistributedTracingConfig contains options for configuring the Distributed Tracing plugin
type DistributedTracingConfig struct {
	// TempoStack points to the TempoStack instance of which logs should be displayed.
	//
	TempoStack TempoStackReference `json:"tempoStack"`

	// Timeout is the maximum duration before a query timeout.
	//
	// The value is expected to be a sequence of digits followed by a unit suffix, which can be 's' (seconds)
	// or 'm' (minutes).
	//
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OCP Console Query Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:ocpConsoleTimeout"}
	// +kubebuilder:validation:Pattern:="^([0-9]+)([sm]{1})$"
	Timeout string `json:"timeout,omitempty"`
}


// TempoStackReference is used to configure a reference to a TempoStack that should be used
// by the Distributed Tracing console plugin.
//
// +structType=atomic
type TempoStackReference struct {
	// Name of the TempoStack resource.
	//
	// +kubebuilder:validation:MinLength:=1
	Name string `json:"name"`

	// Namespace of the TempoStack resource.
	//
	// +kubebuilder:validation:MinLength:=1
	Namespace string `json:"namespace"`
}
```

pkg/controllers/uiplugin/compatibility_matrix_test.go
```diff
@@ -34,6 +34,44 @@ func TestCompatibilityMatrixSpec(t *testing.T) {
...
+		},
+		{
+			pluginType:     uiv1alpha1.TypeDistributedTracing,
+			clusterVersion: "4.11",
+			expectedKey:    "ui-troubleshooting-panel",
+			expectedErr:    nil,
+		},
+		{
+			pluginType:     uiv1alpha1.TypeDistributedTracing,
+			clusterVersion: "4.24.0-0.nightly-2024-03-11-200348",
```
periklis, line 71: Is this cluster version a typo or accidental mistake when testing locally?

periklis (changes requested):

pkg/apis/uiplugin/v1alpha1/types.go
jgbernalp: There is no single TempoStack instance. We [list](https://github.com/openshift/distributed-tracing-console-plugin/blob/main/pkg/server.go#L96) the available TempoStack instances so the frontend can choose one to connect.

periklis: Ah good to know, didn't have that in my mind when reviewing the code.

periklis (approved):

tremes (approved):

danielmellado (approved):

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,79 @@ type UIPluginList struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Korrel8r Instance"
+	Korrel8r TroubleshootingPanelKorrel8rConfig `json:"korrel8r,omitempty"`
+}
+
+type TroubleshootingPanelKorrel8rConfig struct {
+	// Name of the korrel8r instance
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Korrel8r Instance Name"
+	Name string `json:"name,omitempty"`
```
simonpasquier, line 69: Should it really be optional?

Reading the code we assume a default value of "korrel8r" for both but I'd rather make these fields mandatory to be explicit.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,79 @@ type UIPluginList struct {
...
 type UIPluginSpec struct {
 	// Type defines the UI plugin.
 	// +required
 	// +kubebuilder:validation:Required
 	Type UIPluginType `json:"type"`
+
+	// TroubleshootingPanel contains configuration for the troubleshooting console plugin.
+	//
+	// +kubebuilder:validation:Optional
+	TroubleshootingPanel *TroubleshootingPanelConfig `json:"troubleshootingPanel,omitempty"`
```
simonpasquier, line 101: Similar to what's being done in #477 we need some additional validation to ensure that when type != TroubleshootingPanel, this field isn't defined.
And if we make TroubleshootingPanelKorrel8rConfig a mandatory field, it should be validated too.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -31,20 +31,79 @@ type UIPluginList struct {
...
+
+	// DistributedTracing contains configuration for the distributed tracing console plugin.
+	//
+	// +kubebuilder:validation:Optional
+	DistributedTracing *DistributedTracingConfig `json:"distributedTracing,omitempty"`
```
simonpasquier, line 106: same remark here, we need to check the type to enforce that the field isn't defined when type != DistributedTracing

simonpasquier (commented):
Sorry for the late review. My comments weren't meant to be blocking but it'd be good to address sooner than later.

pkg/apis/uiplugin/v1alpha1/types.go
PeterYurkovich: The korrel8r deployment is being attached to the troubleshooting panel UIPlugin in #497 so this specific comment might be helpful to address in that PR.

pkg/apis/uiplugin/v1alpha1/types.go
PeterYurkovich: I've created a [jira](https://issues.redhat.com/browse/COO-178) for visibility and will make a follow up PR to address these.

pkg/apis/uiplugin/v1alpha1/types.go
PeterYurkovich: I've created a [jira](https://issues.redhat.com/browse/COO-178) for visibility and will make a follow up PR to address these.

docs/user-guides/observability-ui-plugins.md
shwetaap: Changed it in the PR https://github.com/rhobs/observability-operator/pull/497

pkg/apis/uiplugin/v1alpha1/types.go
shwetaap: Removed this is PR https://github.com/rhobs/observability-operator/pull/497

pkg/apis/uiplugin/v1alpha1/types.go
shwetaap: "Korrel8r" will always be deployed  when the Troubleshooting Panel is installed in the same namespace where the observability-operator is installed. So there is no need for any configuration.

shwetaap: Removed it in https://github.com/rhobs/observability-operator/pull/497

PR #481, opened 2024-05-14
feat: restrict UIPlugin CRD names to allow a single instance per type

jgbernalp:
[OU-402](https://issues.redhat.com/browse/OU-402)
[COO-164](https://issues.redhat.com/browse/COO-164)

This PR adds a pattern restriction for the UIPlugin name so only a single instance of a specific plugin is allowed

### DIFF
--- a/bundle/manifests/observability.openshift.io_uiplugins.yaml
+++ b/bundle/manifests/observability.openshift.io_uiplugins.yaml
@@ -251,6 +251,16 @@ spec:
             - conditions
             type: object
         type: object
+        x-kubernetes-validations:
+        - message: UIPlugin name must be 'logging' if type is Logging
+          rule: self.spec.type != 'Logging' || self.metadata.name == 'logging'
+        - message: UIPlugin name must be 'troubleshooting-panel' if type is TroubleshootingPanel
+          rule: self.spec.type != 'TroubleshootingPanel' || self.metadata.name ==
+            'troubleshooting-panel'
+        - message: UIPlugin name must be 'distributed-tracing' if type is DistributedTracing
+          rule: self.spec.type != 'DistributedTracing' || self.metadata.name == 'distributed-tracing'
+        - message: UIPlugin name must be 'dashboards' if type is Dashboards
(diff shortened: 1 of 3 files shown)


jan--f:
/lgtm

Nice! We'll have to chat about OCP specific integration tests soon.

simonpasquier:
AFAIU you can still create conflicting objects if you mess up the names and the types, right? For instance:

````
kind: UIPlugin
metadata:
  name: ui-dashboards
spec:
  type: Dashboards
---
kind: UIPlugin
metadata:
  name: ui-logging
spec:
  type: Dashboards
````

I don't think that it's possible to validate this case with OpenAPI but we might want the operator to stop reconciliation when it detects the issue?

**Comment by jan--f** [comment]
Ahh I think I might have been (or still am) confused what needs to be done. So there can always only be a single `dashboard` UI plugin? An for logging a new type is added, going by #477?

In that case can we not tie the name to the type? Something like
```
// +kubebuilder:validation:XValidation:rule="self.name == self.spec.type",message="Name is invalid"
```

simonpasquier:
@jan--f this would be neat if it works indeed.

jgbernalp:
Will do some tests based on @jan--f suggestion and update the PR

xperimental:
Just a nitpick, but I think I would get rid of the `ui-` prefix for the names. First it mirrors the name of the resource type (`UIPlugin`) and it also currently causes the deployments to be named `observability-ui-ui-*` (which could be changed in the operator though).

jgbernalp:
/hold

@csibbitt please check if changes are needed for STF

simonpasquier (commented):
/lgtm
/hold

:+1:

csibbitt:
> /hold
> 
> @csibbitt please check if changes are needed for STF

STF does not use the UIPlugin, but Telemetry Operator does. I'll take care of this ASAP on  OSPRH-7600

jgbernalp:
since the change for the telemetry operator was merged https://github.com/openstack-k8s-operators/telemetry-operator/pull/406, we can proceed with this.

/unhold

PR #482, opened 2024-05-14
ci: rename commitlint.config.js to commitlint.config.mjs

jan--f:
This fixes commitlint failures caused by
https://github.com/wagoid/commitlint-github-action/pull/763.


commitlint.config.mjs
danielmellado, line 1: We would need to change module.exports to `export default Configuration;` in here to comply with the ES module syntax.

jan--f: Like so?

danielmellado: yeah, I think that should make it.

commitlint.config.mjs
```diff
@@ -9,4 +9,4 @@ const Configuration = {
   ignores: [(msg) => /Signed-off-by: dependabot\[bot]/m.test(msg)],
 };
 
-module.exports = Configuration;
+export default Configuration
```
danielmellado, line 12:
```suggestion
export default Configuration;
```

danielmellado: let's keep the end trailing even if it does seem that it works without it.

Issue #200, opened 2022-10-12
Reconciler fails to Patch at random


sthaha:
Related to #181 . 

We sometimes also see that reconciler fails to `Patch`. The issue seem to be that the `Patch` tries to update a non-existent resource when it should actually be creating it. 

E.g. - https://github.com/rhobs/observability-operator/actions/runs/3232742256/jobs/5294131669#step:4:989 
```
2022-10-12T07:58:40Z	INFO	observability-operator	Reconciling monitoring stack	{"stack": "e2e-tests/valid-loglevel"}
2022-10-12T07:58:40Z	INFO	thanos-querier	watched MonitoringStack changed, checking for matching querier	{"Monitoring Stack": "e2e-tests/valid-loglevel"}
2022-10-12T07:58:41Z	INFO	observability-operator	patch failed	{"stack": "e2e-tests/valid-loglevel", "reconciler": "updater", "name": "valid-loglevel-alertmanager", "ns": "e2e-tests", "error": "roles.rbac.authorization.k8s.io \"valid-loglevel-alertmanager\" not found"}
2022-10-12T07:58:41Z	INFO	observability-operator	Failed to get prometheus object	{"stack": "e2e-tests/valid-loglevel", "err": "Prometheus.monitoring.rhobs \"valid-loglevel\" not found"}
2022-10-12T07:58:41Z	ERROR	Reconciler error	{"controller": "monitoringstack", "controllerGroup": "monitoring.rhobs", "controllerKind": "MonitoringStack", "monitoringStack": {"name":"valid-loglevel","namespace":"e2e-tests"}, "namespace": "e2e-tests", "name": "valid-loglevel", "reconcileID": "45436353-5cee-45b0-947f-aefeed0e2d0a", "error": "roles.rbac.authorization.k8s.io \"valid-loglevel-alertmanager\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234
2022-10-12T07:58:41Z	INFO	observability-operator	Reconciling monitoring stack	{"stack": "e2e-tests/valid-loglevel"}
```

```
[...]

PR #492, opened 2024-05-28
feat: TLS support for Alertmanager and Prometheus web endpoints

vyzigold:
This PR adds a new "WebTLSConfig" field to the PrometheusConfig and AlertmanagerConfig sections of the MonitoringStack CRD, which allows the user to specify secrets containing TLS certificates. The Prometheus and Alertmanager objects are configured to use these secrets. The feature is tested in a new "assertPrometheusScrapesItselfTLS" testcase.

I needed to add the secrets containing CA certificates to the "secrets" field in the Prometheus object in order to use them for the self scraping jobs. This means the "secrets" field is now managed by OBO, so I added a new "secrets" field to the MonitoringStack CRD, which gives the users the ability to specify additional secrets to mount into the prometheus container. Specifying additional secrets is required if the scrape endpoints are also behind TLS with custom CA certs.


pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -215,11 +243,48 @@ func newPrometheus(
...
+						Secret: &corev1.SecretKeySelector{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: alertmanagerTLSSecretName,
+							},
+							Key: TLSCaKey,
+						},
+					},
+					ServerName: ms.Name + "-alertmanager",
+				},
+			}
```
vyzigold, line 266: Is this a known bug? I didn't find anything about it. Is this happening only in the forked operator or is it the same upstream?

simonpasquier: can you try to repro with the upstream operator?

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +278,25 @@ type ThanosQuerierSpec struct {
...
+}
+
+// WebTLSConfig contains configuration to enable TLS on web endpoints
+type WebTLSConfig struct {
+	// Secret containing the TLS key for the server
+	Key SecretKeySelector `json:"key"`
+	// Secret containing the TLS certificate for the server
+	Cert SecretKeySelector `json:"cert"`
+	// Secret containing the CA certificate to verify the server certificates
+	CA SecretKeySelector `json:"ca"`
```
simonpasquier, line 301: Do we need this for Alertmanager?

simonpasquier (commented):
It's be easier for reviewers if you split the PR: 1 for Alertmanager and 1 for Prometheus.

pkg/apis/monitoring/v1alpha1/types.go
vyzigold: This depends. Do we want to support the case, where the TLS certificate for prometheus and the certificate for alertmanager is signed by a different CA? I don't see a reason why somebody would do that, but theoretically there is nothing wrong on such a configuration. But in that case, if we want to have the self-scraping jobs being able to scrape prometheus and alertmanager, we would need to know the CA cert of both.

If we decide, that we require the user to use a single CA for the prometheus and alertmanager web endpoints, then we would need the CA field only once.

pkg/controllers/monitoring/monitoring-stack/components.go
vyzigold: I'm adding this to my personal TODO list and if I see the same behavior with the upstream operator, I'll file a bug. In the meantime I'll remove this commented section, I left it there just to show you what isn't working.

test/e2e/monitoring_stack_controller_test.go
```diff
@@ -647,6 +650,287 @@ func assertPrometheusManagedFields(t *testing.T) {
 	assert.DeepEqual(t, have, expected)
 }
 
+func assertPrometheusScrapesItselfTLS(t *testing.T) {
+	promKey := `
```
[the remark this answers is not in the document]
vyzigold: Good suggestion, I'll do that.

pkg/apis/monitoring/v1alpha1/types.go
simonpasquier: Yes, makes sense. Sorry for the noise.

vyzigold:
> It's be easier for reviewers if you split the PR: 1 for Alertmanager and 1 for Prometheus.

I removed the alertmanager bits from this PR and I'll create another PR with them shortly. This PR now includes only the Prometheus TLS implementation. I renamed the PR and rewrote the description accordingly.

jan--f:
Thanks. I have only one comment regarding the Secrets api addition. Since OBO uses server-side-apply user should still be able to add secrets to that list by also user server-side-apply.
Though its possible this requires an annotation in the Prometheus type. @simonpasquier I seem to remember we discussed this before.

If possible I'd like to avoid adding Secrets to the MonitoringStack API.

simonpasquier:
Regarding the `.spec.secrets` field in Prometheus CRD, it should be possible to modify upstream to validate the field as a set because specifying `[foo, foo]` breaks the statefulset.
Having said that, OBO could also read the secret's value and use `ca` instead of `ca_file` (at least until upstream properly supports SSA).

vyzigold:
I don't think that using `ca` instead of `ca_file` would be entirely possible everywhere.  I think there are 2 places, where we need to use `ca_file`. The first is the the "additional scrape configs".  If I understand that correctly, that is basically just copied into the Prometheus configuration without modification. I'd need to look into it, but I don't think the prometheus-operator interprets this to mount CAs. The second place is the prometheus.Spec.Alerting.Alertmanagers in https://github.com/rhobs/observability-operator/pull/495 as using `ca` didn't seem to work (I still need to try replicating the possible bug with the upstream prometheus-operator). See https://github.com/rhobs/observability-operator/compare/adc714f4792654978f02899429e05c4e26a404ef..9f1e010a60ecf4de6935889e4bcac3e462d59440#diff-1b87d7169ef5b2a3785bd99c951d995909da935035a924781c1f0ddd12369fd2L250

Regarding the SSA. I don't understand the comment about what's exactly missing upstream. Does it mean, that SSA won't work for .spec.secrets at the moment? I'll try and see if I can make it work, but I think I'll be pretty busy for the next 2 weeks unfortunately, so it might take me a little bit more time to get back to this.

jan--f:
Thanks for your work on this, we much appreciate it! No worries if other priorities come up.

The SSA comment refers to a missing kubebuilder annotation upstream. The secrets field needs to set the correct `// +listType` so that the API server validate it propoerly in the SSA context. Right now there can only be one owner of `.spec.secrets`. But with the right annotation each item of it can have a different owner, i.e. OBO can set the secret it needs, but we can leave it to users to add their own, in turn saving us from adding an API field that is merely an additional layer so that we can inject the secret we want in their. I.e. just less code for us to maintain.

Regarding your quesiton about `ca` vs `ca_file` I'm not sure. Maybe @slashpai can shed more light whether prom-op can mount secrets referenced in `additional scrape configs?

vyzigold:
I rebased on top of main, I hopefully fixed all the CI complaints. I removed the commit, which added the Secret to the api. At the moment the code works, but OBO takes ownership of the Prometheus's secret field and users aren't able to use SSA to add new secrets. As you mentioned in previous comments, we'll need some upstream changes for SSA to work for the secrets.

pkg/controllers/monitoring/monitoring-stack/components.go
vyzigold: I replicated the issue with the upstream operator from main branch and created an issue: https://github.com/prometheus-operator/prometheus-operator/issues/6760

vyzigold:
I went ahead and created an upstream PR to add the +listType. I tested it together with this PR and it seems to work quite nicely. OBO would add whatever it needed to the secrets field and I was able to use SSA to add other secrets. OBO was able to remove the TLS secrets when I disabled TLS while leaving the manually added secrets intact and so on.

PR: https://github.com/prometheus-operator/prometheus-operator/pull/6762

jan--f:
Nice I think this
/lgtm
I suppose we'll have the po change in a release before our next point release, but will confirm that before merging.
Apologies about the repeated conflicts, the CSV tracking `generatedAt` is rather annoying when generating and tracking .

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -192,6 +192,9 @@ type PrometheusConfig struct {
 	// Default interval between scrapes.
 	// +optional
 	ScrapeInterval *monv1.Duration `json:"scrapeInterval,omitempty"`
+	// Configure TLS options for the web endpoint
```
simonpasquier, line 195:
```suggestion
	// Configure TLS options for the Prometheus web server.
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	Name string `json:"name"`
+	// The key of the secret to select from.  Must be a valid secret key.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	Key string `json:"key"`
+}
+
+// WebTLSConfig contains configuration to enable TLS on web endpoints
+type WebTLSConfig struct {
+	// Secret containing the TLS key for the server
```
simonpasquier, line 290:
```suggestion
	// Reference to the TLS private key for the web server.
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	Key SecretKeySelector `json:"key"`
+	// Secret containing the TLS certificate for the server
```
simonpasquier, line 292:
```suggestion
	// Reference to the TLS public certificate for the web server.
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	// The key of the secret to select from.  Must be a valid secret key.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	Key string `json:"key"`
+}
+
+// WebTLSConfig contains configuration to enable TLS on web endpoints
+type WebTLSConfig struct {
+	// Secret containing the TLS key for the server
+	Key SecretKeySelector `json:"key"`
```
simonpasquier, line 291:
```suggestion
	// +kubebuilder:validation:Required
	PrivateKey SecretKeySelector `json:"privateKey"`
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	// Secret containing the TLS certificate for the server
+	Cert SecretKeySelector `json:"cert"`
```
simonpasquier, line 293:
```suggestion
	// +kubebuilder:validation:Required
	Certificate SecretKeySelector `json:"certificate"`
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	// Secret containing the CA certificate to verify the server certificates
```
simonpasquier, line 294:
```suggestion
	// Reference to the root Certificate Authority used to verify the web server's certificate.
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
...
+	CA SecretKeySelector `json:"ca"`
```
simonpasquier, line 295: (nit) more verbosity is better here IMHO.
Also do we make it required?
```suggestion
	// +kubebuilder:validation:Required
	CertificateAuthority SecretKeySelector `json:"certificateAuthority"`
```

pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -269,3 +272,25 @@ type ThanosQuerierSpec struct {
 // It should always be reconstructable from the state of the cluster and/or outside world.
 type ThanosQuerierStatus struct {
 }
+
+// SecretKeySelector selects a key of a secret
```
simonpasquier, line 276:
```suggestion
// SecretKeySelector selects a key of a secret.
```

pkg/apis/monitoring/v1alpha1/types.go
vyzigold: Pretty sure required is the default, but sure lets be explicit just in case. Thank you for all the other suggestions. They make sense, I'll include them in the PR.

vyzigold:
@simonpasquier thank you for your suggestions. I rebased on top of main, included Simon's suggestions and regenerated the manifests / bundle.

I'll do the same for the other 3 PRs

vyzigold:
Looks like the managed fields test was only half complete, for some reason I forgot to fully complete it. I switched it to deploy with TLS and added the expected fields to the expected json. I still have a difference between the expected managed fields and the managed fields from the test when running it locally.
```
    monitoring_stack_controller_test.go:703: assertion failed: 
        --- have
        +++ expected
          map[string]any{
          	... // 2 identical entries
          	"f:alerting":                    map[string]any{"f:alertmanagers": map[string]any{}},
          	"f:arbitraryFSAccessThroughSMs": map[string]any{},
        - 	"f:enableFeatures":              map[string]any{},
        + 	"f:enableFeatures":              map[string]any{`v:"otlp-write-receiver"`: map[string]any{}},
          	"f:enableRemoteWriteReceiver":   map[string]any{},
          	"f:externalLabels":              map[string]any{"f:key": map[string]any{}},
        + 	"f:image":                       map[string]any{},
          	"f:logLevel":                    map[string]any{},
          	"f:podMetadata":                 map[string]any{"f:labels": map[string]any{"f:app.kubernetes.io/component": map[string]any{}, "f:app.kubernetes.io/part-of": map[string]any{}}},
          	... // 23 identical entries
          }
```
But I don't see a reason why. AFAIK I didn't do anything with image and enabledFeatures. Maybe it's because I use crc for development instead of kind?

vyzigold:
I deployed in kind and rerun the tests. I can confirm, that the CI was green except for post-e2e, which I think doesn't matter until #200 is fixed. https://paste.opendev.org/show/bYHXIta79uQXR4q0Y9pU/

jan--f (approved):
Thanks, this
/lgtm

PR #495, opened 2024-05-31
feat: TLS support for the Alertmanager web endpoint

vyzigold:
This PR adds a new "WebTLSConfig" field to the AlertmanagerConfig section of the MonitoringStack CRD, which allows the user to specify secrets containing TLS certificates. The Alertmanager object is configured to use these secrets. The Prometheus object is configured to reach Alertmanager behind TLS. The self-scraping job to scrape Alertmanager is configured correctly too. The feature is tested in a new "assertAlertmanagerReceivesAlertTLS" testcase.

The first commit of the PR (which adds the WebTLSConfig struct) is shared with https://github.com/rhobs/observability-operator/pull/492


jan--f (commented):
~Thanks. I have only one comment regarding the `Secrets` api addition. Since OBO uses server-side-apply user should still be able to add secrets to that list by also user server-side-apply.
Though its possible this requires an annotation in the Prometheus type. @simonpasquier I seem to remember we discussed this before.~

~If possible I'd like to avoid adding `Secrets` to the MonitoringStack API.~
Sorry that was the wrong PR.

/lgtm
/hold
Just holding for more eyes.

jan--f:
@vyzigold can you run `make generate` and make the commit linter happy. Itwants commit messages to start lower case. The commit with the tests also has a typo :)

vyzigold:
@jan--f I fixed the commit messages and I run `make generate`, but it didn't produce any change. I don't have the time today to do this for the other PRs or to take a look at the CI results and I'm gone for the rest of the week. The plan is to fully get back to this work by early next week (Monday or Tuesday).

vyzigold:
I rebased on top of main, executed `make generate bundle` and hopefully fixed all the other CI issues. Tomorrow I'll go through the other PRs.

vyzigold:
An important note here. Similarly to the Prometheus PR. This PR right now also takes ownership of the Prometheus's secrets field.

jan--f:
Thanks!
/lgtm

jan--f:
merging this, the test failure is unrelated to this change.

PR #496, opened 2024-05-31
feat: TLS support for the Thanos web endpoint

vyzigold:
This PR adds a new "WebTLSConfig" field to the ThanosQuerier Spec, which allows the user to specify secrets containing TLS certificates. The ThanosQuerier object is configured to use these secrets.

The secrets are then watched for changes and a reconciliation is triggered when one of them changes. In order to know if the contents of the TLS related secrets changed, a hash is computed from the contents of the secrets. The computed hash value is added to the querier deployment as an env variable. This means, that when that variable changes, a new querier pod is automatically created, which will use the new content of the secrets. (when the content of the secrets is the same, the old pod stays running).

The ServiceMonitor for monitoring the Querier is also modified to use TLS when TLS is enabled.

The first 2 commits of the PR (which add the WebTLSConfig struct and the TLS prometheus client for testing) are shared with https://github.com/rhobs/observability-operator/pull/492


pkg/controllers/monitoring/thanos-querier/components.go
```diff
@@ -65,6 +95,12 @@ func newThanosQuerierDeployment(name string, spec *msoapi.ThanosQuerier, sidecar
 					Containers: []corev1.Container{
 						{
 							Name:  "thanos-querier",
+							Env: []corev1.EnvVar{
+								{
+									Name: "TLS_CONFIG_HASH",
+									Value: tlsHash,
+								},
+							},
```
jan--f, line 103: In CMO we use the annotations for the same purpose. Naming is along the lines of `monitoring.openshift.io/<secre-name>-hash`. Is there any reason to prefer an env var here?

I think the annotation has two advantages: A more explicit name. We could use `monitoring.rhobs/<secret-name>-hash`. And future readers will spend no time wondering what thanos might do with the mysterious env var :)

jan--f (commented):
This looks very nice! I have one small question though.

pkg/controllers/monitoring/thanos-querier/components.go
vyzigold: I was wondering how the rest of the world did this :smile: . There is no particular reason to use env vars over annotations. It's just what we use in openstack-k8s-operators, so it's what I'm used to. I'll redo the code to use annotations instead.

vyzigold: I rewrote the code to use annotations as you suggested. I ended up using a bit different name though if that's ok. The format I used is `monitoring.openshift.io/<secret-name>-<secret-key>-hash`.

If we used only `monitoring.openshift.io/<secret-name>-hash`, we would have issues for example when the certificate and the certificate key are in a single secret, but under different keys (which I think will happen most of the time). If we compute the hash only from the keys used for the TLS as it's doing right now and then try to save both into the same annotation, we will overwrite one of them and thus we won't be able to react if the data from the overwritten key changes and the other doesn't.

Other solutions might be:
- Always compute hashes from the whole secret (all of the keys)
- Add a code to determine from which keys to compute the hashes for each secret
- Just name the annotations differently altogether. Like `monitoring.openshift.io/tls-ca-cert-hash`, `monitoring.openshift.io/tls-key-hash` `monitoring.openshift.io/tls-ca-cert-hash`
- Use one annotation for all three values together. The data could be appended and a single hash would be computed.

jan--f:
Thanks!
/lgtm
/approve

PR #474, opened 2024-05-02
Add Korrel8r deployment within COO

shwetaap:
Add Korrel8r deployment within observability operator

PR #497, opened 2024-05-31
 Korrel8r backend creation when the troubleshooting panel is deployed

shwetaap:
Korrel8r backend deployments get created when the troubleshooting panel is created. Uses the same controller as the Troubleshooting panel to setup the backend Korrel8r components. Closing the previous draft PR in favor of this approach - https://github.com/rhobs/observability-operator/pull/474


Makefile.tools
```diff
@@ -188,7 +188,8 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
 		set -ex ;\
 		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
 		cd $$(mktemp -d) ;\
-		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz";\
+		OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
```
alanconway, line 191: Need to rebase, showing many spurious differences with main.

cmd/operator/main.go
```diff
@@ -88,7 +89,7 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
 	flag.Var(images, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", imagesUsed()))
-	flag.BoolVar(&openShiftEnabled, "openshift.enabled", false, "Enable OpenShift specific features such as Console Plugins.")
+	flag.BoolVar(&openShiftEnabled, "openshift.enabled", true, "Enable OpenShift specific features such as Console Plugins.")
```
alanconway, line 92: Is it correct to change the default, or should we assue "--openshift.enabled" will be on the command line?

pkg/controllers/uiplugin/components.go
```diff
@@ -252,10 +268,172 @@ func newService(info UIPluginInfo, namespace string) *corev1.Service {
...
+			},
+			Selector: componentLabels(name),
+			Type:     corev1.ServiceTypeClusterIP,
+		},
+	}
+}
+
+func newKorrel8rConfigMap(name string, namespace string) *corev1.ConfigMap {
+
+	korrel8rData := struct {
```
alanconway, line 397: This is fine but just FYI, a map is also a handy way to feed data to a template:
```
korrel8rData := map[string]string { "Metric": "thanos-querier", ... etc }
```

pkg/controllers/uiplugin/config/korrel8r.yaml
```diff
@@ -0,0 +1,20 @@
+# Default configuration for deploying Korrel8r as a service in an OpenShift cluster.
+# Store service URLs assume that stores are installed in their default locations.
+stores:
+  - domain: k8s
+  - domain: alert
+    metrics: https://{{ .Metric }}.openshift-monitoring.svc:9091
```
alanconway, line 6: Might be good to make the namespaces variables as well. Even if they're fixed today, they might become variables at some point. Not a reason to delay this PR but on the to-do list

alanconway (approved):
Looks good, a few minor comments but none to hold up the PR.

cmd/operator/main.go
shwetaap: Done

pkg/controllers/uiplugin/components.go
shwetaap: Done

pkg/controllers/uiplugin/components.go
```diff
@@ -19,21 +23,59 @@ const (
...
 
-	return []reconciler.Reconciler{
+	components := []reconciler.Reconciler{
 		reconciler.NewUpdater(newServiceAccount(pluginInfo, namespace), plugin),
-		reconciler.NewOptionalUpdater(newRole(pluginInfo), plugin, hasRole),
-		reconciler.NewOptionalUpdater(newRoleBinding(pluginInfo), plugin, hasRoleBinding),
 		reconciler.NewUpdater(newDeployment(pluginInfo, namespace), plugin),
 		reconciler.NewUpdater(newService(pluginInfo, namespace), plugin),
 		reconciler.NewUpdater(newConsolePlugin(pluginInfo, namespace), plugin),
+		reconciler.NewUpdater(newKorrel8rService(kname, namespace), plugin),
```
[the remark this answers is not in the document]
shwetaap: I made the deployment of Korrel8r components conditional. But I thought the components creation can still be in components.go file., since all the resource creation seems to be in there.

danielmellado:
@shwetaap would you mind rebasing this? thanks

pkg/controllers/uiplugin/components.go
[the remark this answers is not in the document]
alanconway: @jgbernalp does this mean that the existing korrel8r links (alerts to logs, logs to metrics) are being dropped, or that they're only available if the troubleshooting panel is present? I've no objection either way, just want to know.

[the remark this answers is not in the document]
jgbernalp: the logging plugin is the one enabling the links if korrel8r is found in the cluster. In this case, the troubleshooting panel will be the trigger to enable korrel8r. So yes, they will be enabled only when the troubleshooting panel is present.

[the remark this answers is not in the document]
alanconway: +1

pkg/controllers/uiplugin/troubleshooting_panel.go
```diff
@@ -99,3 +91,29 @@ func marshalTroubleshootingPanelPluginConfig(cfg *uiv1alpha1.TroubleshootingPane
...
+
+	// Accumulate services that contain "gateway" in their names
+	var gatewayServices []corev1.Service
+	for _, service := range serviceList.Items {
+		if strings.Contains(service.Name, "gateway") {
+			gatewayServices = append(gatewayServices, service)
+		}
+	}
+	key := "app.kubernetes.io/managed-by"
+	if len(gatewayServices) > 0 {
```
tremes, line 110: This check is not necessary

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -109,7 +111,9 @@ func PluginInfoBuilder(plugin *uiv1alpha1.UIPlugin, pluginConf UIPluginsConfigur
 		}
 	case uiv1alpha1.TypeTroubleshootingPanel:
 		{
-			return createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
+			pluginInfo, _ := createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
```
tremes, line 114: The error should not probably be ignored.

pkg/controllers/uiplugin/controller.go
```diff
@@ -158,6 +158,13 @@ func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, err
 	}
 
+	if pluginInfo.Korrel8rImage != "" {
+		var OpenshiftLoggingNs = "openshift-logging"
+		var OpenshiftNetobservNs = "netobserv"
+		pluginInfo.LokiServiceNames[OpenshiftLoggingNs], _ = getLokiServiceName(ctx, rm.k8sClient, OpenshiftLoggingNs)
```
jgbernalp, line 164: Not to block this PR, but maybe we can pass the context and client to the `PluginInfoBuilder` so it can discover the Loki services there

pkg/controllers/uiplugin/config/korrel8r.yaml
danielmellado: Agree, it's an easy thing to do and could be good for the future us.

pkg/controllers/uiplugin/troubleshooting_panel.go
```diff
@@ -29,33 +31,23 @@ func createTroubleshootingPanelPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace
...
 		Proxies: []osv1alpha1.ConsolePluginProxy{
 			{
 				Type:      osv1alpha1.ProxyTypeService,
 				Alias:     "korrel8r",
 				Authorize: false,
 				Service: osv1alpha1.ConsolePluginProxyServiceConfig{
-					Name:      proxyName,
-					Namespace: proxyNamespace,
-					Port:      8443,
+					Name:      "korrel8r",
```
danielmellado, line 48: Just in case the name might change at some point, I'd put this back as a var.

pkg/controllers/uiplugin/components.go
```diff
@@ -2,6 +2,10 @@ package uiplugin
 
 import (
 	"fmt"
+	"embed"
```
danielmellado, line 5: the import list needs to be in alphabetical order or the linter will fail

shwetaap: Done

pkg/controllers/uiplugin/troubleshooting_panel.go
shwetaap: Done

pkg/controllers/uiplugin/plugin_info_builder.go
shwetaap: Done

pkg/controllers/uiplugin/troubleshooting_panel.go
shwetaap: So just return the first service from the list, since loki is the only gateway service expected? Is that what you meant?

pkg/controllers/uiplugin/config/korrel8r.yaml
shwetaap: Done

pkg/controllers/uiplugin/components.go
```diff
@@ -295,6 +316,165 @@ func newService(info UIPluginInfo, namespace string) *corev1.Service {
...
+				},
+			},
+		},
+	}
+	return deploy
+}
+
+func newKorrel8rService(name string, namespace string) *corev1.Service {
+	annotations := map[string]string{
+		"service.alpha.openshift.io/serving-cert-secret-name": name,
```
periklis, line 413:
```suggestion
		"service.beta.openshift.io/serving-cert-secret-name": name,
```

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -108,7 +112,13 @@ func PluginInfoBuilder(plugin *uiv1alpha1.UIPlugin, pluginConf UIPluginsConfigur
 			return pluginInfo, nil
 		}
 	case uiv1alpha1.TypeTroubleshootingPanel:
-		return createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
+		{
+			pluginInfo, err := createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
+			pluginInfo.Korrel8rImage = pluginConf.Images["korrel8r"]
+			pluginInfo.LokiServiceNames[OpenshiftLoggingNs], _ = getLokiServiceName(ctx, k, OpenshiftLoggingNs)
+			pluginInfo.LokiServiceNames[OpenshiftNetobservNs], _ = getLokiServiceName(ctx, k, OpenshiftNetobservNs)
```
periklis, line 119: Don't skip error handling here.

periklis (changes requested):

pkg/controllers/uiplugin/components.go
shwetaap: Done

pkg/controllers/uiplugin/plugin_info_builder.go
shwetaap: Added the error check

alanconway: @periklis can you mark this resolved? The code is fixed, and this is the last blocker.

pkg/controllers/uiplugin/troubleshooting_panel.go
tremes: Ohh sorry I was wrong. I thought there was a for loop so apparently I was looking the wrong way. You can ignore. Sorry.

periklis (approved):

PR #498, opened 2024-06-05
COO-178: fix: prevent other plugin types from using dynamic tracing and troubleshooting panel configurations

PeterYurkovich:
[COO-178](https://issues.redhat.com/browse/COO-178)

This PR looks the address the comments made on #480 about fixing the configuration validations for the distributed tracing and troubleshooting panel console plugins.

These configurations are currently entirely optional for both plugins, so we don't need to enforce the configuration existing if the plugin exists. We only need to prevent other plugins from using the configuration if they don't have the associated type.


pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -149,6 +149,8 @@ type LokiStackReference struct {
 
 // UIPluginSpec is the specification for desired state of UIPlugin.
 //
+// +kubebuilder:validation:XValidation:rule="self.type == 'TroubleshootingPanel' || !has(self.troubleshootingPanel)", message="TroubleshootingPanel type is required to use Troubleshooting Panel configuration"
```
simonpasquier, line 152:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.type == 'TroubleshootingPanel' || !has(self.troubleshootingPanel)", message="Troubleshooting panel configuration is only supported with the TroubleshootingPanel type"
```

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -149,6 +149,8 @@ type LokiStackReference struct {
...
+// +kubebuilder:validation:XValidation:rule="self.type == 'DistributedTracing' || !has(self.distributedTracing)", message="DistributedTracing type is required to use Distributed Tracing configuration"
```
simonpasquier, line 153: same here

pkg/apis/uiplugin/v1alpha1/types.go
PeterYurkovich: Updated the wording of the message of all the TroubleshootingPanel validations

pkg/apis/uiplugin/v1alpha1/types.go
PeterYurkovich: Updated the wording of the message of all the DistributedTracing validations

simonpasquier (commented):
/lgtm
/approve

PR #499, opened 2024-06-05
fix: compatibility matrix unit tests

xperimental:


pkg/controllers/uiplugin/compatibility_matrix_test.go
```diff
@@ -40,7 +40,7 @@ func TestCompatibilityMatrixSpec(t *testing.T) {
 			// to render the "Troubleshooting Panel" button on the alert details page.
 			clusterVersion: "4.15",
 			expectedKey:    "",
-			expectedErr:    fmt.Errorf("no compatible image found for plugin type %s and cluster version %s", uiv1alpha1.TypeTroubleshootingPanel, "4.15"),
+			expectedErr:    fmt.Errorf("no compatible image found for plugin type %s and cluster version %s", uiv1alpha1.TypeTroubleshootingPanel, "v4.15"),
```
tremes, line 43: Just out of curiosity. What's the reason for this `v` prefix? The other error messages here don't use it.

danielmellado: +1 on that question

xperimental: The [semver](https://pkg.go.dev/golang.org/x/mod/semver) library used to do the comparison requires the `v` prefix in front of the version number. [This code](https://github.com/xperimental/rhobs-observability-operator/blob/db0b62f8be16d4ff549abd789c3f26218342590e/pkg/controllers/uiplugin/compatibility_matrix.go#L47-L49) prepends the `v` to the provided version number, so that the library can work with it.

The reason why the prefix does not show up in the [other message](https://github.com/xperimental/rhobs-observability-operator/blob/db0b62f8be16d4ff549abd789c3f26218342590e/pkg/controllers/uiplugin/compatibility_matrix.go#L53) is because it is not using the variable but is a static string. The [other test with the same message](https://github.com/xperimental/rhobs-observability-operator/blob/db0b62f8be16d4ff549abd789c3f26218342590e/pkg/controllers/uiplugin/compatibility_matrix_test.go#L79) also has the prefix, because the message is produced [using the variable](https://github.com/xperimental/rhobs-observability-operator/blob/db0b62f8be16d4ff549abd789c3f26218342590e/pkg/controllers/uiplugin/compatibility_matrix.go#L68).

Using a library which does not require the `v` prefix would be an alternative.

tremes: ohhh I see. Thank you for the explanation...then it probably makes sense, but I guess it might be a bit confusing in future :thinking:

PR #502, opened 2024-06-06
fix: update distributed_tracing.go to include a service proxy

zhuje:
Updates are needed to reflect the changes in PR: https://github.com/openshift/distributed-tracing-console-plugin/pull/15, which allows the distributed-tracing-console-plugin to use this proxy service  at the endpoint : `/api/proxy/plugin/distributed-tracing-console-plugin/backend`

### DIFF
--- a/pkg/controllers/uiplugin/distributed_tracing.go
+++ b/pkg/controllers/uiplugin/distributed_tracing.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	osv1alpha1 "github.com/openshift/api/console/v1alpha1"
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/uiplugin/distributed_tracing.go
```diff
@@ -29,13 +30,36 @@ func createDistributedTracingPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace,
...
+	proxyName, proxyNamespace := "distributed-tracing-console-plugin", "openshift-tracing"
+
+	if plugin.Spec.TroubleshootingPanel != nil {
+		if plugin.Spec.TroubleshootingPanel.Korrel8r.Name != "" {
+			proxyName = plugin.Spec.TroubleshootingPanel.Korrel8r.Name
+		}
+		if plugin.Spec.TroubleshootingPanel.Korrel8r.Namespace != "" {
+			proxyNamespace = plugin.Spec.TroubleshootingPanel.Korrel8r.Namespace
+		}
+	}
```
jgbernalp, line 42: The code here is for korrel8r and the troubleshooting panel. But I think the intention here is to define this fields in the `DistributedTracingConfig` spec so they can be configured if not using the defaults

pkg/controllers/uiplugin/distributed_tracing.go
```diff
@@ -29,13 +30,36 @@ func createDistributedTracingPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace,
 		extraArgs = append(extraArgs, fmt.Sprintf("-features=%s", strings.Join(features, ",")))
 	}
 
+	proxyName, proxyNamespace := "distributed-tracing-console-plugin", "openshift-tracing"
```
jgbernalp, line 33:
```suggestion
	defaultServiceName, defaultServiceNamespace := "distributed-tracing-console-plugin", "openshift-tracing"
```

jgbernalp (changes requested):

pkg/controllers/uiplugin/distributed_tracing.go
PeterYurkovich: Name and namespace are passed in as parameters to the `createDistributedTracingPluginInfo` function and we should use them so that the service and the proxy don't get out of sync.

zhuje: I updated this in the latest commit.

jgbernalp (approved):

PR #503, opened 2024-06-07
COO-181: fix: Change only the console plugins to avoid mutating existing cluster configuration

jgbernalp:
This PR adjusts the merge object so only console plugins are changed in the console resource to avoid affecting other fields.


simonpasquier:
/lgtm
/approve
/hold

unhold at your will

PR #504, opened 2024-06-07
COO-180: fix: bind service with correct container port for korrel8r

jgbernalp:
use correct port for korrel8r deployment so the service binds correctly


periklis (approved):

simonpasquier (commented):
/lgtm

PR #510, opened 2024-06-12
fix: shellcheck wrong curl

marioferh:
Action publish-olm-candidate https://github.com/rhobs/observability-operator/actions/runs/9462114451/job/26065129373 is failing due an error in shellheck curl.

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -188,8 +188,8 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
 		set -ex ;\
 		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
 		cd $$(mktemp -d) ;\
-		OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.$$(OS).$$(ARCH).tar.xz";\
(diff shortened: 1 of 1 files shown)


Makefile.tools
```diff
@@ -188,8 +188,8 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
 		set -ex ;\
 		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
 		cd $$(mktemp -d) ;\
-		OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-		curl -sSLo shellcheck-stable.tar.xz "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.$$(OS).$$(ARCH).tar.xz";\
+		OS=$(shell go env GOOS) && ARCH=$(shell uname -m) && \
```
simonpasquier, line 191: I don't think that we need to change the way the ARCH var is filled?

marioferh: output of arch is:
`ARCH=amd64`
then:
`https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.limux.amd64.tar.xz`
and that's 404

Correct one is:
`https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz`

simonpasquier: :+1:

simonpasquier (approved):
/lgtm

PR #515, opened 2024-06-13
COO-186: fix: add korrel8r proxy configuration to logging view plugin

jgbernalp:
This PR adds the korrel8r proxy configuration so the the logging view plugin can discover korrel8r and correlation links are shown.

### DIFF
--- a/pkg/controllers/uiplugin/logging.go
+++ b/pkg/controllers/uiplugin/logging.go
@@ -62,6 +62,16 @@ func createLoggingPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace, name, image
 					Port:      8080,
 				},
 			},
+			{
+				Type:      "Service",
+				Alias:     "korrel8r",
+				Authorize: true,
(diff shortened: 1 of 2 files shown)


pkg/controllers/uiplugin/logging.go
```diff
@@ -62,6 +62,16 @@ func createLoggingPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace, name, image
 					Port:      8080,
 				},
 			},
+			{
```
simonpasquier, line 65: Is it ok to create the proxy entry even when the korrel8r plugin isn't installed?

jgbernalp: I think is ok as the console proxy just allows the plugin to communicate to the in cluster service. If korrel8r deployment fails, or is not there, requests will fail and the plugin won't show the correlation links.

periklis (approved):
LGTM

PR #483, opened 2024-05-14
bump to go v1.22

jan--f:

PR #516, opened 2024-06-13
Bump go 1.22

marioferh:
copy of https://github.com/rhobs/observability-operator/pull/483/commits as cachito issues are fixed


danielmellado:
This fails on CI, looks like a little bit of chicken and egg problem.

PR #517, opened 2024-06-13
Add Clusterrole to allow Korrel8r to view all signal data

shwetaap:
This PR creates the required clusterroles and clusterrolebindingss, so that korrel8r can query all the required resources
Adds additional clusterrolebinding and rolebinding so that korrel8r can connect to Prometheus endpoints and retrieve relevant data. 
All of the roles and rolebindings are created within the operator

### DIFF
--- a/bundle/manifests/observability-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/observability-operator.clusterserviceversion.yaml
@@ -271,6 +271,25 @@ spec:
           - ""
           resources:
           - configmaps
+          - endpoints
+          - events
+          - namespaces
+          - nodes
+          - persistentvolumeclaims
+          - persistentvolumes
+          - pods
+          - replicationcontrollers
+          - secrets
+          - serviceaccounts
+          - services
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          resources:
+          - configmaps
           - serviceaccounts
           - services
           verbs:
@@ -307,6 +326,17 @@ spec:
         - apiGroups:
           - apps
           resources:
+          - daemonsets
+          - deployments
+          - replicasets
+          - statefulsets
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - apps
+          resources:
           - deployments
           verbs:
           - create
@@ -314,6 +344,23 @@ spec:
           - list
           - patch
           - update
+          - watch
+        - apiGroups:
+          - autoscaling
+          resources:
+          - horizontalpodautoscalers
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - batch
+          resources:
+          - cronjobs
+          - jobs
+          verbs:
+          - get
+          - list
           - watch
         - apiGroups:
           - config.openshift.io
@@ -364,6 +411,34 @@ spec:
           - infrastructure
           verbs:
           - get
+        - apiGroups:
+          - loki.grafana.com
+          resources:
+          - application
+          - audit
+          - infrastructure
+          - network
+          verbs:
+          - get
+        - apiGroups:
+          - monitoring.coreos.com
+          resourceNames:
+          - main
+          resources:
+          - alertmanagers/api
+          verbs:
+          - get
+          - list
+        - apiGroups:
+          - monitoring.coreos.com
+          resourceNames:
+          - k8s
+          resources:
+          - prometheuses/api
+          verbs:
+          - create
+          - get
+          - update
         - apiGroups:
           - monitoring.rhobs
           resources:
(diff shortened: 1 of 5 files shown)


deploy/dependencies/korrel8r/serviceaccount.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: korrel8r-sa
```
jgbernalp, line 4: should this be:
```suggestion
  name: obo-korrel8r-sa
```
based on the change on the deployment?

danielmellado: Agree with that, also note that changing this would also change the ServiceAccountName later on

pkg/controllers/uiplugin/components.go
```diff
@@ -373,7 +373,7 @@ func newKorrel8rDeployment(name string, namespace string, info UIPluginInfo) *ap
 					Labels:    componentLabels(name),
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: info.Name + serviceAccountSuffix,
+					ServiceAccountName: "obo-" + name + serviceAccountSuffix,
```
danielmellado, line 376: if you just name the SA obo-korrel8r-sa, you'd need to drop the prefix here

deploy/dependencies/korrel8r/serviceaccount.yaml
shwetaap: From what I understand The kustomization file updates the sa and all other resources with a prefix "obo" https://github.com/rhobs/observability-operator/blob/main/deploy/dependencies/kustomization.yaml#L43. So when I run the `make bundle-image`  the new manifests created already have the `obo` prefix automatically. Is there another way this is commonly built where the prefix  does not get added automatically?

danielmellado: Oh, yeah, forgot about that part tbh! disregard this then ;) @jan--f mind taking a look as well? Thanks!

anpingli:
Test using this PR, the korrel8r is deployed in openshift-operator, while obo-korrel8r-view is bind to **operators/obo-korrel8r-sa**. it should be **openshift-operators/obo-korrel8r-sa**


oc get ClusterRoleBinding obo-korrel8r-view -o json
{
    "apiVersion": "rbac.authorization.k8s.io/v1",
    "kind": "ClusterRoleBinding",
    "metadata": {
        "creationTimestamp": "2024-06-24T08:31:00Z",
        "labels": {
            "app.kubernetes.io/part-of": "observability-operator",
            "olm.managed": "true",
            "olm.owner": "observability-operator.v0.0.1-dev"
        },
        "name": "obo-korrel8r-view",
        "resourceVersion": "251006",
        "uid": "65b72a1a-408c-4584-8862-8ede4e4b7d04"
    },
    "roleRef": {
        "apiGroup": "rbac.authorization.k8s.io",
        "kind": "ClusterRole",
        "name": "obo-korrel8r-view"
    },
    "subjects": [
        {
            "kind": "ServiceAccount",
            "name": "obo-korrel8r-sa",
            "namespace": "operators"
        }
    ]
}

deploy/dependencies/korrel8r/clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: korrel8r-view
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: korrel8r-view
+subjects:
+- kind: ServiceAccount
+  name: korrel8r-sa
```
anpingli, line 11: It should be <plugin_name>-sa.

deploy/dependencies/korrel8r/clusterrole.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ClusterRole
+metadata:
+  name: korrel8r-view
+rules:
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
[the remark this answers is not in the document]
danielmellado: I still see quite a lot of broad access to the resources here. Is this really needed for korrel8r.

jan--f:
I think we should consider the following points:

1. If we deploy korrel8r resources that needs certain permissions we should consider adding that to the operator. It doesn't seem like a great user experience to tell a cluster admin to modify/create a resource so that the operator can function properly.
2. Those permissions should be as minimal as possible while allowing all needed functionality.

So  lets pare down the cluster role to what korrel8r needs. Then the role and rolebinding can be created by the controller, making the subject namespace dynamic. Additionally this adds the needed permissions to the operator.
The service account can either stay in the bundle (OLM will add the correct namespace) or also move to be created by the controller.

shwetaap:
@jan--f Made changes to the code as suggested. All (cluster)role and (cluster)rolebinding creation has moved to the operator. The controller now creates the clusterrole needed for korrel8r and creates the clusterrolebinding& rolebinding using existing monitoring clusterrrole and role respectively  for allowing access to prometheus and alertmanager endpoints via korrle8r

periklis (approved):
LGTM

PR #519, opened 2024-06-14
chore: update GitHub actions to setup-go@v5

simonpasquier:
This change updates all actions to use actions/setup-go@v5. In this version, caching is enabled by default hence it removes all `cache: true` occurrences.


marioferh (commented):
/lgtm

PR #530, opened 2024-07-10
COO-225: fix: Add support for ConsolePlugin v1

jgbernalp:
This PR adds support for ConsolePlugin CR v1 for OCP 4.17. For older versions the v1alpha1 support was kept

### DIFF
--- a/pkg/controllers/uiplugin/components.go
+++ b/pkg/controllers/uiplugin/components.go
@@ -7,9 +7,12 @@ import (
 	"hash/fnv"
 	"io"
 	"sort"
+	"strings"
 	"text/template"
 
+	osv1 "github.com/openshift/api/console/v1"
 	osv1alpha1 "github.com/openshift/api/console/v1alpha1"
+	"golang.org/x/mod/semver"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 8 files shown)


pkg/controllers/uiplugin/components.go
```diff
@@ -45,14 +48,32 @@ var (
 	korrel8rConfigYAMLTmplFile embed.FS
 )
 
-func pluginComponentReconcilers(plugin *uiv1alpha1.UIPlugin, pluginInfo UIPluginInfo) []reconciler.Reconciler {
+func isClusterVersionAheadOrEqual(clusterVersion, nextClusterVersion string) bool {
```
lihongyan1, line 51: a nitpicking, feels the variable names of isClusterVersionAheadOrEqual is unreasonable and confusing, maybe better change as `func isClusterVersionAheadOrEqual(currentClusterVersion, clusterVersion string) bool {`

lihongyan1:
Verified the bug COO-225 with this PR, the issue is gone on OCP4.17

pkg/controllers/uiplugin/components.go
jan--f: Yeah I agree, the naming is awkward currently. I see how future devs might think the clusterVersion in the name is related to the variable name.

jgbernalp: updated to `func isVersionAheadOrEqual(currentVersion, version string) bool {`

jan--f:
/lgtm
Thanks!

PR #531, opened 2024-07-17
test: enable OCP mode and add auto case for uiplugin dashboard

lihongyan1:
#[COO-237](https://issues.redhat.com//browse/COO-237) Change OCP script to enable OCP mode
#[COO-160](https://issues.redhat.com//browse/COO-160) Add uiplugin-DB auto testing

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 	"testing"
 
-	openshiftClientset "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -24,11 +23,11 @@ import (
 )
 
 type Framework struct {
-	kubernetes kubernetes.Interface
-	openshift  openshiftClientset.Interface
-	Config     *rest.Config
-	K8sClient  client.Client
-	Retain     bool
+	kubernetes         kubernetes.Interface
+	Config             *rest.Config
+	K8sClient          client.Client
+	Retain             bool
+	IsOpenshiftCluster bool
 }
 
 // StartPortForward initiates a port forwarding connection to a pod on the localhost interface.
@@ -126,18 +125,6 @@ func (f *Framework) getKubernetesClient() (kubernetes.Interface, error) {
 	return f.kubernetes, nil
 }
 
-func (f *Framework) getOpenshiftClient() (openshiftClientset.Interface, error) {
-	if f.openshift == nil {
-		c, err := openshiftClientset.NewForConfig(f.Config)
-		if err != nil {
-			return nil, err
-		}
-		f.openshift = c
-	}
-
-	return f.openshift, nil
-}
-
 func (f *Framework) Evict(pod *corev1.Pod, gracePeriodSeconds int64) error {
 	delOpts := metav1.DeleteOptions{
 		GracePeriodSeconds: &gracePeriodSeconds,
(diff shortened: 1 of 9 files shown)


test/run-e2e.sh
```diff
@@ -209,6 +213,10 @@ parse_args() {
 			RUN_REGEX=$1
 			shift
 			;;
+		--ocp)
```
jan--f, line 216: I would prefer to have the tests determine whether they run on OCP or not. I think checking for the presence of the `ClusterVersion` or `ClusterOperator` CRD  should be good enough. We can get this up once during test setup and the `framework` can pass it to all tests. Then a test can `t.Skip` if it can't run.

test/e2e/framework/framework.go
```diff
@@ -156,3 +170,17 @@ func (f *Framework) CleanUp(t *testing.T, cleanupFunc func()) {
 		}
 	})
 }
+
+func (f *Framework) IsOpenshiftCluster() bool {
+	c, err := f.getOpenshiftClient()
```
jan--f, line 175: I think we might be better off to initialize this in https://github.com/rhobs/observability-operator/blob/62bcd73f3cce816269e038d7d87f74ba24aa42ae/test/e2e/main_test.go#L84 and rely on the client being there. Otherwise we have to remember to initialize this in future functions that needs the openshift client.

test/e2e/framework/framework.go
```diff
@@ -156,3 +170,17 @@ func (f *Framework) CleanUp(t *testing.T, cleanupFunc func()) {
...
+	if err != nil {
+		fmt.Printf("failed to get openshiftClient: %s", err)
+		return false
+	}
+	_, err = c.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+	if err != nil {
```
jan--f, line 181: Not sure about this one. To me this reads we implicitly assume the error is a 404, in which case we can skip a test. If this returns a different error (say a 5xx) we would want to propagate an error. Otherwise we risk skipping the openshift tests even if we are in an openshift cluster. @simonpasquier @machine424 wdyt?

We might be better off to only return false `if apierrors.IsNotFound(err)`.

lihongyan1:
ci/prow/images job failed not for obo, ocp-testplatform team solve in progress

lihongyan1:
@jan--f @simonpasquier @jgbernalp please help review, I would like this to merge so that we can run on IBM P/Z prow job in future.

test/e2e/main_test.go
```diff
@@ -83,17 +86,28 @@ func main(m *testing.M) int {
 
 func setupFramework() error {
 	cfg := config.GetConfigOrDie()
+	scheme := operator.NewScheme(&operator.OperatorConfiguration{})
+	err := configv1.AddToScheme(scheme)
```
simonpasquier, line 90: The doc says that AddToScheme is deprecated.
```suggestion
	err := configv1.Install(scheme)
```

test/e2e/main_test.go
```diff
@@ -119,3 +133,15 @@ func createNamespace(name string) (func(), error) {
 
 	return cleanup, nil
 }
+
+func isOpenshiftCluster(k8sClient client.Client) (bool, error) {
+	clusterVersion := &configv1.ClusterVersion{}
+	err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: "version"}, clusterVersion)
+	if err == nil {
+		return true, nil
+	} else if strings.Contains(err.Error(), `no matches for kind "ClusterVersion" in version "config.openshift.io/v1"`) {
```
simonpasquier, line 142: Can we use https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#IsNoMatchError to check the error type rather than looking at the error string?

PR #539, opened 2024-08-01
chore: update the Korrel8r Image

shwetaap:
Updated the Korrel8r Image with the changes after the Enhanced Dev Preview


simonpasquier:
@shwetaap you'd need to fix the commit message (the linter says ` subject may not end with full stop `").

cmd/operator/main.go
```diff
@@ -43,7 +43,7 @@ var defaultImages = map[string]string{
 	"ui-troubleshooting-panel": "quay.io/openshift-observability-ui/troubleshooting-panel-console-plugin:v0.1.0",
 	"ui-distributed-tracing":   "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v0.1.0",
 	"ui-logging":               "quay.io/openshift-logging/logging-view-plugin:6.0.0",
-	"korrel8r":                 "quay.io/korrel8r/korrel8r:0.6.6",
```
jgbernalp, line 46: Bare in mind that this only changes the default image if none is passed. The actual image will be passed in the `--images` flag. If the intention is to also update the image that is passed when building the operator, we might need to update CPaaS as well.

PR #541, opened 2024-08-05
fix: adjust max version semver comparison to avoid excluding valid cluster versions

jgbernalp:
[COO-260](https://issues.redhat.com/browse/COO-260)

### DIFF
--- a/pkg/controllers/uiplugin/compatibility_matrix.go
+++ b/pkg/controllers/uiplugin/compatibility_matrix.go
@@ -44,14 +44,14 @@ var compatibilityMatrix = []CompatibilityEntry{
 	{
 		PluginType:        uiv1alpha1.TypeLogging,
 		MinClusterVersion: "v4.11",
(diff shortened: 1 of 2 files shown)


pkg/controllers/uiplugin/compatibility_matrix.go
```diff
@@ -93,10 +93,11 @@ func lookupImageAndFeatures(pluginType uiv1alpha1.UIPluginType, clusterVersion s
 	for _, entry := range compatibilityMatrix {
 		if entry.PluginType == pluginType {
 			canonicalMinClusterVersion := fmt.Sprintf("%s-0", semver.Canonical(entry.MinClusterVersion))
+			canonicalMaxClusterVersion := semver.Canonical(entry.MaxClusterVersion)
```
simonpasquier, line 96: why is it that we suffix with "-0" for min version and not for max version?

PR #556, opened 2024-08-28
build: add api-support annotations to monitoring CRDs

jan--f:


danielmellado (approved):
Thanks for this @jan--f, while I do approve of the changes at some point I'd love seeing a -even if quick'n dity- add to the docs somehow explaining this or the future us may forget at some point

jan--f:
We have https://docs.google.com/document/d/1sWAe_LfK-0gr-bu6L3MyQo8Pna9Z8YUppSnwHLV-sRU/edit#heading=h.bupciudrwmna for now. This will have to find its way into the official docs anyway. I'll think about whether this makes sense to add here as well.

PR #557, opened 2024-08-28
docs: fix UI docs

iNecas:
When following the original docs, I was getting:

```
Danger alert:An error occurred
Error "Invalid value: "object": UIPlugin name must be
'distributed-tracing' if type is DistributedTracing" for field "<nil>".
```


simonpasquier (commented):
/lgtm

PR #558, opened 2024-08-29
build: update prometheus-operator manifests to 0.75.2

jan--f:
This syncs the version with the included go module.


simonpasquier (commented):
/lgtm

PR #560, opened 2024-08-30
chore: flag PrometheusAgent CRD as internal

simonpasquier:
This is aligned with the other CRDs like Prometheus and Alertmanager.


rexagod:
/lgtm

Makes sense, these resources are not meant for user manipulation.

marioferh (commented):
/lgtm

PR #563, opened 2024-09-06
fix: avoid trying to run UI plugins as root

iNecas:
When trying to use the troubleshooting-panel UIPlugin, I was getting:

```
container has runAsNonRoot and image will run as root
(pod: "troubleshooting-panel-...
 container: troubleshooting-panel)'
 reason: CreateContainerConfigError
```

I guess it's because the troubleshooting panel is build without setting the non-root user https://github.com/openshift/troubleshooting-panel-console-plugin/blob/main/Dockerfile.

I'm not sure why anyone else wasn't hitting this issue… I was running this against OCP 4.16.


simonpasquier:
I would defer to @PeterYurkovich and @jgbernalp but I'm not sure why we would need to set the user id?

jgbernalp:
We don't need this anymore as non root users are set in the docker files: https://github.com/rhobs/konflux-coo/blob/main/Dockerfile.ui-logging#L39 

There was also an issue that prevented to create the right security context when the operators namespace had the flag `openshift.io/run-level`: https://access.redhat.com/solutions/6495371

PR #574, opened 2024-09-18
chore: support `tools` target for arm64

rexagod:
Support Macs, and fixes some broken links.

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -126,7 +126,7 @@ $(OC) oc: $(TOOLS_DIR)
 		[[ -f $(OC) ]] && exit 0 ;\
 		OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
 		if [[ $${OS} == "darwin" ]]; then OS="mac"; fi ;\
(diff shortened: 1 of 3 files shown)


simonpasquier:
/approve
/hold
let's sort which we want to merge first between this and #566

Makefile.tools
```diff
@@ -189,11 +188,10 @@ $(SHELLCHECK) shellcheck: $(TOOLS_DIR)
 		[[ -f $(SHELLCHECK) ]] && exit 0 ;\
 		cd $$(mktemp -d) ;\
 		OS=$(shell go env GOOS) && ARCH=$(shell uname -m) && \
-		curl -sSLo shellcheck-stable.tar.xz https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.$${OS}.$${ARCH}.tar.xz ;\
-		tar -xJf shellcheck-stable.tar.xz ;\
-		cp shellcheck-stable/shellcheck $(SHELLCHECK) ;\
-		version=$(SHELLCHECK_VERSION) ;\
```
jan--f, line 195: why remove the version checks?

rexagod: With this patch, the exact version is specified in the URL, so we can be sure the same version is fetched.

simonpasquier (commented):
/lgtm
/hold
let's hold it until we're sure that we don't need to cook a 0.4.3 release.

PR #575, opened 2024-09-23
OU-407: Add monitoring-plugin uiplugin

PeterYurkovich:
This PR looks to add an ACM specific version of the monitoring plugin into obo. It is reliant on changes in this [PR](https://github.com/openshift/monitoring-plugin/pull/180) which add the ACM specific flags.

This plugin should be deployable alongside the monitoring plugin deployed by CMO to allow for the local-cluster to still have the observability pages, while enabling the observability pages under ACM using the obo deployment.


pkg/controllers/uiplugin/monitoring.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("alertmanager location can not be empty for plugin type %s", plugin.Spec.Type)
+	}
+	if config.ThanosQuerier.Name == "" || config.ThanosQuerier.Namespace == "" {
+		return nil, fmt.Errorf("ThanosQuerier location can not be empty for plugin type %s", plugin.Spec.Type)
+	}
+
+	pluginInfo := &UIPluginInfo{
+		Image:       image,
+		Name:        name,
+		ConsoleName: "acm-monitoring-plugin",
```
jgbernalp, line 31: Although ACM is our first customer for the monitoring plugin, this is not specific to ACM. For example, customers installing the plugin for getting support for Perses dashboards or incidents will load the plugin into the cluster along with the monitoring-plugin provided by CMO. In this case we need to only enable the specific features needed to avoid duplicating the views. Maybe this could be:

```suggestion
		ConsoleName: "monitoring-console-plugin",
```

pkg/controllers/uiplugin/compatibility_matrix_test.go
```diff
@@ -185,6 +185,26 @@ func TestLookupImageAndFeatures(t *testing.T) {
...
+			pluginType:     uiv1alpha1.TypeMonitoring,
+			clusterVersion: "v4.16",
+			expectedKey:    "ui-monitoring",
+			expectedErr:    fmt.Errorf("plugin %q: no compatible image found for cluster version %q", uiv1alpha1.TypeMonitoring, "v4.16"),
+		},
+		{
+			pluginType:       uiv1alpha1.TypeMonitoring,
+			clusterVersion:   "v4.17",
+			expectedKey:      "ui-monitoring",
+			expectedFeatures: []string{"acm"},
```
jgbernalp, line 198: maybe this is a collection of more specific features like `["alerting", "metrics"]`, WDYT?

pkg/controllers/uiplugin/monitoring.go
jgbernalp: I addition to that we could do a check if the `monitoring-plugin` is present on the cluster to clean up the features to avoid duplication. WDYT?

pkg/controllers/uiplugin/compatibility_matrix_test.go
PeterYurkovich: As per our discussion, this has been updated to "acm-alerting" in the most recent commit.

pkg/controllers/uiplugin/monitoring.go
PeterYurkovich: 1) The ConsoleName has been changed to "monitoring-console-plugin" and the display name has been changed to "Monitoring Console Plugin".
2) As per our discussion, this deployment will not include the typical monitoring plugin behavior, only the "acm-alerting" specific ones. Checks to enable extra features if they are missing will be left to future work.

PeterYurkovich:
Updates to the monitoring-plugin [PR](https://github.com/openshift/monitoring-plugin/pull/180) have been completed which sync feature names and parameters with changes in this PR

pkg/controllers/uiplugin/monitoring.go
```diff
@@ -0,0 +1,123 @@
...
+				Authorization: "UserToken",
+				Endpoint: osv1.ConsolePluginProxyEndpoint{
+					Type: osv1.ProxyTypeService,
+					Service: &osv1.ConsolePluginProxyServiceConfig{
+						Name:      name,
+						Namespace: namespace,
+						Port:      9445,
+					},
+				},
+			},
```
jgbernalp, line 75: Do we need this independent proxies or could we use only the `backend` adding a `proxy` endpoint?, e.g.
`backend/proxy/alertmanager/...` this would mean we have control of the proxies from the plugin side.

PeterYurkovich: Unfortunately, the "backend" parameter only works for GET requests, a "proxy" is needed for every other type. The alert-manager proxy uses DELETE for silences, and while I'm not sure if the thanos-proxy needs any other request types, I left it like this for safety

pkg/controllers/uiplugin/components.go
```diff
@@ -258,8 +259,9 @@ func newDeployment(info UIPluginInfo, namespace string, config *uiv1alpha1.Deplo
 					ServiceAccountName: info.Name + serviceAccountSuffix,
 					Containers: []corev1.Container{
 						{
-							Name:  info.Name,
-							Image: info.Image,
+							Name:            info.Name,
+							Image:           info.Image,
+							ImagePullPolicy: "Always",
```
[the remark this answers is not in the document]
PeterYurkovich: Removed in the latest commit

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -132,6 +135,43 @@ type LokiStackReference struct {
 	Name string `json:"name"`
 }
 
+// MonitoringConfig contains options for configuring the monitoring console plugin.
+type MonitoringConfig struct {
+	// Alertmanager points to the alertmanager instance of which it should create a proxy to.
+	//
+	// +kubebuilder:validation:Required
```
jgbernalp, line 142: This would be required only on the case the monitoring plugin has been started with the acm-alerting flag. Should we consider doing this optional or add some validation based on the features enabled?

pkg/controllers/uiplugin/compatibility_matrix.go
```diff
@@ -81,6 +81,15 @@ var compatibilityMatrix = []CompatibilityEntry{
 			"dev-alerts",
 		},
 	},
+	{
+		PluginType:        uiv1alpha1.TypeMonitoring,
+		MinClusterVersion: "v4.14",
+		MaxClusterVersion: "",
+		ImageKey:          "ui-monitoring",
+		Features: []string{
+			"acm-alerting",
```
jgbernalp, line 90: can we detect we are running in an ACM hub cluster?

PeterYurkovich: ACM detection has been added in the most recent changes. As per the conversation [here](https://redhat-internal.slack.com/archives/C063J3F88MV/p1729256903923739?thread_ts=1729058727.278719&cid=C063J3F88MV), we will be reconciling the uiplugin based on ACM version, just as the cluster version is also taken into account.

pkg/controllers/uiplugin/compatibility_matrix_test.go
```diff
@@ -44,34 +47,60 @@ func TestCompatibilityMatrixMaxVersions(t *testing.T) {
...
 		clusterVersion   string
+		listFunction     ListFunction
 		expectedKey      string
 		expectedErr      error
 		expectedFeatures []string
 	}{
 		{
 			pluginType:     uiv1alpha1.TypeDashboards,
 			clusterVersion: "4.10",
+			listFunction:   stubbedFunc,
```
jgbernalp, line 82: could we pass the ACM version here instead of a function?

pkg/controllers/uiplugin/compatibility_matrix.go
```diff
@@ -75,15 +96,28 @@ var compatibilityMatrix = []CompatibilityEntry{
...
 			"dev-alerts",
 		},
 	},
+	{
+		PluginType:        uiv1alpha1.TypeMonitoring,
+		MinClusterVersion: "v4.14",
+		MaxClusterVersion: "v4.17",
+		ImageKey:          "ui-monitoring",
+		MinAcmVersion:     "v2.11",
+		MaxAcmVersion:     "v2.13",
```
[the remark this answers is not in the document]
PeterYurkovich: I set the top end as unbonded for now in the most recent commit

pkg/controllers/uiplugin/compatibility_matrix_test.go
PeterYurkovich: I wasn't able to move the check out of the reconciliation loop as the check was missing the reconciliation context. However, I did refactor a bit to have the function call happen a few levels upward so that this does only pass in the version string rather than a function.

jgbernalp (approved):

PR #581, opened 2024-10-07
COO-451: fix: use expected plugin console names

jgbernalp:
This PR uses the plugin names the console expect so the routing for resources and proxies works correclty


simonpasquier (commented):
/lgtm

PR #584, opened 2024-10-09
COO-300: `Updater` reconciles should create object if it doesn't exist

rexagod:
Check if the object doesn't exist, and if so create it. Prefer patching
over updating and replacing the entire object.

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -35,6 +36,16 @@ func (r Updater) Reconcile(ctx context.Context, c client.Client, scheme *runtime
 	}
 
 	if err := c.Patch(ctx, r.resource, client.Apply, client.ForceOwnership, client.FieldOwner("observability-operator")); err != nil {
+		if apierrors.IsNotFound(err) {
```
simonpasquier, line 39: IIUC c.Patch() should already take care of creating the object if missing so I suppose that the issue isn't here.

jan--f: Agreed. This has worked since very early on. What is the behavior that needs fixing?

rexagod: Ah, right, we are already using `client.Apply` here ([ref](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L635-L637)). I'll take another look into this.

rexagod: @jan--f The issue reported was that certain [logging objects](https://github.com/rhobs/observability-operator/blob/4ce18d98ec93542f1bea400ec320f66bf2ceeaa3/pkg/controllers/uiplugin/logging.go#L174) deployed by the controller at UIPlugin creation are not recreated when they are deleted. Nonetheless, now that I can setup OBO locally (thanks to [Beaker](http://beaker.engineering.redhat.com)), I'll be able to test my changes before pushing them. 😅

rexagod:
I verified that this is no longer an issue, closing. Sorry for the noise here folks! 😓

PR #604, opened 2024-10-23
COO-318: add uiplugin support annotations when created

PeterYurkovich:
This PR looks to address [COO-318](https://issues.redhat.com/browse/COO-318) by adding support level annotations to the UIPlugins based on their type after they are created.

This adds an extra couple of reconciliations for each plugin, but doesn't trigger a reconciliation loop and is able to stabilize.

PR #607, opened 2024-10-28
UIPlugin: add support level annotation during reconciliation

jan--f:
Fixes: [COO-449](https://issues.redhat.com//browse/COO-449)


jan--f:
/close
already implemented in #604 .

PR #608, opened 2024-10-28
fix: add finalizers to cleanup cluster scoped resources on stack dele…

jan--f:
…tion

Fixes: [COO-495](https://issues.redhat.com//browse/COO-495)

### DIFF
--- a/pkg/controllers/monitoring/monitoring-stack/controller.go
+++ b/pkg/controllers/monitoring/monitoring-stack/controller.go
@@ -19,6 +19,7 @@ package monitoringstack
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/monitoring/monitoring-stack/controller.go
```diff
@@ -144,11 +147,40 @@ func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, nil
 	}
 
+	// Check if the plugin is being deleted
 	if !ms.ObjectMeta.DeletionTimestamp.IsZero() {
-		logger.V(6).Info("skipping reconcile since object is already schedule for deletion")
+		logger.V(6).Info("removing cluster scoped resources")
+
+		reconcilers := stackComponentCleanup(ms)
```
danielmellado, line 154: just a nit over here, if we have the prometheus and the alertamanager names, why not adding that also as a log? It may be useful when debugging. Otherwise lgtm!

jan--f: This is just an info log. In line 158 an error is logged in case something goes wrong, that'll include the resource name that creates an error. Is that what you're after?

Not sure we should log 6 resource names on the happy path.

danielmellado: nvm, mistook the 'Info' for 'Err' in my mind, LGTM then, thanks!

PR #627, opened 2024-11-14
build: use busybox as container runner

jan--f:
The new gather script requires a shell.


jan--f:
Fixes: COO-498

simonpasquier:
Should/could we use RHEL UBIs?

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html-single/building_running_and_managing_containers/index#assembly_types-of-container-images_building-running-and-managing-containers

jan--f:
For upstream though? Not sure just anyone can pull those.
/retest

alanconway:
@jan--f  FYI: the RHEL UBI images are public, I'm using them in korrel8r and I've just verified that I can `podman pull registry.access.redhat.com/ubi9/ubi-micro` as a fresh user with no special permissions.

alanconway:
@jan--f ^^ this is not a statement of any policy that I know of, it's just what I did. I used to use some google minimal base images, I switched to UBI as I figured it might ease security scanning etc.

simonpasquier (commented):
/lgtm

marioferh (commented):
/lgtm

PR #665, opened 2025-01-30
[OSD-28074] Change SSS resourceApplyMode to Upsert

apahim:
The goal is to Orphan the SSS resources now, to later remove the SSS without removing the objects listed.

According to the docs:
https://github.com/openshift/hive/blob/master/docs/syncset.md

"Defaults to "Upsert", which indicates that objects will be created and updated to match the SyncSet. Existing SyncSet resources that are not listed in the SyncSet are not deleted."


simonpasquier (approved):

simonpasquier:
I merged it without the CI approval as it was discussed with @apahim and @jan--f . Since it's only changing the SelectorSyncSet template, it must have no effect on the operator itself.

PR #667, opened 2025-02-04
OSD-28131: Deploying COO in place of OBO on SC clusters

Nikokolas3270:
This PR removes OBO subscription from Hypershift service clusters (SCs)

The following MR install COO on those clusters:
https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/1043

This MR must be merged after above PR

### DIFF
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -50,7 +50,6 @@ objects:
             operator: In
             values:
               - management-cluster
-              - service-cluster
       resourceApplyMode: Sync
       resources:
         - apiVersion: operators.coreos.com/v1alpha1


deploy/olm/syncselector-template.yaml
```diff
@@ -50,7 +50,6 @@ objects:
             operator: In
             values:
               - management-cluster
-              - service-cluster
```
apahim, line 53: I believe this will have to be a step-by-step approach.

First remove it here, then clean up the CSVs, then install COO.

Nikokolas3270: Nope, COO can be installed in parallel or even before OBO is uninstalled.
Of course while OBO is installed, COO install will fail... but the install will succeed as soon as OBO is uninstalled, that is to say once OBO subcription + CSV have been removed

apahim: Yeah, ok, if that’s not noise for SRE, then it’s fine.

Another topic… does it still make sense to control the installation from here? I believe OSDFM was the right choice for installing components in SCs/MCs, no?

Nikokolas3270: Ok, I will add the subscription in this file:
https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/blob/main/config/resources/service-cluster.yaml
Speaking of that: shouldn't we get rid of this template file after ALL clusters use COO in place of OBO? I mean I always found it strange to have a file in the operator codebase telling how the operator should be deployed on the infrastructure. To me, this should be decorrelated; remark that this operator is not the only operator suffering from this lack of boundaries.
WDYT?

apahim: Agreed. The attention point is the Namespace in the first SSS. If we delete it, even for a few seconds, the MonitoringStack CR in the SCs/MCs will be removed, as well as the Prom and AM instances and monitoring will be down for ROSA HCP :)

We will have to change the SSS to “Upsert” first, to them remove it, making sure it’s also deleted from Hive (delete:true in the saas-file target).

Then we can control the Namespace from somewhere else, like the OSDFM.

Nikokolas3270: COO deployment code moves there:
https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/1043

Nikokolas3270:
> @Nikokolas3270 @apahim is still pr still required?

@jan--f , yes it is still required... but most of the code moved there:
https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/1043

jan--f:
/hold
for https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/1043

jan--f:
Checkout the commit lint results. We follow conventional commits. A prefix of `chore: ` should fix it.

deploy/olm/syncselector-template.yaml
Nikokolas3270: Hello @apahim 
Actually, moving COO install/deployment in `osd-fleet-manager` is a not possible.
The problem is that [config/resources/service-cluster.yaml](https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/blob/main/config/resources/service-cluster.yaml) triggers too late in the install process, it is actually triggering after installing ACM which needs to install a RHOBS service monitor which is not possible if COO or OBO is not already installed.
I will dig in `osd-fleet-manager` code but I think we may need to keep defining COO subscription there.

PR #678, opened 2025-02-22
feat: add Incidents backend

falox:
This PR adds the Incidents backend.

When the `incidents` feature flag is set, the UIPlugin controller installs the [Health Analyzer](https://github.com/openshift/cluster-health-analyzer)  and the resources associated with it.

Fixes: OBSINTA-471


cmd/operator/main.go
```diff
@@ -45,6 +45,7 @@ var defaultImages = map[string]string{
 	"ui-logging":               "quay.io/openshift-observability-ui/logging-view-plugin:v6.0.0",
 	"ui-monitoring":            "quay.io/openshift-observability-ui/monitoring-console-plugin:latest",
```
falox, line 46: @jgbernalp if you don't pin a version, `latest` needs to be updated at least to https://github.com/openshift/monitoring-plugin/commit/16acec71e38c8c489fa4412f42b34a48efbb5ebb

falox: Thanks @PeterYurkovich

PR #690, opened 2025-02-28
feat: discover default lokistack

jgbernalp:
Fixes: [COO-332](https://issues.redhat.com/browse/COO-332)

This PR:

- Uses the lokiStack configuration if provided, otherwise discovers a lokistack in a default or provided namespace
- Extracts the dashboard plugin creation on a new file for consistency


jgbernalp:
cc @alanconway

pkg/controllers/uiplugin/logging.go
```diff
@@ -191,3 +200,39 @@ func loggingClusterRole(tenant string) *rbacv1.ClusterRole {
...
+var lokiStackResource = schema.GroupVersionResource{
+	Group: "loki.grafana.com", Version: "v1", Resource: "lokistacks",
+}
+
+// getLokiStack returns the LokiStack resource to use for the logging plugin.
+// It either uses the explicitly configured LokiStack or discovers one from the cluster.
+func getLokiStack(plugin *uiv1alpha1.UIPlugin, ctx context.Context, client dynamic.Interface) (*types.NamespacedName, error) {
+	config := plugin.Spec.Logging
+
+	if config != nil && config.LokiStack.Name != "" && config.LokiStack.Namespace != "" {
```
PeterYurkovich, line 213: It is a valid configuration to provide:

```yaml
lokistack:
  name: logging-loki
```

So we should complete the searchNamespace replacement for the namespace first and then make this if statement just check for the name.

Also if we are performing server side validation to determine if the namespace contains a lokistack should we also perform server side validation to check to see if the defined name+namespace combo exists?

jgbernalp: Refactored to check for a LokiStack with a specific name in the default or provided namespace.

PR #716, opened 2025-03-21
feat: deploying COO in place of OBO on SC clusters

Nikokolas3270:
Follow up of this PR: https://github.com/rhobs/observability-operator/pull/667

It was actually a bad idea to deploy COO through `osd-fleet-manager` as discussed [there](https://github.com/rhobs/observability-operator/pull/667#discussion_r2007813490).

Indeed `osd-fleet-manager` come too late in the install process of (Hypershift) SC clusters.
Typically ACM will be installed before COO is installed... but ACM needs to deploy OBO/COO specific service monitor which is only possible if COO is installed before ACM.

### DIFF
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -96,6 +96,41 @@ objects:
   - apiVersion: hive.openshift.io/v1
     kind: SelectorSyncSet
     metadata:
+      name: cluster-observability-operator-hypershift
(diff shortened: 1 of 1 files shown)


Nikokolas3270:
For info, related MR on `osd-fleet-manager`:
https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/1059

PR #719, opened 2025-03-26
fix: web UI case failure in downstream

lihongyan1:
web UI url changed in downstream
[COO-775](https://issues.redhat.com/browse/COO-775)


jan--f:
:+1: That is part of the Prometheus 3 upgrade
/lgtm

PR #721, opened 2025-03-31
COO-784: fix: raise resource requests and limits for operators

jan--f:


juzhao:
is it possible we don't set resources.limit to avoid more user projects created in the cluster?

jan--f:
> is it possible we don't set resources.limits to avoid such issue if more user projects created in the cluster?

Possible but not recommended: https://sdk.operatorframework.io/docs/best-practices/managing-resources/#general-guidelines

lihongyan1:
Test pass with PR, deployment has new resources set up

PR #748, opened 2025-05-22
Add CRD scaffolding for cluster observability / signal manager

pavolloffay:
Replaces #724 

This PR just adds scaffolding CRD (without controller) for the ClusterObservability CR. 

I would like to rename the original name `SignalManager` to `ClusterObservability`. I think it is less ambiguous and aligns with MCO API that we have agreed on in the past `MultiClusterObservability`. 

The long-term goal for the COO is to become a swiss-army knife for deploying observability stack on a single cluster and be used by MCO. Aligning the APIs will provide straight forward user experience any going from single cluster to multi-cluster deployments.

* MCO https://github.com/stolostron/multicluster-observability-operator/blob/a4b5212f393e5093701484d81c836eed4e84322f/operators/multiclusterobservability/api/v1beta2/multiclusterobservability_types.go#L15


alanconway (approved):
/lgtm

I did  notice some left-over references to "signal manager"

PR #773, opened 2025-05-29
feat: update perses operator to v0.1.12

PeterYurkovich:
Update the perses operator to the newest version, add secret retrieval to permission set of perses operator, and pass in the secret name and paths to retrieve the cert


zhuje (approved):
/lgtm

Issue #757, opened 2025-05-28
ClusterObservability progress


pavolloffay:
This issue tracks progress on the CusterObservability CRD 

- [x] API/CRD #754 
- [x] Controller - https://github.com/rhobs/observability-operator/pull/793
- [x] Install operators - https://github.com/rhobs/observability-operator/pull/793
- [x] Deploy instance - https://github.com/rhobs/observability-operator/pull/793
- [ ] Upgrade 
- [x] e2e tests #800 
- [x] Status -https://github.com/rhobs/observability-operator/pull/793
- [x] Can users add tenants? - no the tenants is a map field that is overriden by the operator
- [x] How could we support editting collector config?  - my merging it in the code

The initial design doc https://docs.google.com/document/d/1VpcxsbGAohBNLa0gaHod_S0mhjGozXOR_I9Ov9DPSyk/edit?tab=t.0

pavolloffay:
Done

PR #784, opened 2025-06-20
feat: add controller for clusterobservability

pavolloffay:
Updates #757 

* Recording: https://drive.google.com/file/d/1nnLdC7I31ALvigmBQIsksOKe0fxVU-Zs/view?usp=drive_link
* Readme: https://github.com/rhobs/observability-operator/blob/main/docs/clusterobservability.md#deploy-capability-but-dont-deploy-the-operators 

This PR adds controller for the `ClusterObservability` CR. The controller is enabled only on Openshift - when `--openshift.enabled=true`.

This PR does not add `ClusterObservability` CR to the bundle so even when the `--openshift.enabled=true` the implementation is not usable. This is on purpose, the feature will be enabled later on once all items from #757 are resolved.

To deploy the operator with this feature set `--openshift.enabled=true` and enable the CR https://github.com/rhobs/observability-operator/blob/c4564860c698cb8201d368c02de21650a8b7034c/deploy/crds/common/kustomization.yaml#L7


jan--f:
Thanks and apologies for the latency in reviewing this. Since iiuc the original design doc is a bit outdated I'll leave some stuff here for now, but perhaps its worth it updating the original doc or setup a new one.

I think we can discuss three facets of this for a bit of structure: sane default configurations for operators, an API that covers multiple signals and the (perhaps optional) installation of additional operators via Subscriptions.
Currently all of these are wrapped up in a single API and while those three features certainly have dependencies, I think a bit more separation would be good to have.

1. We had discussions before that perhaps it would be better for default configs to not just exist in an operator. There are OpenShift flavors that don't bring OLM but rely on other install methods. Can we address this within this proposal?
2. I'm not sure I understand the `CommonCapabilitiesSpec` type very well. What is the use case to disable capabilities but install them? Wouldn't that just duplicate the functionality the user has already by creating a Subscription CR?
Perhaps it might be better to ask the user to supply names of subscription CRs to use for a given signal. If the CR exists there is no need to install, otherwise we create the Subscription. If there is no subscription configured for a given signal, the components will not be deployed.
We can then also require at least on operator subscription to be present to avoid a CR that has no effect.

3. The install flow would be addressed by directly referencing a subscription and creating it if its not found. Additionally identifying an existing subscription would allow us to get some information from it and create more detailed status entries, e.g. about compatibility issues.

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -94,5 +120,5 @@ type OperatorsSpec struct {
 	// When the capability is enabled, the install is set to true, otherwise it is set to false.
 	// +optional
 	// +kubebuilder:validation:Optional
-	Install bool `json:"install,omitempty"`
+	Install *bool `json:"install,omitempty"`
```
tremes, line 123: I think this is related to Jan's comment  and I am not sure I understand it (and yes I probably missed that in the previous PR so apologies). I thought when a capability is enabled then everything required is installed. It's not the case?

pavolloffay: Correct, when a capability is enabled, the operator installs the required operators.

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -42,11 +42,35 @@ type ClusterObservabilitySpec struct {
...
 
 // ClusterObservabilityStatus defines the observed state of ClusterObservability.
-type ClusterObservabilityStatus struct{}
+type ClusterObservabilityStatus struct {
+	// OpenTelemetry defines the status of the OpenTelemetry capability.
+	// +optional
+	OpenTelemetry string `json:"opentelemetry,omitempty"`
+	// Tempo defines the status of the Tempo capability.
+	// +optional
+	Tempo string `json:"tempo,omitempty"`
```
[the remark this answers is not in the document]
pavolloffay: The value is `namespace/instance-name (version)`. I am happy to change it. Do you have any suggestions?

PR #790, opened 2025-06-25
Bump various dependencies

jan--f:
This bumps a bunch of dependencies.
I tried out letting an AI tool doe that, which works rather poorly since it needs prompting about the `openshift/api/console` dependencies and is rather slow to figure things out.
It did generate the `DEPENDENCY_CONSTRAINTS.md` which seems worthwhile keeping if actually correct. @jgbernalp can perhaps comment on that?

/cc @jgbernalp


jgbernalp:
@jan--f the version pinning on the OpenShift api is necessary as long as we have support for < 4.17. Some versions were also updated in release-1.2, pointing to the perses forks, included in this [sync](https://github.com/rhobs/observability-operator/pull/780) PR

jgbernalp:
LGTM

PR #791, opened 2025-06-27
COO-1044: test: unify test setups for local and ghub action runs

jan--f:
This change attempts to introduce a test setup script that works both locally and in ghub actions. Until now our test setups differed and sometimes lead to configuration drift and diverging test results. This should improve that by using the setup script as the single source of truth for the test setup.

Generated-with-ai: claude-4-sonnet


hack/setup-e2e-env.sh
```diff
@@ -0,0 +1,529 @@
...
+    fi
+
+    # Check /etc/hosts for local-registry entry (skip in CI)
+    if ! $SKIP_HOST_CHECK; then
+        if grep -q "local-registry" /etc/hosts 2>/dev/null; then
+            ok "/etc/hosts contains local-registry entry"
+        else
+            warn "/etc/hosts does not contain local-registry entry"
+            info "Add local-registry entry:"
+            printf "    ❯ echo \"127.0.0.1\tlocal-registry\" | sudo tee -a /etc/hosts"
```
lihongyan1, line 186: change as 
`printf "127.0.0.1\tlocal-registry\n" | sudo tee -a /etc/hosts`

lihongyan1: Maybe it'd better differentiate local_setup and github action setup, both call some common lib scripts. For me, it's difficult for me to know what should be executed locally.

lihongyan1: If not add the entry by executing the following, script will stop here with output https://github.com/rhobs/observability-operator/pull/791#issuecomment-3031501758
`printf "127.0.0.1\tlocal-registry\n" | sudo tee -a /etc/hosts`

If added the entry, will fail to create kind cluster with issue [COO-1048](https://issues.redhat.com/browse/COO-1048)

jan--f: `kind` also runs with docker, though not sure if that works better on a rhel system.

PR #793, opened 2025-07-01
feat: add controller for clusterobservability

pavolloffay:
Updates #757 

* Recording: https://drive.google.com/file/d/1nnLdC7I31ALvigmBQIsksOKe0fxVU-Zs/view?usp=drive_link
* Readme: https://github.com/rhobs/observability-operator/blob/main/docs/clusterobservability.md#deploy-capability-but-dont-deploy-the-operators 

This PR adds controller for the `ClusterObservability` CR. The controller is enabled only on Openshift - when `--openshift.enabled=true`.

This PR does not add `ClusterObservability` CR to the bundle so even when the `--openshift.enabled=true` the implementation is not usable. This is on purpose, the feature will be enabled later on once all items from #757 are resolved.

To deploy the operator with this feature set `--openshift.enabled=true` and enable the CR https://github.com/rhobs/observability-operator/blob/c4564860c698cb8201d368c02de21650a8b7034c/deploy/crds/common/kustomization.yaml#L7


pkg/apis/observability/v1alpha1/types.go
```diff
@@ -42,11 +42,35 @@ type ClusterObservabilitySpec struct {
...
+	// +optional
+	Tempo string `json:"tempo,omitempty"`
+
+	// Conditions provide status information about the instance.
+	// +listType=atomic
+	// +optional
+	Conditions []StatusConditions `json:"conditions,omitempty"`
+}
+
+type StatusConditions struct {
```
alanconway, line 65: Nit: there's a standard library for status conditions.
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition
Keeps API in line with "k8s best practice" such as it is: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1623-standardize-conditions/README.md

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -42,11 +42,35 @@ type ClusterObservabilitySpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Capabilities",order=2
 	Capabilities *CapabilitiesSpec `json:"capabilities,omitempty"`
 }
 
 // ClusterObservabilityStatus defines the observed state of ClusterObservability.
-type ClusterObservabilityStatus struct{}
+type ClusterObservabilityStatus struct {
+	// OpenTelemetry defines the status of the OpenTelemetry capability.
+	// +optional
+	OpenTelemetry string `json:"opentelemetry,omitempty"`
```
alanconway, line 54: What is the string, why isn't it a Condition?

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -94,5 +120,5 @@ type OperatorsSpec struct {
 	// When the capability is enabled, the install is set to true, otherwise it is set to false.
 	// +optional
 	// +kubebuilder:validation:Optional
-	Install bool `json:"install,omitempty"`
+	Install *bool `json:"install,omitempty"`
```
alanconway, line 123: Not keen on names "enabled" and "operators" but agree to disagree.
It does not make sense for "operators" to be an extension point, but "enabled" is not. 
We talked about adding "pattern" as an extension point for the enabled deployment, the extension point for the operator installation shouldn't be a separate thing. E.g. 
```
  Install bool  // If true, install the operators required for this capabillity
  Enable bool // If true, deploy the resources to enable this capability
  Pattern string // Use the named pattern to define a specific configuration for operators and resources 
                            // Optional: if not specified a default pattern will be used.
```

alanconway (approved):
/lgtm
Some suggestions but I'm happy to see this committed and continue the debate as it develops.

pkg/apis/observability/v1alpha1/types.go
pavolloffay: It's ns/instance-name (version) . I will add a comment in a follow up PR

PR #798, opened 2025-07-02
COO-1046: revert changes to required fields within the CRD

PeterYurkovich:
This PR fixes a change which added some required fields to the perses CRD, without incrementing the version, which caused the operator upgrade to fail.


zhuje (approved):
/lgtm

PR #801, opened 2025-07-07
feat: new Go module for the API

tremes:
This is a next step for the https://issues.redhat.com/browse/COO-862.


tremes:
The current CI failures are related to the `RUN go mod download` step in the Dockerfile. It doesn't work with the Go mod `replace`. We can temporarily comment the step in the Dockerfile or accept the failures since this will require a follow-up PR for removing the `replace` anyway.

jan--f:
> The current CI failures are related to the `RUN go mod download` step in the Dockerfile. It doesn't work with the Go mod `replace`. We can temporarily comment the step in the Dockerfile or accept the failures since this will require a follow-up PR for removing the `replace` anyway.

Can't we fix this by adding a temporary cp line?
```
 COPY pkg/apis/go.mod pkg/apis/go.mod
 ```

jan--f:
Sorry I think we need the same in https://github.com/rhobs/observability-operator/blob/main/build/Dockerfile

PR #808, opened 2025-07-11
Merge release-1.2 into main

jgbernalp:


jgbernalp:
I'm not sure if I did something wrong because there are 0 files, but I included the changes from release-1.2

jan--f:
hmm yeah very strange indeed, let me try as well.

jan--f:
Added #819 , wdyt?

Issue #795, opened 2025-07-01
Define common storage config in ClusterObservability


pavolloffay:
The goal of this ticket is to define `storage` seciton of the CR that will work well for all supported backends - loki, tempo, prometheus. 

Current schema:

```yaml
apiVersion: observability.openshift.io/v1alpha1
kind: ClusterObservability
metadata:
  name: logging-tracing
spec:
  storage:
    secret:
      name: minio
      type: s3
  capabilities:
    logging:
      enabled: true
    tracing:
      enabled: true
```

Other requirements for the storage spec:
- [ ] The secret can be missing if only the install operators is defined

pavolloffay:
Done

PR #814, opened 2025-07-16
feat: proposal for storage configuration for clusterobservability

pavolloffay:
Updates #795


alanconway (approved):
/lgtm

jan--f:
I realize I'm a bit late but it would be nice if we can get some more reviews on PRs like this.

PR #825, opened 2025-08-04
feat: support various object storage configurations

pavolloffay:
Updates https://github.com/rhobs/observability-operator/issues/795 

Continuation of https://github.com/rhobs/observability-operator/pull/814/files 

## Run instructions
Image: docker.io/pavolloffay/observability-operator-bundle:0.0.0-1754994815-dev

`kubectl apply -f https://raw.githubusercontent.com/grafana/tempo-operator/refs/heads/main/minio.yaml`

```
./tmp/bin/operator-sdk run bundle docker.io/pavolloffay/observability-operator-bundle:0.0.0-1754994815-dev --namespace operators --install-mode AllNamespaces


kubectl create secret generic minio-secret \
--from-literal=access_key_secret="supersecret"


kubectl apply -f - <<EOF
kind: ClusterObservability
metadata:
  name: logging-tracing
spec:
  storage:
    objectStorage:
      s3:
        bucket: tempo
        endpoint: http://minio.minio.svc:9000
        accessKeyID: tempo
        accessKeySecret:
          name: minio-secret
          key: access_key_secret
        region: us-east-1
  capabilities:
    tracing:
      enabled: true
EOF
```


docs/clusterobservability.md
```diff
@@ -166,19 +166,77 @@ The various storage configuration per capability can be achieved by multiple `Cl
...
+The object storage configuration in COO follows the existing design principles of the Tempo and Loki operators.
+The `ClusterObservability` CR contains the type of the object storage and a secret holds the configuration.
+
+##### Alternative thanos-io/objstore
+
+The https://github.com/thanos-io/objstore client is used by Loki and Thanos to connect to the object storage.
+It offers a unified interface for different object storage types with a common configuration file.
+
+At the time of writing this document, the COO does not support Thanos, Loki operator does not directly support the `thanos-io/objstore`
+configuration file and the Tempo does not support it at all.
```
xperimental, line 182: Why does the support of the other operators have any impact on what COO can use as its configuration? Either the COO or the other operator needs to be modified to make use of this new configuration anyway, so just picking a single configuration format that can support all required backends should be fine (which was why objstore was suggested in the first place).

It should be possible to "skip" the step of providing the configuration in the format currently supported by, for example, the Loki Operator and directly go with objstore.

docs/clusterobservability.md
```diff
@@ -166,19 +166,77 @@ The various storage configuration per capability can be achieved by multiple `Cl
...
+##### Short lived - Amazon S3 with Security Token Service (STS)
+
+Supported by Tempo and Loki.
+
+```yaml
+kubectl create secret generic storage-secret \
+--from-literal=bucket="<BUCKET_NAME>" \
+--from-literal=role_arn="<AWS_ROLE_ARN>" \
+--from-literal=region="<AWS_REGION_YOUR_BUCKET_LIVES_IN>"
+```
```
xperimental, line 208: This does not really have something to do with how we specify the storage configuration, but I wonder why we always specify Secrets using these `kubectl create` commands instead of just providing a YAML example.

docs/clusterobservability.md
```diff
@@ -166,19 +166,77 @@ The various storage configuration per capability can be achieved by multiple `Cl
 ### Object storage types
 
 Each object storage type has its own set of required fields which are configured in the secret.
-Object storage types will be added separately, there are plans to support the all the storage types supported by the capabilities.
+There are plans to support all storage types required by the capabilities.
```
xperimental, line 169: Putting this here because I can only comment on changed lines:

> The storage section of the `ClusterObservability` CRD allows users to configure the storage for all supported observability backends.

If I am not misunderstanding this, then this approach seems to assume that you can put all three signals into a single bucket side-by-side. Have we tested this? For example, I know that Loki is currently not able to run with a prefix (subdirectory) inside a bucket.

docs/clusterobservability.md
pavolloffay: There are two approaches we can take: 

1. use `thanos-io/objstore` conf file format

* The `objstore` conf file is not supported by Loki/operator at the moment. Either we would need to wait until it is supported by Loki (and operator) or COO would translate the `objstore` conf file to the storage secret(s) supported by the operator https://loki-operator.dev/docs/object_storage.md/#aws-s3
* The `objstore` client is not used by Tempo and there are no such plans. So the COO would have translate it into the storage secret https://grafana.com/docs/tempo/latest/setup/operator/object-storage/  

2. use storage secret

* the storage secret is already supported by Tempo and Loki operators, therefore the COO makes only a minimal modification for the secret used by Tempo and Loki (because there are some differences e.g. `bucket` vs `bucketnames`)



What I would suggest is to stick with the storage secret as it is supported now, and once the COO supports Loki operator (or Thanos / Tempo) that can be configured with the `thanos-io/objstore` we can introduce the storage secret type e.g. `thanos-objstore`.

docs/clusterobservability.md
pavolloffay: That is a great comment, I will validate it. However, if there are issues there is still value in having a common configuration API for all capabilities.

alanconway: +1 - perhaps we need a common config to provide a default S3 bucket, but allow each backend to partially or completely override the default settings. E.g. one option would be to have separate buckets per back-end, another might be to use entirely separate S3 stores.

docs/clusterobservability.md
pavolloffay: I have updated the proposal based on the discussion on the arch call. Now most of the object storage configuration lives in the CR, only sensitive fields are in a secret.

pkg/apis/observability/v1alpha1/objectstorage.go
```diff
@@ -0,0 +1,119 @@
...
+	Audience string `json:"audience,omitempty"` // Optional audience for the Azure WIF
+}
+
+type GCSSpec struct {
+	// Bucket is the name of the Google Cloud Storage bucket.
+	// +kubebuilder:validation:Required
+	Bucket string `json:"bucket"`
+	// KeyFile is path to the key.json file.
+	// +kubebuilder:validation:Required
+	KeyFile string `json:"keyJSON"`
```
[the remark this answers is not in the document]
pavolloffay: I am not sure how the file is passed to the deployment. 

This property contains just the path to the file.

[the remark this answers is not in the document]
IshwarKanse: We need to fetch the keyJSON from the secret instead of adding it in CR directly. similar to what we are doing for AWS and Azure for access_key_secret and account_key. If we add the key.json to the CR, its shown in plaintext in the CR instance.
 
For example, suggestion on using key.json from secret. 
    objectStorage:
      gcs:
        bucket: ikanse-tempostack-gcs
        keyJSON:
          secretName: gcs-secret
          key: key.json

The required secret to which we add the GCP key.json:
kubectl -n $NAMESPACE create secret generic gcs-secret \
  --from-file=key.json="$GCS_KEY_FILE_PATH"

docs/clusterobservability.md
```diff
@@ -158,29 +167,202 @@ spec:
...
+#### Amazon S3 S3 / MinIO
+
+Supported by Tempo and Loki.
 
 ```yaml
-kubectl create secret generic s3 \
+spec:
+  storage:
+    objectStorage:
+      s3:
```
IshwarKanse, line 200: @pavolloffay For STS we only need bucket, roleARN and region. The rest of the fields are not required. 
% oc create -f coo.yaml 
The ClusterObservability "tracing" is invalid: 
* spec.storage.objectStorage.s3.accessKeyID: Required value
* spec.storage.objectStorage.s3.accessKeySecret: Required value
* spec.storage.objectStorage.s3.endpoint: Required value

pkg/controllers/observability/tempo_components.go
```diff
@@ -56,27 +66,90 @@ func tempoSecretName(name string) string {
...
+		err := k8sClient.Get(ctx, client.ObjectKey{
+			Namespace: operatorNamespace,
+			Name:      instance.Spec.Storage.ObjectStorageSpec.GCS.KeyJSON.SecretName,
+		}, accountKeySecret)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get GCS keyJSON secret %s: %w", instance.Spec.Storage.ObjectStorageSpec.GCS.KeyJSON.SecretName, err)
+		}
+
+		tempoSecret.Data = map[string][]byte{
+			"bucket":   []byte(instance.Spec.Storage.ObjectStorageSpec.GCS.Bucket),
```
IshwarKanse, line 147: The secret created for Tempo in the cluster-observability namespace need to have bucketname as the key instead of bucket.

pkg/controllers/observability/tempo_components.go
```diff
@@ -93,3 +183,26 @@ func uiPlugin() *uiv1alpha1.UIPlugin {
...
+		return tempov1alpha1.ObjectStorageSecretS3
+	} else if objStorage.Azure != nil || objStorage.AzureWIF != nil {
+		return tempov1alpha1.ObjectStorageSecretAzure
+	} else if objStorage.GCS != nil || objStorage.GCSSTSSpec != nil {
+		return tempov1alpha1.ObjectStorageSecretGCS
+	}
+	return ""
+}
+
+func toTempoCredentialMode(objStorage obsv1alpha1.ObjectStorage) tempov1alpha1.CredentialMode {
```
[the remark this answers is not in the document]
pavolloffay: The audience code looks also fine https://github.com/rhobs/observability-operator/blob/1313fe0d696d9fcd8fd8a63de7d05d105fa3705a/pkg/controllers/observability/tempo_components.go#L165 

Could you please share the entire CR?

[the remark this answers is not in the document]
IshwarKanse: The credential mode one was my mistake I used gcs instead of gcsSTS but we still have the secret issue. 

This is the CR:
```
kind: ClusterObservability
metadata:
  name: tracing
spec:
  storage:
    objectStorage:
      gcsSTS:
        audience: openshift
        bucket: ikanse-gcp-wif-tempo
        keyJSON:
          secretName: gcs-secret
          key: key.json
  capabilities:
    tracing:
      enabled: true
```
Created the secret in observability-operator.
```
% oc get secrets gcs-secret -n observability-operator -o yaml
apiVersion: v1
data:
  key.json: REDACTED
kind: Secret
metadata:
  creationTimestamp: "2025-08-12T03:21:27Z"
  name: gcs-secret
  namespace: observability-operator
  resourceVersion: "54870"
  uid: 0c0c6c99-ba0f-4f51-a954-d5daa0e05a2c
type: Opaque
```
The secret created in the cluster-observability namespace doesn't have any key:values added. 
```
% oc get secrets coo-tracing -n cluster-observability -o yaml
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: "2025-08-12T03:37:20Z"
  name: coo-tracing
  namespace: cluster-observability
  ownerReferences:
  - apiVersion: observability.openshift.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: ClusterObservability
    name: tracing
    uid: e141081f-bf03-406a-b45c-42e497f48e3a
  resourceVersion: "64416"
  uid: 0edab11f-f036-4107-8638-ea2c64d938ea
type: Opaque
```
Operator logs:
```
2025-08-12T03:37:24Z	ERROR	cluster-observability	Failed to reconcile	{"error": "cluster-observability/coo (tempo.grafana.com/v1alpha1, Kind=TempoStack): updater failed to patch: admission webhook \"vtempostack.tempo.grafana.com\" denied the request: TempoStack.tempo.grafana.com \"coo\" is invalid: [spec.storage.secret.name: Invalid value: \"coo-tracing\": storage secret must contain \"bucketname\" field, spec.storage.secret.name: Invalid value: \"coo-tracing\": storage secret must contain \"key.json\" field]"}
github.com/rhobs/observability-operator/pkg/controllers/observability.clusterObservabilityController.Reconcile
	/workspace/pkg/controllers/observability/observability_controller.go:127
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:340
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:300
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:202
```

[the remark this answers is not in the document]
pavolloffay: > storage secret must contain \"bucketname\" field, spec.storage.secret.name: Invalid value: \"coo-tracing\": storage secret must contain \"key.json\" field]"}

This is weird, I am setting those two fields in the secret. Could you please re-rerun with updated bundle (I rebuilt it)? 

If it does not work could you please get the secret in a way that it shows the fields?

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -64,24 +64,49 @@ type ClusterObservabilityStatus struct {
...
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret",displayName="Certificate Secret"
+	Cert string `json:"certName,omitempty"`
+
+	// MinVersion defines the minimum acceptable TLS version.
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Min TLS Version"
+	MinVersion string `json:"minVersion,omitempty"`
```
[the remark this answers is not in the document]
pavolloffay: Yes, but it's not supported by the underlying capability. Let's add it once it is supported.

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -64,24 +64,49 @@ type ClusterObservabilityStatus struct {
...
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enabled",order=1,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	Enabled bool `json:"enabled"`
+
+	// CA is the name of a ConfigMap containing a CA certificate (service-ca.crt).
+	// It needs to be in the same namespace as the Tempo custom resource.
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:ConfigMap",displayName="CA ConfigMap"
+	CA string `json:"caName,omitempty"`
```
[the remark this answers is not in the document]
pavolloffay: The same as above the struct from core does not provide any validation. 

I have defined the struct to consume the CA from a config map

pkg/apis/observability/v1alpha1/types.go
```diff
@@ -64,24 +64,49 @@ type ClusterObservabilityStatus struct {
...
+	// PersistentVolumeClaimSpec defines the persistent volume claim(s) for the capabilities that require it.
+	PersistentVolumeClaimSpec PersistentVolumeClaimSpec `json:"pvc,omitempty"`
 }
 
-// SecretSpec defines the secret for the storage.
-type SecretSpec struct {
-	// Name is the name of the secret for the storage.
-	Name string `json:"name,omitempty"`
+// TLSSpec is the TLS configuration.
+type TLSSpec struct {
```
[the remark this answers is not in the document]
pavolloffay: I would keep `TLSSpec` it aligns better with other struct names in the package

PR #830, opened 2025-08-14
feat: move the storage under capability

pavolloffay:
This PR moves the storage config under each capability. This way each capability can better control its storage configuration and make sure only supported ones are exposed.  At the same time each storage has the same schema for each capability.

* Each capability might support a different storage configuration.
* now a single CR can be used to configure multiple capabilities with a different storage config (e.g. buckets)


docs/clusterobservability.md
```diff
@@ -24,20 +24,29 @@ metadata:
...
       enabled: true
+      storage:
+        objectStorage:
+          s3:
+            bucket: tempo
+            endpoint: http://minio.minio.svc:9000
+            accessKeyID: tempo
+            accessKeySecret:
+              name: minio-secret
+              key: access_key_secret
```
jan--f, line 49: I would argue to keep the single storage spec. Having to specify per capability just makes things more complicated, especially with each capability supporting a different spec.
But perhaps I'm missing the motivation for this change?

PR #865, opened 2025-09-09
COO-1183: Add DefaultLabelSelector to cache options

jan--f:


pkg/operator/operator.go
```diff
@@ -229,16 +228,36 @@ func New(ctx context.Context, cfg *OperatorConfiguration) (*Operator, error) {
 			Metrics:                metricsOpts,
 			HealthProbeBindAddress: cfg.HealthProbeAddr,
 			PprofBindAddress:       "127.0.0.1:8083",
+			Cache: cache.Options{
+				DefaultLabelSelector: labels.SelectorFromSet(map[string]string{ctrlutil.ResourceLabel: ctrlutil.OpName}),
```
simonpasquier, line 232: should we explain the rationale for the default label selector and why there are exceptions for certain resources?

simonpasquier (commented):
/lgtm
/hold

Thanks for splitting into several commits. It was very easy to review :)
I've got only one nit but feel free to merge if you don't agree or want to follow-up.

pkg/operator/operator.go
jan--f: Very good point. added.
Additionally I added ClusterObservability and further narrowed down the Service exception.

simonpasquier (commented):
/lgtm

PR #874, opened 2025-09-17
fix: clean up UIPlugin cluster-scoped resources in COO

danielmellado:
Handle cleanup of cluster-scoped resources for UIPlugin when Health
Analyzer or Perses are enabled.

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>


zhuje:
> cc @zhuje is this similar to #873?, or can we reuse the suggestions?
Yes, I'm doing something similar directly in `components.go` with `reconciler.NewDeleter`. 
```
if !incidentsEnabled {
    components = append(components, reconciler.NewDeleter(newClusterRoleBinding(namespace, serviceAccountName, "cluster-monitoring-view", "cluster-monitoring-view")),
     ... 
 } else if pluginInfo.HealthAnalyzerImage != "" {
    components = append(components, reconciler.NewUpdater(newClusterRoleBinding(namespace, serviceAccountName, "cluster-monitoring-view", "cluster-monitoring-view"), plugin))
    ...
}
```

PR #875, opened 2025-09-17
COO-1185: fix version of prometheus-operator client

simonpasquier:


simonpasquier:
/retest-required

looks like no cluster could be provisioned :grimacing: (`{  failed to wait for the created cluster claim to become ready: timed out waiting for the condition}`).

PR #878, opened 2025-09-18
chore: add Grafana dashboard for Telemetry

simonpasquier:
This dashboard will eventually be deployed to the App-SRE Grafana instances via app-interface.


jan--f:
I anticipate some folks expecting dashboards for users in the subdir, but I don't have a better proposal either. And the readme clearly states the purpose. So
/lgtm

PR #885, opened 2025-09-24
chore: rename ClusterObservability to ObsevervabilityInstaller

pavolloffay:
The motivation for renaming is to better fit the use-case for simplifying the installation. 

image

The change was discussed on https://docs.google.com/document/d/1QGHIvzCbl6X6Nll6Ic6WOKbGQMUve8j9kE_jHlUpebI/edit?tab=t.0#heading=h.ij84uw9zov2q without strong objections. 

Vote on the name http://poll-maker.com/poll5591925x5364453B-164 

Once this PR is merged I will revert `go.mod` and `Dockerfiles` to re-pin the `api` module. Changing only the `api` module in a separate PR would not likely work as it would remove the CR form the bundle which is required by the e2e tests.


jan--f:
/approve

The prow e2e test is flaky so no worries, but the generate test shouldn't fail. Mind looking into that?

pavolloffay:
@jan--f the tests are now passing

PR #886, opened 2025-09-25
COO-1246: fix: add CreateClusterRoleBinding field to MonitoringStack

jan--f:
In a follow up I'll add the controller code.

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/types.go
+++ b/pkg/apis/monitoring/v1alpha1/types.go
@@ -52,6 +52,23 @@ const (
 	Error LogLevel = "error"
 )
 
+// +kubebuilder:validation:Enum=CreateClusterRoleBindings;NoClusterRoleBindings
+type ClusterRoleBindingPolicy string
+
+const (
+	// CreateClusterRoleBindings instructs the MonitoringStack to create the
+	// default ClusterRoleBindings if a NamespaceSelector is present. Note that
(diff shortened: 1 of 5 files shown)


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -71,6 +71,21 @@ type MonitoringStackSpec struct {
...
+	// ServiceAccount access to all namespaces, allowing the
+	// ServiceDiscovery to work across namespaces out of the
+	// box. However by impersonating this ServiceAccount a user could elevate
+	// their access in unintended ways.
+	// To avoid this set CreateClusterRoleBinding to false. Note
+	// that admins must create the needed namespaced RoleBindings manually
+	// so that endpoint discovery works as expected.
+	// +kubebuilder:default="true"
+	// +optional
+	CreateClusterRoleBinding bool `json:"createClusterRoleBinding,omitempty"`
```
simonpasquier, line 87: Kube API conventions discourage the use of boolean fields and suggest enum values instead.
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

jan--f: Thanks, good point. What do you think  of the naming?

simonpasquier (approved):
/lgtm

PR #888, opened 2025-09-30
COO-1249: feat: create ClusterRoleBIndings based on API field

jan--f:


simonpasquier (commented):
/lgtm

pkg/controllers/monitoring/monitoring-stack/components.go
```diff
@@ -52,6 +52,7 @@ func stackComponentReconcilers(
 	alertmanagerName := ms.Name + "-alertmanager"
 	additionalScrapeConfigsSecretName := ms.Name + "-self-scrape"
 	hasNsSelector := ms.Spec.NamespaceSelector != nil
+	createCRB := hasNsSelector && ms.Spec.CreateClusterRoleBindings == "CreateClusterRoleBindings"
```
simonpasquier, line 55: (nit) we could use `CreateClusterRoleBindings`?

jan--f: Yes good point! Fixed.

PR #889, opened 2025-09-30
COO-1248: fix: use apiReader for correct ui plugin registration

jgbernalp:
This PR:

- fixes the empty response when reading the cluster.console.operator.openshift.io, this is caused from reading from the cache when using the k8sClient. As the cluster plugins is read occasionally we can use the apiReader to get latest results.
- Fixes the names of some role bindings so labels are valid


zhuje:
Found out I'm unauthorized to override the test. Only [maintainers](https://github.com/rhobs/observability-operator/blob/main/OWNERS_ALIASES#L19C3-L19C14) can override.

PR #897, opened 2025-10-02
chore: re-pin API module

pavolloffay:


simonpasquier:
more a question for @jan--f  maybe but I don't remember why we don't use a local replace directive?

For instance, prometheus-operator does this:
https://github.com/prometheus-operator/prometheus-operator/blob/e289f399a93d4858083b1bfa570634908fa7d5df/go.mod#L153-L156

PR #937, opened 2025-11-21
fix: fix & rename health-analyzer and korrel8r clusterrolebindings

tremes:
There was an issue with two `cluster-monitoring-view` clusterrolebindings (one for korrel8r and second one for health analyzer) during the release. 

We quickly fixed this during the release by renaming the crb for the health analyzer. This is an attempt to improve the hotfix:
- have only one function for crb creation
- crb names are prefixed with the plugin name


jgbernalp (approved):

anpingli:
LGTM

PR #941, opened 2025-11-26
feat: allow to set replicas number for Alertmanager

simonpasquier:
This commit introduces a new `replicas` field in the Alertmanager configuration to the `MonitoringStack` CRD. This makes it consistent with the Prometheus configuration and allows to run Alertmanager on single-node clusters.

It also adds a `replace` directive in `go.mod` to use the local copy of `github.com/rhobs/observability-operator/pkg/apis` which simplifies the development of API changes.


jan--f (approved):
/lgtm
Thanks!

PR #943, opened 2025-11-27
COO-1384: fix(monitoringstack): correctly configure OTLP receiver

too-common-name:
## Context: Prometheus 3.0 changes
Starting from **Prometheus 3.0.0-beta1**, the mechanism to enable the OTLP Receiver [has changed](https://github.com/prometheus/prometheus/blob/main/CHANGELOG.md#300--2024-11-14):

- Legacy (< 3.0.0): Requires `--enable-feature=otlp-write-receiver`
- Modern (>= 3.0.0): Requires --web.enable-otlp-receiver`

The Prometheus Operator is able to handle this logic automatically if `enableOtlpHttpReceiver` and `Version` are setted. You can check it [here](https://github.com/rhobs/obo-prometheus-operator/blob/rhobs-rel-0.86.2-rhobs1/pkg/prometheus/promcfg.go)  at line 1190. 

## The Problem
The current implementation was hardcoding the legacy feature flag:
```go
EnableFeatures: []monv1.EnableFeature{"otlp-write-receiver"} // Old logic
```
This fails to enable OTLP on modern/default Prometheus versions (which require the new mechanism).

## The Fix Logic
I switched the implementation to use the proper upstream API field:
```go
EnableOTLPReceiver: config.EnableOtlpHttpReceiver
```
This enables the downstream operator to manage the configuration dynamically. **However, this introduced a regression for legacy versions.**
Because the Observability Operator was not passing the `Version` field to the `Prometheus` CR, the downstream operator defaulted to its internal version (>= 3.0.0).
- **Scenario:** A user runs image v2.45.0 (Legacy).
- **Behavior:** Operator assumes v3.0.0, sees EnableOTLPReceiver=true, and injects the modern flag (--web.enable-otlp-receiver).
- **Result:** Prometheus v2.45.0 crashes.

To address this, the PR implements a version inference strategy, in particular:
1. **Use** `EnableOTLPReceiver`: Allows the downstream operator to handle modern flags correctly.
2. **Infer Version**: The reconciler now parses the version from the `RELATED_IMAGE_PROMETHEUS` tag and explicitly sets the `Version` field in the `Prometheus` CR.
3. **Result**: The downstream operator sees `Version: "2.45.0"`, correctly falls back to the legacy feature flag, and the crash is avoided.

## Architectural Note / Limitations
This fix relies on parsing the image tag (e.g., extracting `2.45.0` from `quay.io/...:v2.45.0`).

- **Standard Usage:** Works perfectly for pinned version tags.
- **Custom Images:** If a user provides a custom image with a non-semantic tag (e.g., `:latest` or `:custom`), parsing will fail, and behavior falls back to the default (assuming 3.0.0).

_Suggestion_: To fully support custom images, we should update the `MonitoringStack` CRD to accept an explicit `spec.prometheusConfig.version` field.

### Manual verification on OCP 
1. **OTLP unsupported test:** Configured operator with Prometheus `v2.45.0`. Applied stack with OTLP enabled. The Prometheus Operator skipped the enablement since OTLP is not supported.
2. **Legacy test**: Configured operator with Prometheus `v2.47.0`. Applied stack with OTLP enabled. Pod started successfully. Downstream operator applied legacy feature flag.
3. **Modern test**: Configured operator with default image. Applied stack with OTLP enabled. Pod started successfully. Downstream operator applied modern argument.
### Tests
- Added E2E test `Assert_OTLP_receiver_flag_is_set_when_enabled_in_CR` to verify the happy path (Default/Modern Version).
- Updated `managed_fields` expectations to reflect that `enableOTLPReceiver` is now managed field.

## Checklist 
[x] Fixes OTLP configuration for modern Prometheus (v3.0+)
[x] Fixes regression/crash for legacy Prometheus (v2.x)
[x] Verified manually on cluster
[x] Updated E2E tests
[x] Ran `make lint`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,8 +44,6 @@ replace (
 	github.com/rhobs/observability-operator/pkg/apis => ./pkg/apis
 )
 
-require github.com/blang/semver/v4 v4.0.0
-
 require (
 	cel.dev/expr v0.24.0 // indirect
 	dario.cat/mergo v1.0.2 // indirect
@@ -54,6 +52,7 @@ require (
 	github.com/antlr4-go/antlr/v4 v4.13.1 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/brunoga/deep v1.2.5 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
--- a/pkg/controllers/monitoring/monitoring-stack/components.go
+++ b/pkg/controllers/monitoring/monitoring-stack/components.go
@@ -4,9 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"reflect"
-	"strings"
 
-	"github.com/blang/semver/v4"
 	monv1 "github.com/rhobs/obo-prometheus-operator/pkg/apis/monitoring/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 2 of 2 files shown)


simonpasquier:
We don't support custom images.

too-common-name:
Ok, I will remove the version inference. Thank you Simon!

simonpasquier (commented):
/lgtm

I think that for a better coverage, we should exercise the OTLP endpoint in the e2e tests but I can take in a follow-up PR.

simonpasquier:
/retitle [COO-1384](https://issues.redhat.com//browse/COO-1384): fix(monitoringstack): correctly configure OTLP receiver

PR #946, opened 2025-11-28
feat: support size-based retention for MonitoringStack

### LINKED ISSUE #890: [RFE] Allow to specifiy retention Size
Hello team,

I want to configure prometheus retention size with `monitoring.rhobs_monitoringstacks` CRD but there is no possibility to do it, can you add it ?

Appreciated

Regards

simonpasquier:
This commit exposes a new `retentionSize` field in the MonitoringStack CRD which mimics the API already present in the Prometheus CRD.

Closes #890

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/types.go
+++ b/pkg/apis/monitoring/v1alpha1/types.go
@@ -103,10 +103,24 @@ type MonitoringStackSpec struct {
 	// +optional
 	CreateClusterRoleBindings ClusterRoleBindingPolicy `json:"createClusterRoleBindings,omitempty"`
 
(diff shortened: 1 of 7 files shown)


pkg/apis/monitoring/v1alpha1/types.go
```diff
@@ -108,6 +108,10 @@ type MonitoringStackSpec struct {
 	// +kubebuilder:default="120h"
 	Retention monv1.Duration `json:"retention,omitempty"`
 
+	// retentionSize defines the maximum number of bytes used by the Prometheus data.
```
jan--f: Should this perhaps mention how `Retention` and `RetentionSize` interact if both are present?

PR #950, opened 2025-12-02
chore: bump p-o to v0.87.0

jan--f:


jan--f:
> * Scheme: ptr.To(monv1.SchemeHTTP),
> 
> * Scheme: ptr.To(monv1.Scheme("http")),

Afaiu using the contstant `monv1.SchemeHTTP` should work the same :thinking:

simonpasquier:
I think that only the v0.87.0 CRDs will support both "http" and "HTTP". Older CRDs only allow "http".

PR #994, opened 2026-02-10
COO-1597: Raise memory limit of observability-operator deploy

muellerfabi:
On a larger OCP 4.18 cluster with 66 nodes observability-operator 1.3.0 gets OOMKilled right after start.

Issue is reported in https://access.redhat.com/support/cases/#/case/04368491
Issue is tracked in https://issues.redhat.com/browse/COO-1597

**Update** due to missing information about the actual issue:
We are aware of the fact that it is possible to set resource requests and limits in the subscription. We use the following as a workaround:
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: cluster-observability-operator
  namespace: openshift-operators-redhat
spec:
  channel: stable
  config:
    resources:
      limits:
        memory: 3Gi
     requests:
       memory: 500Mi
```

The problem is that the given config is set on all COO components equally:
- obo-prometheus-operator
- obo-prometheus-operator-admission-webhook
- observability-operator
- perses-operator

Because of the elevated memory usage of perses in contrast to the other components, we were forced to set the memory limit to 3Gi.
Now all the components have a needlessly too high memory limit:
```bash
$ oc get deploy -o custom-columns=NAME:.metadata.name,RESOURCES:.spec.template.spec.containers[0].resources
NAME                                        RESOURCES
logging                                     map[]
loki-operator-controller-manager            map[]
obo-prometheus-operator                     map[limits:map[memory:3Gi] requests:map[memory:500Mi]]
obo-prometheus-operator-admission-webhook   map[limits:map[memory:3Gi] requests:map[memory:500Mi]]
observability-operator                      map[limits:map[memory:3Gi] requests:map[memory:500Mi]]
perses-operator                             map[limits:map[memory:3Gi] requests:map[memory:500Mi]]


$ oc adm top po
NAME                                                        CPU(cores)   MEMORY(bytes)   
logging-7c8b5bfdf4-v5p67                                    1m           23Mi            
loki-operator-controller-manager-7c5b4ffbfb-wvt9b           49m          5622Mi          
obo-prometheus-operator-545cdc864f-wxmzh                    61m          317Mi           
obo-prometheus-operator-admission-webhook-57d54bf6d-4p6j6   1m           11Mi            
obo-prometheus-operator-admission-webhook-57d54bf6d-x5pfh   1m           12Mi            
observability-operator-595c984dfb-24lsc                     3m           536Mi           
perses-operator-5fc9687477-8g9jc                            11m          1701Mi          
```

### DIFF
--- a/bundle/manifests/observability-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/observability-operator.clusterserviceversion.yaml
@@ -953,9 +953,7 @@ spec:
                 - containerPort: 8080
                   name: http
                 resources:
-                  limits:
-                    cpu: 100m
-                    memory: 500Mi
+                  limits: {}
                   requests:
                     cpu: 5m
                     memory: 150Mi
@@ -1026,9 +1024,7 @@ spec:
                 - containerPort: 8443
                   name: https
                 resources:
-                  limits:
-                    cpu: 200m
-                    memory: 200Mi
+                  limits: {}
                   requests:
                     cpu: 50m
                     memory: 50Mi
@@ -1095,9 +1091,7 @@ spec:
                     path: /healthz
                     port: 8081
                 resources:
-                  limits:
-                    cpu: 400m
-                    memory: 512Mi
+                  limits: {}
                   requests:
                     cpu: 100m
                     memory: 256Mi
@@ -1169,9 +1163,7 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 10
                 resources:
-                  limits:
-                    cpu: 500m
-                    memory: 512Mi
+                  limits: {}
                   requests:
                     cpu: 100m
                     memory: 128Mi
--- a/deploy/dependencies/kustomization.yaml
+++ b/deploy/dependencies/kustomization.yaml
@@ -88,9 +88,7 @@ patches:
                 requests:
                   cpu: 5m
                   memory: 150Mi
-                limits:
-                  cpu: 100m
-                  memory: 500Mi
+                limits: {}
               terminationMessagePolicy: FallbackToLogsOnError
           securityContext:
             runAsNonRoot: true
--- a/deploy/olm/syncselector-template.yaml
+++ b/deploy/olm/syncselector-template.yaml
@@ -7,15 +7,9 @@ parameters:
   - name: CHANNEL
     value: development
     required: true
-  - name: RESOURCE_LIMIT_CPU
-    required: true
-    value: 200m
   - name: RESOURCE_REQUEST_CPU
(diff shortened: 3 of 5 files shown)


jan--f:
@jgbernalp wdyt? It's a small adjustment, so not sure how many user this will help out of the box.

jgbernalp:
> @jgbernalp wdyt? It's a small adjustment, so not sure how many user this will help out of the box.

Is there a way for users to configure these limits? this might apply for this case.  @muellerfabi do you have a pprof from this cluster that we can analyze?

simonpasquier:
@jgbernalp yes it's possible (https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#resources) but the original report complained that it "should" just work out of the box.

duritong:
@simonpasquier the main issue with: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#resources is that it overwrites the resource constraints globally and thus for example if you have a deployment with the actual operator and for example kube-rbac-proxy sidecar (very typical), your large requirements will also apply there, see https://github.com/redhat-cop/patch-operator/issues/76 as an example from another operator.

muellerfabi:
I updated the initial comment with further information, why overwriting resources in the Subscription kind is not ideal.

> @jgbernalp  Is there a way for users to configure these limits? this might apply for this case. @muellerfabi do you have a pprof from this cluster that we can analyze?
A must-gather from the affected cluster could be uploaded to the support case, if that helps?

Raising the memory limit to 550Mi is sufficient until the cluster grows to 90 nodes or so.
Maybe it was better to drop the limit entirely. Cluster-Operators usually do not have limits.
WDYT?

simonpasquier:
> Raising the memory limit to 550Mi is sufficient until the cluster grows to 90 nodes or so.

I'm worried that while it works for this cluster, we'll hear about other clusters still hitting the limit.

> Maybe it was better to drop the limit entirely. Cluster-Operators usually do not have limits.
> WDYT?

My initial assumption was that resource limits were a strong requirement but apparently not: https://sdk.operatorframework.io/docs/best-practices/managing-resources/#general-guidelines
We'd need to discuss the downsides of removing the limits but it might be the best course of action.

muellerfabi:
I updated my commit deleting resource limits for the components.

KCS article https://access.redhat.com/solutions/6958023 states, beside some rationale:
> Therefore, cluster components SHOULD NOT be configured with resource limits.
>
> However, cluster components MUST declare resource requests for both CPU and memory.

PR #1031, opened 2026-03-10
COO-1699:chore: migrate to controller-runtime client.Apply over client.Patch

jan--f:


pkg/reconciler/reconciler.go
```diff
@@ -124,3 +124,44 @@ func NewOptionalUnmanagedUpdater(r client.Object, c metav1.Object, cond bool) Re
...
+	k := a.obj.GetObjectKind().GroupVersionKind().Kind
+	return &k
+}
+
+func (a *clientObjectApplyConfig) GetAPIVersion() *string {
+	av := a.obj.GetObjectKind().GroupVersionKind().GroupVersion().String()
+	return &av
+}
+
+func (a *clientObjectApplyConfig) MarshalJSON() ([]byte, error) {
```
jgbernalp, line 167: Should also unmarshalling be delegated to `a`?

jan--f: Yes indeed, good catch!

PR #1038, opened 2026-03-16
OBSINTA-1219: add e2e tests for UIPlugin incident detection

DavidRajnoha:
Add end-to-end tests that validate the monitoring UIPlugin with cluster-health-analyzer: a deployment readiness check and a functional test that triggers a CrashLoopBackOff alert and verifies the cluster_health_components_map incident metric is produced.

Also introduce AssertPromQLResultWithOptions to allow callers to override the default poll interval and timeout, and generalize waitForDBUIPluginDeletion to waitForUIPluginDeletion.

### DIFF
--- a/test/e2e/framework/assertions.go
+++ b/test/e2e/framework/assertions.go
@@ -423,11 +423,26 @@ func (f *Framework) GetPodMetrics(pod *v1.Pod, opts ...func(*HTTPOptions)) ([]by
 // the callback function for additional checks.
 func (f *Framework) AssertPromQLResult(t *testing.T, expr string, callback func(model.Value) error) error {
 	t.Helper()
+	return f.AssertPromQLResultWithOptions(t, expr, callback)
+}
+
+// AssertPromQLResultWithOptions is like AssertPromQLResult but accepts
+// WithTimeout and WithPollInterval options to override the default polling
+// parameters.
+func (f *Framework) AssertPromQLResultWithOptions(t *testing.T, expr string, callback func(model.Value) error, fns ...OptionFn) error {
+	t.Helper()
+	option := AssertOption{
+		PollInterval: 20 * time.Second,
+		WaitTimeout:  3 * DefaultTestTimeout,
+	}
+	for _, fn := range fns {
+		fn(&option)
+	}
 	var (
 		pollErr error
 		v       model.Value
 	)
-	if err := wait.PollUntilContextTimeout(context.Background(), 20*time.Second, 3*DefaultTestTimeout, true, func(context.Context) (bool, error) {
+	if err := wait.PollUntilContextTimeout(context.Background(), option.PollInterval, option.WaitTimeout, true, func(context.Context) (bool, error) {
 		v, pollErr = f.getPromQLResult(context.Background(), expr)
 		if pollErr != nil {
 			t.Logf("error from getPromQLResult(): %s", pollErr)
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -14,6 +14,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -25,6 +26,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -256,4 +258,102 @@ func (f *Framework) CleanUp(t *testing.T, cleanupFunc func()) {
 			cleanupFunc()
 		}
 	})
+}
+
+// SkipIfClusterVersionBelow skips the test if the cluster version is below
+// minVersion. The minVersion string should be a semver-compatible version
+// (e.g. "4.19" or "v4.19").
+func (f *Framework) SkipIfClusterVersionBelow(t *testing.T, minVersion string) {
+	t.Helper()
+	cv := &configv1.ClusterVersion{}
+	err := f.K8sClient.Get(t.Context(), client.ObjectKey{Name: "version"}, cv)
+	if err != nil {
+		t.Fatalf("failed to determine cluster version: %v", err)
+		return
+	}
+
+	actual := cv.Status.Desired.Version
+	if actual == "" {
+		t.Fatal("cluster version is empty")
+		return
+	}
+	t.Logf("Detected cluster version: %s", actual)
+
+	if !strings.HasPrefix(actual, "v") {
+		actual = "v" + actual
+	}
+	if !strings.HasPrefix(minVersion, "v") {
+		minVersion = "v" + minVersion
+	}
+
+	canonicalActual := fmt.Sprintf("%s-0", semver.Canonical(actual))
+	canonicalMin := fmt.Sprintf("%s-0", semver.Canonical(minVersion))
+
+	if semver.Canonical(actual) == "" || semver.Canonical(minVersion) == "" {
+		t.Fatalf("Unable to parse version (actual=%q, min=%q)", actual, minVersion)
+		return
+	}
+
+	if semver.Compare(canonicalActual, canonicalMin) < 0 {
+		t.Skipf("Skipping: cluster version %s is below minimum required %s", cv.Status.Desired.Version, minVersion)
+	}
+}
+
+// DumpNamespaceDebug logs deployments (with conditions), pods (with container
+// statuses), and events for the given namespace. Useful as a t.Cleanup or
+// on-failure diagnostic helper.
+func (f *Framework) DumpNamespaceDebug(t *testing.T, namespace string) {
+	t.Helper()
+	ctx := context.WithoutCancel(t.Context())
+
+	t.Log("=== BEGIN DEBUG DUMP ===")
+	defer t.Log("=== END DEBUG DUMP ===")
+
+	var deployments appsv1.DeploymentList
+	if err := f.K8sClient.List(ctx, &deployments, client.InNamespace(namespace)); err != nil {
+		t.Logf("Failed to list deployments in %s: %v", namespace, err)
+	} else {
+		t.Logf("Deployments in namespace %s: %d", namespace, len(deployments.Items))
+		for _, d := range deployments.Items {
+			t.Logf("  Deployment: name=%s replicas=%d readyReplicas=%d availableReplicas=%d",
+				d.Name, ptr.Deref(d.Spec.Replicas, 0), d.Status.ReadyReplicas, d.Status.AvailableReplicas)
+			for _, c := range d.Status.Conditions {
+				t.Logf("    condition: type=%s status=%s reason=%s message=%s",
+					c.Type, c.Status, c.Reason, c.Message)
+			}
+		}
+	}
+
+	var pods corev1.PodList
+	if err := f.K8sClient.List(ctx, &pods, client.InNamespace(namespace)); err != nil {
+		t.Logf("Failed to list pods in %s: %v", namespace, err)
+	} else {
+		t.Logf("Pods in namespace %s: %d", namespace, len(pods.Items))
+		for _, p := range pods.Items {
+			t.Logf("  Pod: name=%s phase=%s", p.Name, p.Status.Phase)
+			for _, cs := range p.Status.ContainerStatuses {
+				switch {
+				case cs.State.Running != nil:
+					t.Logf("    container=%s ready=%v restarts=%d state=Running", cs.Name, cs.Ready, cs.RestartCount)
+				case cs.State.Waiting != nil:
+					t.Logf("    container=%s ready=%v restarts=%d state=Waiting reason=%s message=%s",
+						cs.Name, cs.Ready, cs.RestartCount, cs.State.Waiting.Reason, cs.State.Waiting.Message)
+				case cs.State.Terminated != nil:
+					t.Logf("    container=%s ready=%v restarts=%d state=Terminated reason=%s exitCode=%d",
+						cs.Name, cs.Ready, cs.RestartCount, cs.State.Terminated.Reason, cs.State.Terminated.ExitCode)
+				}
+			}
+		}
+	}
+
+	var events corev1.EventList
+	if err := f.K8sClient.List(ctx, &events, client.InNamespace(namespace)); err != nil {
+		t.Logf("Failed to list events in %s: %v", namespace, err)
+	} else {
+		t.Logf("Events in namespace %s: %d", namespace, len(events.Items))
+		for _, e := range events.Items {
+			t.Logf("  Event: involvedObject=%s/%s reason=%s message=%s type=%s count=%d",
+				e.InvolvedObject.Kind, e.InvolvedObject.Name, e.Reason, e.Message, e.Type, e.Count)
+		}
+	}
 }
--- a/test/e2e/uiplugin_cluster_health_analyzer_test.go
+++ b/test/e2e/uiplugin_cluster_health_analyzer_test.go
@@ -1,0 +1,221 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"testing"
+	"time"
+
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus/common/model"
+	"gotest.tools/v3/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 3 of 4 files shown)


test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+	f.GetResourceWithRetry(t, healthAnalyzerDeploymentName, uiPluginInstallNS, &haDeployment)
+	f.AssertDeploymentReady(healthAnalyzerDeploymentName, uiPluginInstallNS, framework.WithTimeout(5*time.Minute))(t)
+
+	suffix := strconv.FormatInt(time.Now().UnixNano()%100000, 10)
+	ruleName := "e2e-crashloop-" + suffix
+	alertName := "E2ECrashLoop" + suffix
+	deployName := "e2e-crasher-" + suffix
+
+	rule := newCrashLoopRule(t, ruleName, alertName, deployName)
+	err = f.K8sClient.Create(context.Background(), rule)
```
tremes: nit: you can use `t.Context()`

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+	healthAnalyzerDeploymentName = "health-analyzer"
+	prometheusRuleNamespace      = "openshift-monitoring"
+)
+
+func clusterHealthAnalyzer(t *testing.T) {
+	skipIfClusterVersionBelow(t, "4.19")
+
+	monv1.AddToScheme(f.K8sClient.Scheme())
+
+	plugin := newMonitoringUIPlugin(t)
```
tremes: Maybe it's enabled in the tests, but I think this requires the [OpenShift feature gate](https://github.com/rhobs/observability-operator/blob/main/pkg/operator/operator.go#L50) to be enabled.

DavidRajnoha: It is enabled [here](https://github.com/rhobs/observability-operator/blob/cbd6ba3c7706a20b99340e91a3aaf9654da9ab99/test/run-e2e-ocp.sh#L54) when running the tests.

tremes:
Thank you!
/approve
/lgtm

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+	t.Cleanup(func() {
+		if t.Failed() {
+			dumpClusterHealthAnalyzerDebug(t, plugin.Name)
+		}
+	})
+
+	t.Log("Waiting for health-analyzer deployment to become ready...")
+	haDeployment := appsv1.Deployment{}
+	f.GetResourceWithRetry(t, healthAnalyzerDeploymentName, uiPluginInstallNS, &haDeployment)
```
simonpasquier, line 47: (nit) do we really to get the resource since we check that it's ready just after?

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+		},
+		Spec: monv1.PrometheusRuleSpec{
+			Groups: []monv1.RuleGroup{{
+				Name: "crashloop-test-" + ruleName,
+				Rules: []monv1.Rule{{
+					Alert: alertName,
+					Expr: intstr.FromString(fmt.Sprintf(
+						`max_over_time(kube_pod_container_status_waiting_reason{reason="CrashLoopBackOff", namespace="%s", pod=~"%s.*", job="kube-state-metrics"}[5m]) >= 1`,
+						e2eTestNamespace, podPrefix)),
+					For:    ptr.To(monv1.Duration("1m")),
```
simonpasquier: removing the for clause would speed up the test.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func ptrInt32(p *int32) int32 {
+	if p == nil {
+		return 0
+	}
+	return *p
+}
+
+func skipIfClusterVersionBelow(t *testing.T, minVersion string) {
```
simonpasquier: (nit) could it be a framework function?

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+		for _, e := range events.Items {
+			t.Logf("  Event: involvedObject=%s/%s reason=%s message=%s type=%s count=%d",
+				e.InvolvedObject.Kind, e.InvolvedObject.Name, e.Reason, e.Message, e.Type, e.Count)
+		}
+	}
+
+	t.Log("=== END DEBUG DUMP ===")
+}
+
+func ptrInt32(p *int32) int32 {
```
simonpasquier: (nit) ptr.To() already exists

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+			}
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("pod with label app=%s in %s never entered CrashLoopBackOff: %v", deploymentName, namespace, err)
+	}
+}
+
+func dumpClusterHealthAnalyzerDebug(t *testing.T, pluginName string) {
```
simonpasquier, line 182: eventually this should be generalized for reuse across different tests.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+			}
+			return nil
+		},
+		framework.WithPollInterval(30*time.Second),
+		framework.WithTimeout(10*time.Minute),
+	)
+	assert.NilError(t, err, "alert %s never fired", alertName)
+
+	t.Log("Waiting for cluster-health-analyzer to expose incident metric...")
+	incidentQuery := fmt.Sprintf(`cluster_health_components_map{src_alertname="%s"}`, alertName)
```
simonpasquier: apart from checking the metric generated by the cluster health analyzer, is there any other outcome that we should verify?

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+	err = f.AssertPromQLResultWithOptions(t, incidentQuery,
+		func(v model.Value) error {
+			vec, ok := v.(model.Vector)
+			if !ok || len(vec) == 0 {
+				return fmt.Errorf("expected incident metric, got: %v", v)
+			}
+			for _, sample := range vec {
+				if string(sample.Metric["src_alertname"]) != alertName {
```
simonpasquier: already checked by the PromQL expression

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,372 @@
...
+					return fmt.Errorf("expected src_alertname=%s, got %s", alertName, sample.Metric["src_alertname"])
+				}
+				if string(sample.Metric["src_severity"]) != "warning" {
+					return fmt.Errorf("expected src_severity=warning, got %s", sample.Metric["src_severity"])
```
simonpasquier: could also be checked in the promql expression

simonpasquier (commented):
I've checked the test output and it adds 3m30s to the total duration which is not negligible. Do we really need to deploy a crashlooping pod? Can't we write a "dummy" alerting rule with a static expression (`vector(1)`) and hardcoded labels?

test/e2e/uiplugin_cluster_health_analyzer_test.go
DavidRajnoha: The getResourceWithRetry has a shorter timeout and provides explicit error message in case when the deployment is never created.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,191 @@
...
+		waitForUIPluginDeletion(existing)
+	} else if !errors.IsNotFound(err) {
+		t.Fatalf("failed to check for existing UIPlugin: %v", err)
+	}
+
+	f.CleanUp(t, func() {
+		f.K8sClient.Delete(t.Context(), plugin)
+		waitForUIPluginDeletion(plugin)
+	})
+	return plugin
```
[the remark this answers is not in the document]
DavidRajnoha: The delete and recreation is intentional to ensure a fresh install.

DavidRajnoha:
@simonpasquier I updated the test to use a static rule. For reference, here is the current timing breakdown of the test:

| Step | Duration | Elapsed |
|---|---|---|
| UIPlugin creation | 824ms | 0.8s |
| Deployment ready | 5.4s | 6.2s |
| PrometheusRule creation | 239ms | 6.4s |
| Alert firing (PromQL poll) | 32.2s | 38.6s |
| Incident metric (PromQL poll) | 32.0s | 1m 10.6s |

test/e2e/framework/assertions.go
```diff
@@ -422,12 +422,26 @@ func (f *Framework) GetPodMetrics(pod *v1.Pod, opts ...func(*HTTPOptions)) ([]by
 // It returns an error if the request fails. Otherwise the result is passed to
 // the callback function for additional checks.
 func (f *Framework) AssertPromQLResult(t *testing.T, expr string, callback func(model.Value) error) error {
+	return f.AssertPromQLResultWithOptions(t, expr, callback)
```
simonpasquier, line 426:
```suggestion
	t.Helper()
	return f.AssertPromQLResultWithOptions(t, expr, callback)
```

test/e2e/framework/framework.go
```diff
@@ -257,3 +259,96 @@ func (f *Framework) CleanUp(t *testing.T, cleanupFunc func()) {
 		}
 	})
 }
+
+// SkipIfClusterVersionBelow skips the test if the cluster version is below
+// minVersion. The minVersion string should be a semver-compatible version
+// (e.g. "4.19" or "v4.19").
+func (f *Framework) SkipIfClusterVersionBelow(t *testing.T, minVersion string) {
```
simonpasquier, line 266: in general, I'd rather skip only when we're sure that we don't meet the min version criteria. Any unhandled situation should default to test failure.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,200 @@
...
+			dumpClusterHealthAnalyzerDebug(t, plugin.Name)
+		}
+	})
+
+	t.Log("Waiting for health-analyzer deployment to become ready...")
+	haDeployment := appsv1.Deployment{}
+	f.GetResourceWithRetry(t, healthAnalyzerDeploymentName, uiPluginInstallNS, &haDeployment)
+	f.AssertDeploymentReady(healthAnalyzerDeploymentName, uiPluginInstallNS, framework.WithTimeout(5*time.Minute))(t)
+
+	suffix := strconv.FormatInt(time.Now().UnixNano()%100000, 10)
```
simonpasquier, line 50: is this to ensure that we can run the test multiple times locally and not hit conflicts?

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+			return nil
+		},
+		framework.WithPollInterval(30*time.Second),
+		framework.WithTimeout(15*time.Minute),
+	)
+	assert.NilError(t, err, "incident metric for %s never appeared", alertName)
+}
+
+func newMonitoringUIPlugin(t *testing.T) *uiv1.UIPlugin {
```
simonpasquier: The name is a bit misleading since it will delete the plugin if one already exists. IIUC and given that UI plugin resources are singletons, we can't test a given plugin in parallel. It might be better to enforce this invariant in the structure of `TestUIPlugin` but it can be a follow-up exercise.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	uiv1 "github.com/rhobs/observability-operator/pkg/apis/uiplugin/v1alpha1"
+	"github.com/rhobs/observability-operator/test/e2e/framework"
+)
+
+const (
+	healthAnalyzerDeploymentName = "health-analyzer"
+	prometheusRuleNamespace      = "openshift-monitoring"
```
simonpasquier: (nit) I'd rather use a throw-away namespace than openshift-monitoring.

simonpasquier (commented):
Most of my comments are non-blocking and can be addressed in a follow-up (my main concern is the failure mode if/when the version can't be determined).

test/e2e/uiplugin_cluster_health_analyzer_test.go
DavidRajnoha: Yes, to ensure that potential leftovers from prior tests would not cause any potential conflicts.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+
+	t.Log("Waiting for health-analyzer deployment to become ready...")
+	haDeployment := appsv1.Deployment{}
+	f.GetResourceWithRetry(t, healthAnalyzerDeploymentName, uiPluginInstallNS, &haDeployment)
+	f.AssertDeploymentReady(healthAnalyzerDeploymentName, uiPluginInstallNS, framework.WithTimeout(5*time.Minute))(t)
+
+	suffix := strconv.FormatInt(time.Now().UnixNano()%100000, 10)
+	ruleName := "e2e-health-analyzer-" + suffix
+	alertName := "E2EHealthAnalyzer" + suffix
```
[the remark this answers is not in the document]
DavidRajnoha: 5 digits is more then enough for our usecase, it is expected that in majority of cases all leftover resources would be deleted.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	uiv1 "github.com/rhobs/observability-operator/pkg/apis/uiplugin/v1alpha1"
+	"github.com/rhobs/observability-operator/test/e2e/framework"
+)
+
+const (
+	healthAnalyzerDeploymentName = "health-analyzer"
+	prometheusRuleNamespace      = "e2e-health-analyzer-rules"
+)
```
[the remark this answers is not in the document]
DavidRajnoha: If the namespace already exists, it is reused by the tests. Even though the tests are not intended to run in parallel, the resources within the namespace have unique names so the collisions would not happen there and the cluster would not be polluted by orphaned namespaces in case of multiple aborted runs or runs with --no-uninstall flag, which would leave existing namespace there.

test/e2e/uiplugin_cluster_health_analyzer_test.go
DavidRajnoha: Extracted into DumpNamespaceDebuge, here are left the component specific diagnostics. Some of then might be usefull also for other UI plugins, but I would focus on them once we would decide to reuse this in the other tests.

test/e2e/framework/framework.go
```diff
@@ -257,3 +259,101 @@ func (f *Framework) CleanUp(t *testing.T, cleanupFunc func()) {
...
+
+	if semver.Compare(canonicalActual, canonicalMin) < 0 {
+		t.Skipf("Skipping: cluster version %s is below minimum required %s", cv.Status.Desired.Version, minVersion)
+	}
+}
+
+// DumpNamespaceDebug logs deployments (with conditions), pods (with container
+// statuses), and events for the given namespace. Useful as a t.Cleanup or
+// on-failure diagnostic helper.
+func (f *Framework) DumpNamespaceDebug(t *testing.T, namespace string) {
```
simonpasquier, line 305: I'd like if we have a follow-up Jira ticket or PR to generalize the gathering of post-test data to help with troubleshooting.

test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+func clusterHealthAnalyzer(t *testing.T) {
+	f.SkipIfClusterVersionBelow(t, "4.19")
```
simonpasquier, line 30: (nit) a follow-up change could collect the cluster version before in the setup phase so that we don't have to query ClusterVersion multiple times (and avoid failures if the API is flaky).

test/e2e/uiplugin_cluster_health_analyzer_test.go
simonpasquier: can you add a comment about this?

simonpasquier (commented):
/lgtm
/hold

Feel free to unblock and address my non-blocking remarks in a follow-up PR.

DavidRajnoha:
Followup work tracked in https://github.com/rhobs/observability-operator/pull/1071 and https://github.com/rhobs/observability-operator/pull/1072

PR #1041, opened 2026-03-25
Add support for TLS profiles

ozzywalsh:
# Changes
* Retrieve cluster TLS security profile on operator startup
* Restart operator when cluster TLS profile changes

# Test Plan
* Manually tested locally; watch is configured; operator restarts on `ApiServer` change.

# Reference
Similar PR in `opentelemetry-operator` https://github.com/open-telemetry/opentelemetry-operator/pull/4669


go.mod
```diff
@@ -35,15 +35,20 @@ require (
...
+	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
+	sigs.k8s.io/controller-runtime v0.22.5
 )
 
 replace (
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20240404200104-96ed2d49b255
 	github.com/rhobs/observability-operator/pkg/apis => ./pkg/apis
 )
 
+require (
```
simonpasquier: (nit) can we move this to the top-level require section?

pkg/operator/operator.go
```diff
@@ -297,6 +317,35 @@ func New(ctx context.Context, cfg *OperatorConfiguration) (*Operator, error) {
 	}
 
 	if cfg.FeatureGates.OpenShift.Enabled {
+		setupLog := ctrl.Log.WithName("setup")
+
+		// Create a temporary client to fetch initial TLS profile before manager starts.
```
simonpasquier: question: isn't the manager's client ready to use at this stage?

simonpasquier (commented):
as commented over Slack, I'd prefer this PR to focus only on watching the TLS profile. Propagation to UI plugins should be a follow-up.

pkg/operator/operator.go
ozzywalsh: Good catch; i've updated it to use the manager's client.

go.mod
ozzywalsh: done!

pkg/operator/operator.go
```diff
@@ -297,6 +317,27 @@ func New(ctx context.Context, cfg *OperatorConfiguration) (*Operator, error) {
 	}
 
 	if cfg.FeatureGates.OpenShift.Enabled {
+		setupLog := ctrl.Log.WithName("setup")
+
+		initialTLSProfileSpec, err := openshifttls.FetchAPIServerTLSProfile(ctx, mgr.GetClient())
```
simonpasquier: Can any authenticated user read the TLS profile? I'd expect that we need to grant some permission to the operator.

zhuje:
@jan--f  I think this PR address https://redhat.atlassian.net/browse/COO-1548 Add a controller to watch the API server configuration

simonpasquier:
no need to retest, the operator fails to start due to the reason already highlighted in https://github.com/rhobs/observability-operator/pull/1041#discussion_r3028823832

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/rhobs_observability-operator/1041/pull-ci-rhobs-observability-operator-main-observability-operator-e2e/2041587447938158592/artifacts/observability-operator-e2e/gather-extra/artifacts/pods/coo_observability-operator-54fd7bb799-52tvb_operator.log

pkg/operator/operator.go
ozzywalsh: @simonpasquier 
It seems we do need to set up our own client after all; during manual testing I discovered that using the manager's client will throw this error on startup. Using a temp client to fetch fixes it.

```
2026-04-13T12:56:27Z    ERROR   setup   cannot create a new operator    {"error": "failed to fetch TLS profile from cluster: failed to get APIServer \"/cluster\": the cache is not started, can not read objects"}
main.main
        /workspace/cmd/operator/main.go:160
runtime.main
        /usr/local/go/src/runtime/proc.go:285
```

pkg/operator/operator.go
ozzywalsh: I've added the permissions.

https://github.com/rhobs/observability-operator/pull/1041/changes#diff-cd8b8b006e2ce49cae44ccfc080b9b00b8ab7e55b2fca66a304ad13f57980f56

https://github.com/rhobs/observability-operator/pull/1041/changes#diff-962e0c38ff15a8039b9d5199376b0af16a9c4148bc3ae5bccc0ebf0df354755d

ozzywalsh:
> no need to retest, the operator fails to start due to the reason already highlighted in [#1041 (comment)](https://github.com/rhobs/observability-operator/pull/1041#discussion_r3028823832)
> 
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/rhobs_observability-operator/1041/pull-ci-rhobs-observability-operator-main-observability-operator-e2e/2041587447938158592/artifacts/observability-operator-e2e/gather-extra/artifacts/pods/coo_observability-operator-54fd7bb799-52tvb_operator.log

The RBAC for apiservers was added in ed95a8e. The e2e failure was actually a different issue; the manager's cached client can't be used before the cache starts. That's fixed now in a41d1bc (using a non-cached client).   
 
Should be good to retest.

deploy/operator/observability-operator-cluster-role.yaml
simonpasquier, line 1: these changes aren't up-to-date.

simonpasquier: to be more precise, the file is generated from kubebuilder annotations and can't be edited manually.

ozzywalsh: I regenerated and pushed them.

IshwarKanse:
There is an ongoing issue with cluster pools due to which the cluster is not being provisioned/claimed in the CI test jobs. I'll run the tests locally as well as test the PR. Will get back to you once the testing is completed.

IshwarKanse:
## Bug: TLS profile not applied to the operator's metrics endpoint

The `SecurityProfileWatcher` correctly detects TLS profile changes on the `APIServer` CR and triggers a graceful restart. However, the fetched TLS profile (ciphers, `minTLSVersion`) is **never applied** to the operator's own metrics serving endpoint.

### What happens today

In `pkg/operator/operator.go`, the metrics server TLS config is set up with only the dynamic serving certificate controller:

```go
metricsOpts.TLSOpts = []func(*tls.Config){
    func(c *tls.Config) {
        c.GetConfigForClient = servingCertController.GetConfigForClient
    },
}
```

The `initialTLSProfileSpec` returned by `openshifttls.FetchAPIServerTLSProfile()` is passed to the `SecurityProfileWatcher` solely for **change detection** — it is not used to configure cipher suites or minimum TLS version on the serving endpoint.

### What should happen

The `controller-runtime-common` library already provides [`NewTLSConfigFromProfile()`](https://github.com/openshift/controller-runtime-common/blob/main/pkg/tls/tls.go#L103) which returns a `func(*tls.Config)` that sets `MinVersion` and `CipherSuites` based on the profile — exactly the type that `metricsserver.Options.TLSOpts` expects.

The fix would be to call `NewTLSConfigFromProfile` with the fetched profile and append the result to `TLSOpts`:

```go
tlsConfigFromProfile, _ := openshifttls.NewTLSConfigFromProfile(initialTLSProfileSpec)

metricsOpts.TLSOpts = []func(*tls.Config){
    func(c *tls.Config) {
        c.GetConfigForClient = servingCertController.GetConfigForClient
    },
    tlsConfigFromProfile,
}
```

### Impact

Without this fix, the operator restarts on TLS profile changes but continues to serve metrics with Go's default TLS settings (all cipher suites, TLS 1.0+), regardless of the cluster's configured security profile.

ozzywalsh:
@IshwarKanse 

This PR does not apply the TLS profile anywhere in the operator.

Per the linked comment; the plan would be to to scope this PR to only fetching the TLS Profile & restarting the operator.

Applying the TLS Profile to metrics server & console plugins would be done in individual follow up PRs.

https://github.com/rhobs/observability-operator/pull/1041#pullrequestreview-4032543739

IshwarKanse:
@ozzywalsh Ok, the comment only mentioned UIPlugins. Anyways, I'll proceed with the testing and get back shortly. Thanks.

IshwarKanse:
@ozzywalsh I finished testing and we can go ahead and merge this PR. All existing e2e tests passed. I also added additional tests for this feature, will create a PR for it.

IshwarKanse:
@ozzywalsh @simonpasquier We need to add this comment /override ci/prow/observability-operator-e2e to override the e2e test job as there are ongoing cloud account issues preventing the test cluster in CI being provisioned.

simonpasquier:
> @ozzywalsh @simonpasquier We need to add this comment /override ci/prow/observability-operator-e2e to override the e2e test job as there are ongoing cloud account issues preventing the test cluster in CI being provisioned.

the cluster provisioning issue has been fixed. I wouldn't override any CI job unless there's an urgent reason for.

simonpasquier (commented):
/lgtm
/approve

PR #1042, opened 2026-03-26
chore: add coderabbit configuration

simonpasquier:


.coderabbit.yaml
```diff
@@ -0,0 +1,26 @@
+# Initial configuration taken from the internal "Using CodeRabbit" document.
```
danielmellado, line 1: Maybe omit this line as Observability Operator is an upstream project and won't have access to it?

.coderabbit.yaml
```diff
@@ -0,0 +1,26 @@
...
+  changed_files_summary: false
+  poem: false
+  path_instructions:
+    - path: '**'
+      instructions: >-
+        -Focus on major issues impacting performance, readability, maintainability and security.
+        Avoid nitpicks and avoid verbosity.
+  auto_review:
+    ignore_title_keywords:
+      - WIP
```
danielmellado, line 19: Are there any additional title keywords we may want to add, such as CI/DNM/Whatever? :D

simonpasquier: `WIP` seems pretty standard. We can adjust later if needed, WDYT?

.coderabbit.yaml
simonpasquier: I wanted to record from where the configuration originally came from but no strong feeling. Also the org is now under the Red Hat enterprise account.

.coderabbit.yaml
danielmellado: works for me, let's adjust later if needed.

.coderabbit.yaml
danielmellado: True! didn't really recall that

PR #1045, opened 2026-03-30
COO-1749: ObservabilityInstaller: do not require TLS spec for https

pavolloffay:
This PRs enables HTTPS/TLS in TempoStack when `https://` is specified in the object storage endpoint.

Before this PR the HTTPS could be be enabled with the following CR:

```
kind: ObservabilityInstaller
metadata:
  name: tracing
  namespace: observability
spec:
  capabilities:
    tracing:
      enabled: true
      operators:
        install: true
      storage:
        objectStorage:
          tls: {}    # <-- add this, empty struct will instruct to use HTTPS in TempoStack                                                                                                        
          s3:                                                                                                              
            bucket: hlab                                                                                                           
            endpoint: https://hel1.your-objectstorage.com/                                                                          
            accessKeyID: <redacted>                                                                                                
            accessKeySecret:                                                                                                       
              name: s3-access-secret                                                                                               
              key: access_key_secret                                                                                               
            region: eu-central                                                                                                     
```

### DIFF
--- a/pkg/controllers/observability/tempo_components.go
+++ b/pkg/controllers/observability/tempo_components.go
@@ -3,6 +3,7 @@ package observability
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"strings"
 
 	tempov1alpha1 "github.com/grafana/tempo-operator/api/tempo/v1alpha1"
@@ -96,19 +97,26 @@ func tempoStorageCAConfigMapName(name string) string {
 	return fmt.Sprintf("coo-%s-tempo-storage-ca", name)
 }
 
+// tempoStorageSecretName returns the name of the secret that contains the TLS cert and key for the object storage.
 func tempoStorageSecretName(name string) string {
 	return fmt.Sprintf("coo-%s-tempo-storage-cert", name)
 }
 
+// tempoSecretName returns the name of the secret that contains the credentials for the object storage.
 func tempoSecretName(name string) string {
 	return fmt.Sprintf("coo-%s-tempo", name)
 }
 
 func s3hasHTTPSEndpoint(storageSpec obsv1alpha1.TracingObjectStorageSpec) bool {
-	if storageSpec.S3 != nil && strings.HasPrefix(storageSpec.S3.Endpoint, "https://") {
-		return true
+	if storageSpec.S3 == nil {
+		return false
 	}
-	return false
+	endpoint := strings.TrimSpace(storageSpec.S3.Endpoint)
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(parsed.Scheme, "https")
 }
 
 type tempoSecrets struct {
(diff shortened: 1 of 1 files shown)


IshwarKanse:
## Test Results: ObservabilityInstaller S3 HTTPS endpoint fix

**Commit:** `f526580` (`fix: ObservabilityInstaller: do not require TLS spec for https`)
**Cluster:** OpenShift 4.x with COO built from the fix commit, tempo-operator v0.20.4

### Setup
- MinIO deployed as S3-compatible storage in `coo-test` namespace
- ObservabilityInstaller CR created for each scenario, verified the generated TempoStack `spec.storage.tls` and the Tempo ConfigMap `storage.trace.s3.insecure` field

### Results

| # | Scenario | `TLS.Enabled` | `insecure` | Result |
|---|----------|:---:|:---:|:---:|
| 1 | `https://` endpoint, no `tls` block | `true` | `false` | :white_check_mark: **PASS** |
| 2 | `http://` endpoint, no `tls` block | `false` | `true` | :white_check_mark: **PASS** |
| 3 | `https://` endpoint, `tls: {minVersion: VersionTLS13}` | `true` | `false` | :white_check_mark: **PASS** |
| 4 | `http://` endpoint, `tls: {minVersion: VersionTLS12}` | `true` | `false` | :white_check_mark: **PASS** |
| 5 | `https://` endpoint, empty `tls: {}` | `true` | `false` | :white_check_mark: **PASS** |
| 6 | No scheme in endpoint, no `tls` block | `false` | `true` | :white_check_mark: **PASS** |

### Details

- **Test 1** (bug fix): Previously failed because `TLS.Enabled` was `false` when no `tls` block was provided, causing Tempo to connect over HTTP. Now the COO detects `https://` in the S3 endpoint and sets `TLS.Enabled = true` automatically.
- **Test 2** (no regression): `http://` endpoints without TLS config remain insecure as expected.
- **Test 3** (no regression): Explicit `tls` block with `minVersion` works as before, with `https://` also triggering TLS enablement.
- **Test 4** (no regression): Explicit `tls` block overrides the `http://` scheme, enabling TLS. This is existing behavior preserved.
- **Test 5** (no regression): Empty `tls: {}` workaround still works. Both `tls != nil` and `s3hasHTTPSEndpoint()` trigger TLS.
- **Test 6** (edge case): Endpoint without a scheme is treated as insecure (HTTP), consistent with the tempo-operator's `getS3Params()` behavior.

### Note on Azure / GCS
The fix only checks S3 endpoints (`s3hasHTTPSEndpoint`). This is correct because:
- Azure and GCS always connect over HTTPS using the system CA trust store
- The tempo-operator **rejects** `TLS.Enabled = true` for Azure and GCS with: _"custom TLS settings are not supported for Azure/Google Cloud Storage"_
- No changes needed for non-S3 storage backends

pkg/controllers/observability/tempo_components.go
```diff
@@ -96,6 +104,13 @@ func tempoSecretName(name string) string {
 	return fmt.Sprintf("coo-%s-tempo", name)
 }
 
+func s3hasHTTPSEndpoint(storageSpec obsv1alpha1.TracingObjectStorageSpec) bool {
+	if storageSpec.S3 != nil && strings.HasPrefix(storageSpec.S3.Endpoint, "https://") {
```
jan--f: nit: Should this perhaps use the url package?

jan--f (commented):
Just one small question, otherwise this looks good.

pkg/controllers/observability/tempo_components.go
pavolloffay: Fixed

PR #1055, opened 2026-04-14
test: add e2e tests for TLS security profile watcher

IshwarKanse:
## Summary
- Adds e2e tests for the TLS security profile watcher feature introduced in #1041
- Tests verify the operator correctly restarts when the `APIServer` CR's `tlsSecurityProfile` changes and remains stable when non-TLS fields change
- Test file is prefixed with `zz_` to run last in the suite, since modifying the APIServer TLS profile triggers a MachineConfigPool rollout

## Test scenarios

| Test | What it verifies |
|------|-----------------|
| **Default TLS profile** | APIServer CR is readable; operator is running with default (nil/Intermediate) profile |
| **Restart on Old profile** | Changing TLS profile to `Old` triggers operator container restart; operator recovers to ready |
| **Restart on Custom profile** | Setting a `Custom` profile with specific ciphers triggers restart; operator recovers |
| **No restart on non-TLS change** | Adding an annotation to the APIServer CR does NOT cause the operator to restart |

## Design decisions

- **Tracks container restart count AND pod UID**: the TLS watcher cancels the operator's context causing the process to exit; Kubernetes either restarts the container (restart count increases) or replaces the pod entirely (new UID) depending on backoff state
- **Uses `waitForStableRestartCount`**: waits for restart count to remain unchanged for 30s AND the container to have been running for 15s+ before considering the operator stable — prevents race conditions between test cleanup restarts and subsequent tests
- **Uses annotation changes instead of audit profile changes**: modifying `spec.audit.profile` triggers disruptive MachineConfigPool rollouts; annotations trigger the watcher reconcile without cluster impact
- **Resets TLS profile at suite start**: ensures a clean nil/Intermediate baseline regardless of what previous test runs may have left behind

## Test plan
- [x] All 4 TLS profile tests pass in isolation
- [x] All 4 TLS profile tests pass as part of the full e2e suite
- [x] No regressions in existing e2e tests

### DIFF
--- a/test/e2e/zz_tls_profile_test.go
+++ b/test/e2e/zz_tls_profile_test.go
@@ -1,0 +1,433 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"gotest.tools/v3/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/rhobs/observability-operator/test/e2e/framework"
+)
+
+const (
+	operatorDeploymentName = "observability-operator"
+	operatorContainerName  = "operator"
+)
+
+func TestTLSProfileWatcher(t *testing.T) {
+	if !f.IsOpenshiftCluster {
+		t.Skip("TLS profile watcher requires an OpenShift cluster")
+	}
+
+	// Ensure a clean baseline by resetting the TLS profile to nil (Intermediate).
+	// Previous test runs may have left a non-default profile on the APIServer.
+	setTLSProfile(t, nil)
+	f.AssertDeploymentReady(operatorDeploymentName, f.OperatorNamespace,
+		framework.WithTimeout(5*time.Minute))(t)
+
+	ts := []testCase{
+		{
+			name:     "operator reads the default TLS profile and is running",
+			scenario: assertOperatorRunningWithDefaultTLSProfile,
+		},
+		{
+			name:     "operator restarts when TLS profile changes from default to Old",
+			scenario: assertOperatorRestartsOnTLSProfileChange,
+		},
+		{
+			name:     "operator restarts when TLS profile changes to Custom",
+			scenario: assertOperatorRestartsOnCustomTLSProfile,
+		},
+		{
+			name:     "operator does not restart when APIServer non-TLS field changes",
+			scenario: assertOperatorStableOnNonTLSChange,
+		},
+	}
+	for _, tc := range ts {
+		t.Run(tc.name, tc.scenario)
+	}
+}
+
+// assertOperatorRunningWithDefaultTLSProfile verifies that the operator is
+// running and healthy when the cluster uses the default TLS profile (no
+// tlsSecurityProfile set on the APIServer CR, which defaults to Intermediate).
+func assertOperatorRunningWithDefaultTLSProfile(t *testing.T) {
+	apiServer := &configv1.APIServer{}
+	err := f.K8sClient.Get(context.Background(), types.NamespacedName{Name: "cluster"}, apiServer)
+	assert.NilError(t, err, "failed to get APIServer CR")
+
+	f.AssertDeploymentReady(operatorDeploymentName, f.OperatorNamespace,
+		framework.WithTimeout(2*time.Minute))(t)
+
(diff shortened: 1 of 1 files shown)


jan--f:
@coderabbitai review

test/e2e/zz_tls_profile_test.go
```diff
@@ -0,0 +1,412 @@
...
+// least 15 seconds. This ensures any pending restarts from previous test
+// cleanups have completed and the watcher has had time to initialize.
+func waitForStableRestartCount(t *testing.T) int32 {
+	t.Helper()
+
+	var lastCount int32 = -1
+	var stableSince time.Time
+
+	err := wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
+		current := getOperatorContainerRestartCount(t)
```
jan--f: `getOperatorContainerRestartCount()` might call `t.Fatal`and in turn cancel out the poll we're in unless I'm mistaken. Perhaps `getOperatorContainerRestartCount()` rahter should return an error instead of calling `t.Fatal` itself?

test/e2e/zz_tls_profile_test.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+		if time.Since(stableSince) < 30*time.Second {
+			return false, nil
+		}
+
+		// Also verify the container has been running long enough for
+		// the watcher to process its initial reconcile event.
+		pod := getRunningOperatorPod(t)
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.State.Running != nil {
```
jan--f: Should this also check for the operator container name?

test/e2e/zz_tls_profile_test.go
```diff
@@ -0,0 +1,444 @@
...
+	originalTLSProfile := apiServer.Spec.TLSSecurityProfile
+
+	f.CleanUp(t, func() {
+		restoreTLSProfile(t, originalTLSProfile)
+	})
+
+	f.AssertDeploymentReady(operatorDeploymentName, f.OperatorNamespace,
+		framework.WithTimeout(2*time.Minute))(t)
+
+	initialRestarts := getOperatorContainerRestartCount(t)
```
jan--f: Any reason this line doesn't use `waitForStableRestartCount(t)`
like in https://github.com/rhobs/observability-operator/pull/1055/changes#diff-0129072e2ffdb4c3a921b921c35b05f51dc21d7b08badaf67f1c4d589dfb17b8R125?

IshwarKanse: Good catch — no reason, it should use `waitForStableRestartCount(t)` for consistency and to guard against delayed restarts from the parent function's `setTLSProfile(t, nil)` reset. Fixed in b5572fd.

test/e2e/zz_tls_profile_test.go
```diff
@@ -0,0 +1,433 @@
...
+	err := f.K8sClient.Get(context.Background(), types.NamespacedName{Name: "cluster"}, apiServer)
+	assert.NilError(t, err, "failed to get APIServer CR")
+
+	f.AssertDeploymentReady(operatorDeploymentName, f.OperatorNamespace,
+		framework.WithTimeout(2*time.Minute))(t)
+
+	t.Logf("APIServer TLS profile type: %v", apiServer.Spec.TLSSecurityProfile)
+}
+
+// assertOperatorRestartsOnTLSProfileChange sets the TLS profile to Old,
```
simonpasquier: why not use the Modern profile? I'd rather test with a recommended setup than something like Old.

simonpasquier (commented):
> Test file is prefixed with zz_ to run last in the suite, since modifying the APIServer TLS profile triggers a MachineConfigPool rollout

I'm not sure about this: the runner executes tests using the alphabetical order of the test names, not filenames.

test/e2e/zz_tls_profile_test.go
IshwarKanse: Good suggestion — switched to the `Modern` profile in the latest commit. Verified on a live cluster that it is accepted by the APIServer and correctly triggers an operator restart. The test now covers a recommended security setup rather than the legacy `Old` profile.

IshwarKanse:
> Test file is prefixed with zz_ to run last in the suite, since modifying the APIServer TLS profile triggers a MachineConfigPool rollout
>
> I'm not sure about this: the runner executes tests using the alphabetical order of the test names, not filenames.

The `zz_` prefix works at the **filename** level, not the test name level. Go's `go test` tool compiles and processes source files in lexicographic order, so `Test*` functions defined in `zz_tls_profile_test.go` are placed after all functions in files that sort earlier (e.g. `uiplugin_test.go`). This ensures `TestTLSProfileWatcher` runs last regardless of its name. The PR description has been updated to clarify this.

Also addressed the `Modern` profile suggestion — switched from `Old` to `Modern` in the latest commit (verified on a live cluster that it is accepted and triggers an operator restart correctly).

PR #1056, opened 2026-04-16
COO-1553: feat: bind the tls profile with the ui plugin deployments

jgbernalp:
Fixes: [COO-1553](https://redhat.atlassian.net/browse/COO-1553)

This PR binds the cluster TLS profile to the plugins, we added a flag in the compatibility matrix as not all the plugins support the command line args yet. We will be setting the value when the upstream plugin images are ready. Then we can completely remove the check when all the plugins support the TLS configuration.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -23,9 +23,9 @@ import (
 	"os"
 	"slices"
 
-	obopo "github.com/rhobs/obo-prometheus-operator/pkg/operator"
 	configv1 "github.com/openshift/api/config/v1"
 	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
+	obopo "github.com/rhobs/obo-prometheus-operator/pkg/operator"
 	"go.uber.org/zap/zapcore"
 	k8sflag "k8s.io/component-base/cli/flag"
 	"k8s.io/utils/ptr"
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -24,10 +24,13 @@ import (
 	"slices"
 
 	obopo "github.com/rhobs/obo-prometheus-operator/pkg/operator"
+	configv1 "github.com/openshift/api/config/v1"
+	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
 	"go.uber.org/zap/zapcore"
 	k8sflag "k8s.io/component-base/cli/flag"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/rhobs/observability-operator/pkg/operator"
@@ -134,6 +137,23 @@ func main() {
 	ctx, cancel := context.WithCancel(signalCtx)
 	defer cancel()
 
+	var initialTLSProfileSpec configv1.TLSProfileSpec
+	if openShiftEnabled {
+		scheme := operator.NewOpenShiftScheme()
+		directClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{Scheme: scheme})
+		if err != nil {
+			setupLog.Error(err, "failed to create client for TLS profile fetch")
+			os.Exit(1)
+		}
+
+		initialTLSProfileSpec, err = openshifttls.FetchAPIServerTLSProfile(ctx, directClient)
+		if err != nil {
+			setupLog.Error(err, "failed to fetch TLS profile from cluster")
+			os.Exit(1)
+		}
+		setupLog.Info("fetched initial TLS profile", "minVersion", initialTLSProfileSpec.MinTLSVersion, "ciphers_len", len(initialTLSProfileSpec.Ciphers), "ciphers", initialTLSProfileSpec.Ciphers)
+	}
+
 	op, err := operator.New(
 		ctx,
 		operator.NewOperatorConfiguration(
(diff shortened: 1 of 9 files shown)


pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -52,12 +54,16 @@ func PluginInfoBuilder(ctx context.Context, k client.Client, dk dynamic.Interfac
...
 
 	namespace := pluginConf.ResourcesNamespace
+
+	var pluginInfo *UIPluginInfo
+	var err error
+
 	switch plugin.Spec.Type {
 	case uiv1alpha1.TypeDashboards:
-		return createDashboardsPluginInfo(plugin, namespace, plugin.Name, image)
+		pluginInfo, err = createDashboardsPluginInfo(plugin, namespace, plugin.Name, image)
```
simonpasquier, line 64: (nit) I'd rather see an early exit in case of error.
```suggestion
		pluginInfo, err = createDashboardsPluginInfo(plugin, namespace, plugin.Name, image)
		if err != nil {
			return nil, err
		}
```

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -78,17 +84,30 @@ func PluginInfoBuilder(ctx context.Context, k client.Client, dk dynamic.Interfac
...
 
 	case uiv1alpha1.TypeMonitoring:
-		return createMonitoringPluginInfo(plugin, namespace, plugin.Name, image, compatibilityInfo.Features, clusterVersion, pluginConf.Images["health-analyzer"], pluginConf.Images["perses"])
+		pluginInfo, err = createMonitoringPluginInfo(plugin, namespace, plugin.Name, image, compatibilityInfo.Features, clusterVersion, pluginConf.Images["health-analyzer"], pluginConf.Images["perses"])
+
+	default:
+		return nil, fmt.Errorf("plugin type not supported: %s", plugin.Spec.Type)
+	}
+
+	if pluginInfo != nil {
```
simonpasquier: related to my other comment above: if we return early on error, we don't have to check for nil pluginInfo.

pkg/operator/operator.go
```diff
@@ -329,6 +329,8 @@ func New(ctx context.Context, cfg *OperatorConfiguration) (*Operator, error) {
 			return nil, fmt.Errorf("failed to fetch TLS profile from cluster: %w", err)
 		}
 
+		cfg.UIPlugins.TLSProfile = initialTLSProfileSpec
```
simonpasquier: it can be fixed later but it feels like the TLS Profile configuration should be injected by main() just like the other options.

pkg/controllers/uiplugin/plugin_info_builder.go
zhuje: Updated in the latest push

pkg/controllers/uiplugin/plugin_info_builder.go
zhuje: Updated in the latest push

pkg/operator/operator.go
zhuje: Updated in the latest push

pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -45,19 +48,66 @@ var pluginTypeToConsoleName = map[uiv1alpha1.UIPluginType]string{
...
+		"ECDHE-ECDSA-AES128-GCM-SHA256":   "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",   
+		"ECDHE-RSA-AES128-GCM-SHA256":     "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",     
+		"ECDHE-ECDSA-AES256-GCM-SHA384":   "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",  
+		"ECDHE-RSA-AES256-GCM-SHA384":     "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",     
+		"ECDHE-ECDSA-CHACHA20-POLY1305":   "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256", 
+		"ECDHE-RSA-CHACHA20-POLY1305":     "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",   
+	}
+}
+
+func translateCiphers(opensslCiphers []string, logger logr.Logger) []string {
```
jgbernalp: we could use `libgocrypto.OpenSSLToIANACipherSuites(opensslCiphers)` from `libgocrypto "github.com/openshift/library-go/pkg/crypto"`

zhuje: Nice! I removed the custom translation function and used the library `libgocrypto`.

cmd/operator/main.go
```diff
@@ -134,6 +137,23 @@ func main() {
...
+			setupLog.Error(err, "failed to create client for TLS profile fetch")
+			os.Exit(1)
+		}
+
+		initialTLSProfileSpec, err = openshifttls.FetchAPIServerTLSProfile(ctx, directClient)
+		if err != nil {
+			setupLog.Error(err, "failed to fetch TLS profile from cluster")
+			os.Exit(1)
+		}
+		setupLog.Info("fetched initial TLS profile", "minVersion", initialTLSProfileSpec.MinTLSVersion, "ciphers_len", len(initialTLSProfileSpec.Ciphers), "ciphers", initialTLSProfileSpec.Ciphers)
```
simonpasquier, line 154: (nit) we don't really the length of the cipher list?

simonpasquier (commented):
/lgtm

cmd/operator/main.go
zhuje: No we don't need the length. I'll remove it in a follow-up PR.

zhuje: The follow up PR https://github.com/rhobs/observability-operator/pull/1074

PR #1063, opened 2026-04-20
fix: Use pointer-to-struct for fields with json omitempty

alanconway:
Use pointers for API struct fields marked "omitempty".

"omitempty" does not work on non-pointer struct fields.
A zero valued struct will serialize at least as "{}" and may contain zero valued fields
if any fields are not "omitempty".
This can cause problems with default values and round-trip (de)serialization
and break even if the kubebuilder "+optional" comment is applied.

Additional fixes:
- Rename TracingObjectStorageSpec.GCSSTSSpec to GCSWIF to match its type and json tag
- Update CEL validation rule to handle nil storage/objectStorage with has() guards
- Add nil guards for Tracing, Storage, ObjectStorageSpec, and Operators in controller code
- Remove spurious omitempty from required S3Spec.AccessKeySecret field
- Fix ConfigMapKeySelector doc comments that incorrectly referenced "Secret"


pavolloffay (approved):

alanconway:
@pavolloffay need a new LGTM, rebased and added nil-safe GetFoo() accessors to get rid of long, error-prone != nil tests.

pavolloffay (approved):

PR #1070, opened 2026-04-24
COO-1550: apply cluster TLS profile to health-analyzer

tremes:
[COO-1550](https://redhat.atlassian.net/browse/COO-1550)


tremes:
Thanks for approval @jgbernalp.

I think this also needs an update of health-analyzer image reference here (to pass the new added E2E test). I will build and tag some health-analyzer rc0 image.

jgbernalp (approved):

PR #1071, opened 2026-04-24
OBSINTA-1219: cache ClusterVersion in Framework to avoid redundant API calls (Tests)

DavidRajnoha:
Cache the full `ClusterVersion` object on `Framework` during `Setup()` so that `SkipIfClusterVersionBelow` reads from the cached value instead of making a redundant API call on every invocation.

Also improves the nil-check error message: a missing `ClusterVersion` (i.e. non-OpenShift cluster) now fails with a clearer `"cluster version not available (non-OpenShift cluster?)"` rather than a generic API error.

Includes a comment in the health analyzer test explaining the unique name suffix used to avoid conflicts on re-runs.


simonpasquier (commented):
/lgtm

PR #1076, opened 2026-04-29
COO-1493: fix: create RoleBinding when NoClusterRoleBindings policy is set (COO…

jan--f:
…-1493)

When a MonitoringStack has createClusterRoleBindings: NoClusterRoleBindings with a namespaceSelector, neither a ClusterRoleBinding nor a RoleBinding was created, leaving ServiceAccounts with no permissions and preventing StatefulSet creation.


simonpasquier (commented):
/lgtm

PR #1083, opened 2026-05-06
[POC] test: e2e test for UIPlugin cleanup on operator uninstall (COO-1404)

DavidRajnoha:
## Summary

- Adds `TestUIPluginUninstallCleanup` e2e test that reproduces the bug where UIPlugin operands are orphaned when COO is uninstalled via OLM without manually deleting UIPlugin CRs first
- Creates a monitoring UIPlugin with health-analyzer, simulates OLM uninstall (CSV + Subscription deletion), then asserts cleanup of UIPlugin CR, Deployments, Services, ServiceAccount, ClusterRole, ClusterRoleBinding, and pods
- All 9 assertions currently **fail**, confirming the bug
- Uses a custom `waitForResourceAbsent` instead of `Framework.AssertResourceAbsent` due to a silent-pass bug in the framework (#1082)
- Includes self-healing `t.Cleanup` that reinstalls the Subscription after the test

### Run modes

```bash
# Normal: auto-restores operator afterward
go test -v -timeout 20m ./test/e2e/... \
  -run TestUIPluginUninstallCleanup -count 1 \
  -args -operatorInstallNS="openshift-cluster-observability-operator"

# Inspect: pauses before restoring (bump -timeout accordingly)
go test -v -timeout 30m ./test/e2e/... \
  -run TestUIPluginUninstallCleanup -count 1 \
  -args -operatorInstallNS="openshift-cluster-observability-operator" \
  -postpone-restoration=10m

# Retain: never restores, full manual control
go test -v -timeout 20m ./test/e2e/... \
  -run TestUIPluginUninstallCleanup -count 1 \
  -args -operatorInstallNS="openshift-cluster-observability-operator" -retain
```

### Example output (all subtests fail as expected)

```
--- FAIL: TestUIPluginUninstallCleanup/UIPlugin_CR_is_deleted (180.00s)
--- FAIL: TestUIPluginUninstallCleanup/monitoring_plugin_deployment_is_deleted (180.00s)
--- FAIL: TestUIPluginUninstallCleanup/health-analyzer_deployment_is_deleted (180.00s)
--- FAIL: TestUIPluginUninstallCleanup/health-analyzer_service_is_deleted (180.00s)
--- FAIL: TestUIPluginUninstallCleanup/monitoring_plugin_service_is_deleted (180.00s)
--- FAIL: TestUIPluginUninstallCleanup/monitoring_plugin_service_account_is_deleted (180.00s)
[...]

PR #1084, opened 2026-05-06
COO-1404: fix: remove ui plugin finalizers in favor of k8s garbage collection

jgbernalp:
This PR:

- Removes the finalizers added to ui plugins to use regular k8s garbage collection as the ConsolePlugin resources have owner references to the UIPlugins. This avoids dangling resources when the operator is uninstalled and does not have time to cleanup the finalizers.
- Removes finalizers form existing resources, for example during upgrades
- Executes a best effort console plugins removal when the manager is being terminated

### DIFF
--- a/hack/dev-deploy.sh
+++ b/hack/dev-deploy.sh
@@ -46,10 +46,7 @@ CAT_NAME=$(oc get catalogsource | grep 'observability-operator' | awk '{print $1
 SUB_NAME=$(oc get subscriptions | grep 'observability-operator' | awk '{print $1}') && oc delete subscriptions "${SUB_NAME}"
 CSV_NAME=$(oc get clusterserviceversion | grep 'observability-operator' | awk '{print $1}') && oc delete clusterserviceversion "${CSV_NAME}"
 
-# delete uiplugin if hanging by unblock finalizer
-kubectl patch uiplugin monitoring --type='merge' -p='{"metadata":{"finalizers":null}}'
-
-# OR Delete the whole operator 
+# OR Delete the whole operator
 operator-sdk cleanup observability-operator -n openshift-operators
 
 # Run the bundle using the fully qualified image tag.
--- a/pkg/controllers/uiplugin/controller.go
+++ b/pkg/controllers/uiplugin/controller.go
@@ -224,35 +224,19 @@ func (rm resourceManager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	if !plugin.ObjectMeta.DeletionTimestamp.IsZero() {
 		logger.V(6).Info("deregistering plugin from the console")
 		if err := rm.deregisterPluginFromConsole(ctx, pluginTypeToConsoleName[plugin.Spec.Type]); err != nil {
-			return ctrl.Result{}, err
+			logger.V(3).Info("best-effort console deregistration failed during deletion", "error", err)
 		}
 
-		// Remove finalizer if present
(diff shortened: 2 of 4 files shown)


DavidRajnoha:
Hi @jgbernal, I tested this PR against OCP 4.21 using a dedicated uninstall e2e test (PR #1083). The image was built from your commit `a55377c` and confirmed by matching the `shutdown-cleanup` logger in cluster event logs.

**This PR fixes the UI uninstall.** The finalizer removal + OwnerReference approach is correct and works well for the primary uninstall path (OpenShift console UI with "delete operands" selected). There are some remaining gaps for secondary uninstall paths that could be addressed as follow-up work.

### What works

- Removing the finalizer eliminates stuck-terminating UIPlugins during normal deletion.
- The OwnerReference chain is set up correctly — deleting a UIPlugin CR cascade-deletes all children (Deployments, Services, RBAC).
- **Uninstall via OpenShift Console UI** (with "delete all operands" checked) works fully: the UI issues DELETE on the UIPlugin CRs, which triggers the cascade, and the operator handles Console deregistration while still alive.

### Remaining gap: CLI / automated uninstall

When uninstalling via CLI (`oc delete csv`, `oc delete sub`) or any path that doesn't explicitly delete UIPlugin CRs first, the cleanup doesn't happen because nobody issues the initial DELETE on the UIPlugin CR — the OwnerReference cascade never starts.

**The shutdown handler also can't compensate.** OLM deletes the operator's RBAC before the SIGTERM handler runs:

```
2026-05-07T13:59:22Z  ERROR  shutdown-cleanup  failed to list UIPlugins during shutdown cleanup
  {"error": "uiplugins.observability.openshift.io is forbidden: User
  \"system:serviceaccount:openshift-cluster-observability-operator:observability-operator-sa\"
  cannot list resource \"uiplugins\" in API group \"observability.openshift.io\" at the cluster scope"}
```

**What's left orphaned depends on whether the namespace is deleted:**

If only CSV+Sub are deleted (namespace preserved):
- All namespaced resources remain: Deployments, Services, ServiceAccounts, running pods
- All cluster-scoped resources remain: UIPlugin CR, ConsolePlugin, ClusterRoles, ClusterRoleBindings
- Console `spec.plugins` retains stale entry — UI tabs still visible

If the namespace is also deleted:
- Namespaced resources are cleaned up (gone with the namespace)
- **Cluster-scoped resources still orphaned**: UIPlugin CR, ConsolePlugin, ClusterRoles, ClusterRoleBindings (6 resources in total for a monitoring+health-analyzer setup)
- Console `spec.plugins` gets cleaned up (observed in testing — likely reconciled by the Console operator)

### Possible follow-up

An OLM pre-delete Job with its own ServiceAccount could cover the CLI path — running before teardown to delete all UIPlugin CRs (triggering the OwnerRef cascade this PR sets up) and deregistering from Console.

The e2e test in PR #1083 reproduces the CLI uninstall scenario. Note: the test currently in #1083 is missing some recent improvements (log capture, `--postpone-restoration` flag support in `run-e2e.sh`) that haven't been pushed yet.

jgbernalp:
@DavidRajnoha I don't think direct deletion of the csv and/or subscription are paths we should cover, this should be delegated to cluster admins. See https://olm.operatorframework.io/docs/tasks/uninstall-operator/

DavidRajnoha:
@jgbernalp thanks, I was not aware of the proper uninstallation path.

PR #1089, opened 2026-05-11
NO-JIRA: remove SupportsTLSProfile

zhuje:
troubleshooting-panel-console-plugin applies TLS configs
main https://github.com/openshift/troubleshooting-panel-console-plugin/pull/232
release-coo-ocp-4.19 https://github.com/openshift/troubleshooting-panel-console-plugin/pull/233
release-coo-ocp-4.22 https://github.com/openshift/troubleshooting-panel-console-plugin/pull/234

Now that all the uiplugins support TLS configuration we can remove the feature flag SupportsTLSProfile

```
operator-sdk run bundle \
  quay.io/jezhu/observability-operator-bundle:1.5.0-tls-may12-0842 \
  --install-mode AllNamespaces \
  --namespace openshift-operators \
  --security-context-config restricted
```


pkg/controllers/uiplugin/compatibility_matrix.go
```diff
@@ -53,7 +53,7 @@ var compatibilityMatrix = []CompatibilityEntry{
 		ImageKey:           "ui-troubleshooting-panel-pf6",
 		SupportLevel:       GeneralAvailability,
 		Features:           []string{},
-		SupportsTLSProfile: false,
+		SupportsTLSProfile: true,
```
PeterYurkovich: With this change all of the plugins support TLS now, any reason to keep it?

jgbernalp: +1. We discussed that it should be removed once all UI plugins support it.

zhuje: Updated in latest push.

PR #1090, opened 2026-05-12
chore: unpin github.com/openshift/api

simonpasquier:
This commit unpins the version of `github.com/openshift/api` to depend on the latest version. To continue supporting older OpenShift versions which require the Console v1alpha1 API (removed since 2024), we fork the `github.com/openshift/api` under `github.com/rhobs/openshift-api` and pinned to the same version that was used before this change.


jan--f:
lgtm but I'll let @jgbernalp or @PeterYurkovich  have final eyes on this.

simonpasquier:
the CI is green now (with the updated ConsolePlugin API, we need to pass an empty `ContentSecurityPolicy` slice otherwise the Kubernetes API rejects the request).

PR #1092, opened 2026-05-19
COO-1531: feat: expose AlertmanagerConfigMatcherStrategy on MonitoringStack

jan--f:
AlertmanagerConfig resources automatically inject a namespace matcher into routes, forcing users to add namespace labels to PrometheusRule alerts. Expose the matcherStrategy field on MonitoringStack.Spec.AlertmanagerConfig with a default of None so routes match all alerts regardless of namespace.


simonpasquier (commented):
The upstream struct has an inner `type` field which offers the possibility to pass additional parameters if required for new strategies:

```
spec:
  alertmanagerConfigMatcherStrategy:
    type: OnFooBar
    fooBar:
      foo: x
      bar: y
```

It might be good to follow a similar pattern for MonitoringStack?

```
spec:
  alertmanager
    configMatcherStrategy:
      type: None
```

jan--f:
> The upstream struct has an inner `type` field which offers the possibility to pass additional parameters if required for new strategies:

Thanks, good point.

simonpasquier (commented):
/lgtm

PR #1094, opened 2026-05-19
COO-1063: fix: set default resource requests and limits on Alertmanager and Thanos sidecar

jan--f:
When a ResourceQuota is set on the namespace, Alertmanager and Prometheus pods fail to start because the Alertmanager container and Thanos sidecar container don't have resource requests/limits. Propagate the MonitoringStack's resource requirements to both components, matching what is already done for the Prometheus container.

### DIFF
--- a/bundle/manifests/monitoring.rhobs_monitoringstacks.yaml
+++ b/bundle/manifests/monitoring.rhobs_monitoringstacks.yaml
@@ -79,6 +79,73 @@ spec:
                     format: int32
                     minimum: 0
                     type: integer
+                  resources:
+                    default:
+                      limits:
+                        cpu: 250m
+                        memory: 256Mi
+                      requests:
+                        cpu: 50m
+                        memory: 128Mi
+                    description: Define resources requests and limits for the Alertmanager
+                      container.
+                    properties:
+                      claims:
+                        description: |-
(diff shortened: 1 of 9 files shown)


simonpasquier:
hmm wouldn't that be too inflexible/restrictive? In general Prometheus requires higher resources than Alertmanager and Thanos sidecar. Also it would double the requests for the Prometheus pods?

jan--f:
> hmm wouldn't that be too inflexible/restrictive? In general Prometheus requires higher resources than Alertmanager and Thanos sidecar. Also it would double the requests for the Prometheus pods?

Yes indeed. On the MonitoringStack API level however we document `Define resources requests and limits for Monitoring Stack Pods.` so on some level one could argue that technically this here is the more correct behavior.

We could alternatively start introducing resource fields for individual components and deprecate the old field.  The more I think about it this seems like the right path, the current field is poorly thought out.

simonpasquier:
> We could alternatively start introducing resource fields for individual components and deprecate the old field. The more I think about it this seems like the right path, the current field is poorly thought out.

It would have my preference too. I'd keep the scope of the original field unchanged and only propagate it to the Prometheus resource.

simonpasquier (commented):
/lgtm

PR #1100, opened 2026-05-27
fix: Use rhobs/openshift-api for console/v1 for 4.17-4.18

PeterYurkovich:
The content security policy field was added to the ConsolePlugin CRD in OCP 4.19, however the field is not supported in OCP 4.17 and 4.18. A recent update to the `openshift/api` console/v1 version makes the field be marshalled by default, leading to a reconciliation error.  

This PR changes OCP 4.17 and 4.18 to use the `rhobs/openshift-api` console/v1 for OCP 4.17 and 4.18. It checks the OCP version on startup when `openshift.enabled` is true, registers the correct scheme (`rhobs/openshift-api/console/v1` or `openshift/api/console/v1`), then creates the appropriate CR version

### DIFF
--- a/DEPENDENCY_CONSTRAINTS.md
+++ b/DEPENDENCY_CONSTRAINTS.md
@@ -30,8 +30,11 @@ Dependabot takes care of dependency updates, the configuration is located at `.g
 
 ## Future Considerations
 
-When OpenShift &lt; 4.18 support is no longer needed, we can:
+When OpenShift &lt; 4.17 support is no longer needed, we can:
 1. Remove `console/v1alpha1` API usage.
-2. Remove dual API support code.
-3. Remove dependency on `github.com/rhobs/openshift-api` 
 4. Update this document
+
+When OpenShift &lt; 4.19 support is no longer needed, we can:
+1. Remove dual API support code.
+2. Remove dependency on `github.com/rhobs/openshift-api` 
+3. Update this document
--- a/DEPENDENCY_CONSTRAINTS.md
+++ b/DEPENDENCY_CONSTRAINTS.md
@@ -9,17 +9,18 @@ This project depends on 2 versions of `github.com/openshift/api`:
 * A [forked version](https://github.com/rhobs/openshift-api).
 
 **Why Forked:** The observability-operator needs to support both OpenShift console API `v1` and `v1alpha1` for backward compatibility:
-- OpenShift >= 4.17 uses `console/v1` API
-- OpenShift < 4.17 uses `console/v1alpha1` API
+- OpenShift >= 4.19 uses `console/v1` API (openshift/api)
+- OpenShift 4.17 - 4.18 uses `console/v1` API (rhobs/openshift-api)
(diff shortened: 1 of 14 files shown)


simonpasquier (commented):
/lgtm

PeterYurkovich:
Holding PR for QE from @etmurasaki

simonpasquier:
/payload-job periodic-ci-rhobs-observability-operator-main-amd64-ocp-4.18-gcp-coo-stage

cmd/operator/main.go
```diff
@@ -155,6 +156,14 @@ func main() {
 			os.Exit(1)
 		}
 		setupLog.Info("fetched initial TLS profile", "minVersion", initialTLSProfileSpec.MinTLSVersion, "ciphers", initialTLSProfileSpec.Ciphers)
+
+		clusterVersion := &configv1.ClusterVersion{}
```
simonpasquier, line 160: Ideally we shouldn't need to read again the cluster version in https://github.com/rhobs/observability-operator/blob/3f2296717156e0c26d71a9d93ee167b99e0cbad4/pkg/controllers/uiplugin/controller.go#L128

pkg/controllers/uiplugin/components.go
```diff
@@ -51,7 +52,7 @@ var (
 	korrel8rConfigYAMLTmplFile embed.FS
 )
 
-func isVersionAheadOrEqual(currentVersion, version string) bool {
+func IsVersionAheadOrEqual(currentVersion, version string) bool {
```
simonpasquier, line 55: eventually this could be moved elsewhere but not needed for now.

DEPENDENCY_CONSTRAINTS.md
```diff
@@ -29,8 +30,8 @@ Dependabot takes care of dependency updates, the configuration is located at `.g
 
 ## Future Considerations
 
-When OpenShift &lt; 4.17 support is no longer needed, we can:
+When OpenShift &lt; 4.18 support is no longer needed, we can:
```
simonpasquier: I guess that we have 2 steps now
1. remove v1alpha1 when we drop OCP < 4.17
2. remove forked v1 when we drop OCP < 4.19

cmd/operator/main.go
PeterYurkovich: Addressed in latest commit, passed the cluster version down through the controller options

DEPENDENCY_CONSTRAINTS.md
PeterYurkovich: Rewrote to show the 2 steps needed. Removing v1alpha1 and then removing `rhobs-openshift/api` v1

PeterYurkovich:
Looks like I don't have permission to remove the hold label

jan--f:
> Looks like I don't have permission to remove the hold label

Hmm that would be surprising. Maybe a github issue?

/unhold

simonpasquier (commented):
/lgtm

PR #1103, opened 2026-05-28
feat: add agent-navigation feature gate to uiplugin.

alanconway:
Add boolean field 'enableAgentNavgation' to uiplugin troubleshootingPanel struct.
If set, the "agent-navigation" feature gate is added to the troubleshooting panel.
See corresponding PR for troubleshooting panel

### DIFF
--- a/pkg/controllers/uiplugin/plugin_info_builder.go
+++ b/pkg/controllers/uiplugin/plugin_info_builder.go
@@ -71,7 +71,11 @@ func PluginInfoBuilder(ctx context.Context, k client.Client, dk dynamic.Interfac
 		}
 
 	case uiv1alpha1.TypeTroubleshootingPanel:
-		pluginInfo, err = createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
+		var tpFeatures []string
+		if plugin.Spec.TroubleshootingPanel != nil && plugin.Spec.TroubleshootingPanel.EnableAgentNavigation {
+			tpFeatures = append(tpFeatures, "agent-navigation")
+		}
+		pluginInfo, err = createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, tpFeatures)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


pkg/controllers/uiplugin/plugin_info_builder.go
```diff
@@ -71,7 +71,11 @@ func PluginInfoBuilder(ctx context.Context, k client.Client, dk dynamic.Interfac
 		}
 
 	case uiv1alpha1.TypeTroubleshootingPanel:
-		pluginInfo, err = createTroubleshootingPanelPluginInfo(plugin, namespace, plugin.Name, image, []string{})
+		var tpFeatures []string
```
PeterYurkovich, line 74: I'd prefer if we did the agent-navigation check in the troubleshooting_panel.go file instead. The rest of the plugins follow the same pattern, with the `features []string` parameter to the `createXPluginInfo` functions behavior more as a `featuresBasedOnOcpVersion` rather than a list directly of the features. We can consider if we want to change that later, but I would prefer to keep everything aligned for now

pkg/controllers/uiplugin/troubleshooting_panel.go
```diff
@@ -128,14 +128,16 @@ func marshalTroubleshootingPanelPluginConfig(cfg *uiv1alpha1.TroubleshootingPane
...
 	}
 
-	if cfg.Timeout == "" {
+	if cfg.Timeout == "" && !cfg.EnableAgentNavigation {
 		return "", nil
 	}
 
 	pluginCfg := struct {
-		Timeout string `yaml:"timeout"`
+		Timeout               string `yaml:"timeout,omitempty"`
```
PeterYurkovich, line 136: I think it would be better to have the enableAgentNavigation be passed as a feature flag rather than as a config variable. Something like:

https://github.com/rhobs/observability-operator/blob/main/pkg/controllers/uiplugin/logging.go#L52

alanconway:
Moved to 
https://github.com/rhobs/observability-operator/pull/1107

PR #1105, opened 2026-05-28
feat: set perses support level to GA, remove hardcoded image

PeterYurkovich:


etmurasaki:
tested and perses-0 had the correct version.

jan--f:
The thanos test seems to take too long, I'll try to reproduce locally.

/retest

jan--f:
I can reproduce this locally. Increasing the timeout of the test suite 15m->20m succeeds. So I'll assume the tests still work but need more time, we can investigate this later. I'll merge past the blocking check for now to unblock the release.

PR #1112, opened 2026-06-02
NO-ISSUE: test: don't crash test/run-e2e.sh when promq isn't installed

simonpasquier:


simonpasquier:
/cc @jan--f 

this should fix the 4.19 periodic job which don't have the promq binary installed (I'm not sure that the test makes much sense).

danielmellado:
> /cc @jan--f
> 
> this should fix the 4.19 periodic job which don't have the promq binary installed (I'm not sure that the test makes much sense).

does it do anything if no promq?

PR #1122, opened 2026-06-05
COO-1849: warn when AI Navigation is used on unsupported OCP

PeterYurkovich:
The AI navigation feature is only enabled as a dev preview feature in OCP 4.22+. Since the feature is dev preview I have added only a warning, not a failure of the reconciliation loop for enabling an unsupported feature


alanconway (approved):
/lgtm

PR #1124, opened 2026-06-08
fix: handle nil Capabilities/Tracing in ObservabilityInstaller controller

IshwarKanse:
## Summary

Resolves bug https://redhat.atlassian.net/browse/COO-1257 

- Fix nil pointer dereference panic in `updateStatus` when `Capabilities` is non-nil but `Tracing` is nil (e.g. `spec: { capabilities: {} }`)
- Add 11 unit tests for `tempoStack()` covering all nil-input combinations (nil Capabilities, nil Tracing, nil Storage, nil ObjectStorageSpec) and all storage backends (S3, S3STS, Azure, GCS) including TLS variants
- Add e2e test `ObservabilityInstallerEmptySpec` that creates an `ObservabilityInstaller` with `spec: {}` on a live cluster, verifies the controller does not crash-loop, no operands are deployed, status fields remain empty, and the resource deletes cleanly

## Root Cause

`observability_controller.go updateStatus` accessed `capabilities.Tracing.Enabled` (a field dereference) without first checking `capabilities.Tracing != nil`. When an `ObservabilityInstaller` is created with `spec: { capabilities: {} }`, `Capabilities` is non-nil but `Tracing` is nil, causing a nil pointer panic. The fix adds the missing nil guard:

```go
// Before
if capabilities.Tracing.Enabled {

// After
if capabilities.Tracing != nil && capabilities.Tracing.Enabled {
```

Note: the related panic in `tempoStack()` (when `spec: {}` leaves `Capabilities` nil) was already fixed via safe getter methods (`GetCapabilities()`, `GetTracing()`, etc.) on the API types.

## Test plan

- [x] `go test ./pkg/controllers/observability/...` — all 17 tests pass including new `TestTempoStack` with nil/empty input cases
- [x] E2e test `TestObservabilityInstallerController/ObservabilityInstallerEmptySpec` passes on a live OCP cluster — controller stays available, no operands deployed, clean deletion.

### DIFF
--- a/pkg/controllers/observability/observability_controller.go
+++ b/pkg/controllers/observability/observability_controller.go
@@ -206,7 +206,7 @@ func (o observabilityInstallerController) getInstance(ctx context.Context, req c
 func (o observabilityInstallerController) updateStatus(ctx context.Context, instance *obsv1alpha1.ObservabilityInstaller, reconcileErr error) reconcile.Result {
 	if instance.Spec.Capabilities != nil {
 		capabilities := instance.Spec.Capabilities
-		if capabilities.Tracing.Enabled {
+		if capabilities.Tracing != nil && capabilities.Tracing.Enabled {
 			otelcol := &otelv1beta1.OpenTelemetryCollector{}
 			err := o.client.Get(ctx, types.NamespacedName{
 				Namespace: instance.Namespace,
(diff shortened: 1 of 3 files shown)


IshwarKanse:
@coderabbitai review

PR #1130, opened 2026-06-15
WIP: feat: install full obs. stack from manifests.

alanconway:
Not finished, this is an early view for comment. See stack/README.md for reasoning.

Started:
1. Separate resource manifests from controllercode as kustomizations.
2. Convert tracing component to use kustomize overlay approach.
3. API for logging component.

Yet to do:
4. Support all obs. components via kustomize resources.

See stack/README.md for reasoning.


alanconway:
Will move to a new PR

PR #1131, opened 2026-06-22
test: use patch instead of apply in run-e2e-ocp.sh

simonpasquier:
A patch command is less likely to fail than apply because it doesn't need to pass a `resourceVersion` value which can be outdated.

### DIFF
--- a/test/run-e2e-ocp.sh
+++ b/test/run-e2e-ocp.sh
@@ -53,29 +53,26 @@ install_obo() {
 }
 
 enable_ocp() {
-  # Get Observability Operator CSV json file
   CSV_NAME=$(oc -n "$OPERATORS_NS" get sub observability-operator -o jsonpath='{.status.installedCSV}')
-  CSV_JSON_FILE=$(mktemp /tmp/"$CSV_NAME"XXXXXX.json)
-  if [ -e "$CSV_JSON_FILE" ]; then
-    rm -f "$CSV_JSON_FILE"
+
+  local deployment_index
+  deployment_index=$(oc -n "$OPERATORS_NS" get csv "${CSV_NAME}" -o json | \
+    jq '[.spec.install.spec.deployments[].name] | index("observability-operator")')
+  if [[ -z "$deployment_index" || "$deployment_index" == "null" ]]; then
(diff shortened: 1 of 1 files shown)


simonpasquier:
@coderabbitai help

simonpasquier:
@coderabbitai review

test/run-e2e-ocp.sh
```diff
@@ -53,29 +53,26 @@ install_obo() {
...
-    oc -n "$OPERATORS_NS" get csv "${CSV_NAME}" -o json > "$CSV_JSON_FILE"
-    # Update the CSV manifest to enable the OpenShift features.
-    ARGS_JSON=$(printf '%s\n' "--openshift.enabled=true" | jq -R . | jq -s .)
-    jq --arg container_name operator --argjson args "$ARGS_JSON" '
-      (.spec.install.spec.deployments[].spec.template.spec.containers[] | select(.name == $container_name) | .args) += $args
-    ' "$CSV_JSON_FILE" > /tmp/tmp.$$.json && mv /tmp/tmp.$$.json "$CSV_JSON_FILE"
-    ok "Added arguments to container operator in '$CSV_JSON_FILE'."
-    if oc -n "$OPERATORS_NS" apply -f "$CSV_JSON_FILE"; then
+    if oc -n "$OPERATORS_NS" patch csv "${CSV_NAME}" --type=json \
+      -p "[{\"op\": \"add\", \"path\": \"/spec/install/spec/deployments/${deployment_index}/spec/template/spec/containers/0/args/-\", \"value\": \"--openshift.enabled=true\"}]"; then
```
danielmellado, line 71: nit here: `containers/0` is hardcoded here. It's totally cosmetic and  correct since the observability-operator deployment only has a single container, but for consistency with the deployment_index lookup above maybe we can resolve the container index by name too.

simonpasquier: I've hard-coded the value deliberately because I don't think that the container index will ever change (which isn't the case for the deployment index).

danielmellado: no worries, I won't block the PR because of that, fair point!

PR #1134, opened 2026-06-23
CNV-90618: feat: add OCP audit log Perses dashboard with Loki datasource

sradco:
## Summary

- Adds an OCP Kubernetes API audit log viewer dashboard powered by Perses + Loki
- Includes a `PersesDatasource` for the OpenShift Logging Loki gateway (audit tenant)
- Uses `LogsTable` panel with detail view for expanding full log entries
- Provides filters: username (regex), exclude system users (multi-select), verb, resource type, namespace, resource name, response code, exclude resources, client, and a general-purpose raw LogQL filter
- Follows the same pattern as the existing accelerators dashboard (Go SDK builder + JSON round-trip for rhobs type compatibility)

## Jira

https://redhat.atlassian.net/browse/CNV-90618

## Context

This dashboard enables cluster admins to investigate who did what on the cluster through the Perses UI plugin.

Standalone dashboard repo with docs, Go SDK code, and deployment manifests:
https://github.com/sradco/ocp-audit-log-perses-dashboard

### Prerequisites for end users
- OpenShift Logging with LokiStack configured to collect audit logs
- Recommended: OTLP data model with LokiStack schema v13 for structured metadata support (enables filtering without JSON parsing, significantly improving query performance)
- Optional: `ClusterLogForwarder` filters to exclude `auditd` logs and unauthenticated requests (reduces noise and storage)

### Related upstream issues for query performance
- Loki Operator — add `k8s.audit.event.verb` as default audit stream label: https://github.com/grafana/loki/issues/22513
- Cluster Logging Operator — add `verb` to default ViaQ audit labelKeys: https://github.com/openshift/cluster-logging-operator/issues/3317

## Test plan

- [ ] Deploy observability-operator with this change on a cluster with OpenShift Logging + LokiStack
- [ ] Verify the `PersesDatasource` and `PersesDashboard` CRs are created in the plugin namespace
- [ ] Confirm audit logs are displayed in the LogsTable panel
- [ ] Test each filter (username, exclude system users, verb, resource, namespace, resource name, response code, exclude resources, client, LogQL filter)
- [ ] Verify clicking a log row expands the full JSON detail view
- [ ] Confirm `go build ./...` passes with no regressions


simonpasquier:
Do we have a Jira ticket for this? Has it been discussed with PM? Do we want a way for users to opt-in or opt-out?

sradco:
@simonpasquier Yes, the Jira ticket is [CNV-90618](https://redhat.atlassian.net/browse/CNV-90618) — I've added it to the PR description.

Regarding opt-in/opt-out: the dashboard and datasource are created as `OptionalUpdater` components gated by `persesEnabled`, so they only deploy when the Perses UI plugin is active. The dashboard itself requires OpenShift Logging with LokiStack to be functional - without it, the datasource simply won't connect and no data will display. 
Is there any other mechanism available? How is it different from other dashboards?
I believe we should move it to a tab under the Logs UI, when possible.

I did discuss the dashboard with PM, Jamie and Eric.

simonpasquier:
>  The dashboard itself requires OpenShift Logging with LokiStack to be functional - without it, the datasource simply won't connect and no data will display.

IMHO it isn't a good user experience and it will create unnecessary noise for the support. I'd prefer a solution where the dashboard is deployed only when the operator knows that it will work properly.

PeterYurkovich:
I believe the original plan was to have COO be responsible for creating the CRD's (the actual definitions), reconciling the created CR's into perses, and then displaying the dashboards in the console. Any other operator (in this case either loki or logging) would be able to create the perses related resources, in this case a `PersesDatasource`/`PersesGlobalDatasource` and a `PersesDashboard` to show whatever default dashboards they would like. This is similar to how the current metrics-only dashboards are handled in the console, where any operator can create a ConfigMap with the correct annotation and it will get picked up and displayed

jgbernalp:
@PeterYurkovich @simonpasquier this dashboard is intended for admins and is created here as it has a similar restriction than the tempo dashboard, which cannot be deployed directly by an upstream operator, in this case Loki. Until we have the capacity to include these dashboards as templates as part of a catalog in which users can decide which dashboards to use, we can use COO to deliver them. @sradco What we can improve is detect the Loki stack configuration to set up the Datasource correctly so the dashboard is functional. WDYT?

sradco:
> @PeterYurkovich @simonpasquier this dashboard is intended for admins and is created here as it has a similar restriction than the tempo dashboard, which cannot be deployed directly by an upstream operator, in this case Loki. Until we have the capacity to include these dashboards as templates as part of a catalog in which users can decide which dashboards to use, we can use COO to deliver them. @sradco What we can improve is detect the Loki stack configuration to set up the Datasource correctly so the dashboard is functional. WDYT?

@simonpasquier , @jgbernalp I updated the code based on your feedback.

PeterYurkovich:
Why is this unable to be deployed by an upstream operator?

sradco:
> Why is this unable to be deployed by an upstream operator?

This is a cluster-level dashboard. 
Audit logs span all workloads and namespaces, so no single upstream operator owns this concern. The observability-operator is the natural integration point since it already manages the Perses instance, LokiStack discovery, and the monitoring console plugin where the dashboard is surfaced.

jgbernalp:
> Why is this unable to be deployed by an upstream operator?

In addition to what @sradco said. This data is related to the Loki Operator (upstream) and specific to OpenShift configuration of the `audit` tenant. If we added to CLO, we would have the same issue we had before with the logging view plugin in which an unrelated operator was deploying the plugin that only used Loki backend. I think in the future there will be more dashboards of this type, for example the ones already present in the community mixings. The long term solution is to add a Perses catalog, so the community mixins or any other OpenShift specific dashboards can live and customers can decide which ones they want to have.

PeterYurkovich:
Thanks for the explanation. I understand better now. I knew we had discussed having dashboards be stored in COO until the marketplace was ready. It makes sense to have it live here.

I think this is fine for now to "just work" if there is exactly 1 lokistack in the openshift-logging and to fail if that isn't true. However I think it would make sense to have a further discussion on how we should automatically create data sources for all our signal types. 

It seems to me like it would make more sense to have a datasource or global datasource created for every lokistack (or signal instance). I'm not exactly sure if it would make sense to put the creation of the datasources in the singal operators, the perses operator or here. The signal operators would be able to do it in their reconcile loop for the signal CR, the perses operator would enable it to be available to users outside of openshift, and COO could make sense as a central location for openshift signals. Maybe adding a "automatically create data sources" feature flag onto the monitoring ui plugin or the perses operator flags which is on by default and can be toggled off. It would need whichever operator to watch even more CR types though, which they are already having memory issues with

simonpasquier:
>  However I think it would make sense to have a further discussion on how we should automatically create data sources for all our signal types.

+100 to this, the argument of "we have no better place right now than COO but we'll find a long-term solution later" hasn't proved to work well (*). I'd rather pause now and discuss our short vs. long term strategies for this type of request. COO office hours and/or Obs Arch call would be the right place but first we need to summarize past discussions and requests into a central document.

(*) one immediate problem is that it's not clear to me who will the owner for the code being added here.

sradco:
We will maintain this dashboard as its high priority for our users, but it is cluster level so it  doesnt make sense to have it only as part of our operator.

I see this as a dashboard that we eventually be integrated also to our resources in the UI and could be added to any resource in OCP as an `Audit Trail` tab.
This dashboard is will probably be for Admins only, where a tabs in the resources would be useful for any user that has access to that tabs.

I will be fine with moving the code where you decide make sense for these type of dashboards.

sradco:
@simonpasquier , @PeterYurkovich , @jgbernalp , Is there anything preventing us from merging this dashboard?

jgbernalp:
@sradco upon discussion we believe is better to add these set of dashboards to the [community-mixins](https://github.com/perses/community-mixins) as the current dashboards that are hosted here will be also moved there. The reason for this is that we don't want to pollute COO with the specific dashboards that users might or might not use. The long term solution is to add the catalog, in which the dashboard templates can be installed, this will allow customers to use only what they need and avoid getting dashboards that they didn't ask for and don't work on their setup. Until we have that solution, the community-mixins is the de facto first source for the catalog and dashboards present there can be installed using the percli tool or applying the pre defined CRs. Would it be fine for you to move this dashboard there?

sradco:
I agree with this approach.
I raised this issue about how operators deploy the dashbords and the issue with them being added by default and reconciled.
So what about dashboards that are deployed by an operator like OCP-V?
How will they integrate with the cataloge?
When is the cataloge planned to be shipped?

sradco:
The issue I see for CNV is that our customers really need the UI and until the cataloge is available, this is a gap.

jgbernalp:
> I agree with this approach. I raised this issue about how operators deploy the dashbords and the issue with them being added by default and reconciled. So what about dashboards that are deployed by an operator like OCP-V? How will they integrate with the cataloge? When is the cataloge planned to be shipped?

If you remember our discussion around dashboard templates https://github.com/perses/perses-operator/discussions/230 An operator can define a template CR, that will be displayed in the catalog and the user can "apply" the template.

The UI customization is our first priority, which is currently in progress, after that we will add the catalog.

jgbernalp:
> The issue I see for CNV is that our customers really need the UI and until the cataloge is available, this is a gap.

As an interim solution the customer can apply the dashboard from the community dashboard repo

sradco:
> > I agree with this approach. I raised this issue about how operators deploy the dashbords and the issue with them being added by default and reconciled. So what about dashboards that are deployed by an operator like OCP-V? How will they integrate with the cataloge? When is the cataloge planned to be shipped?
> 
> If you remember our discussion around dashboard templates [perses/perses-operator#230](https://github.com/perses/perses-operator/discussions/230) An operator can define a template CR, that will be displayed in the catalog and the user can "apply" the template.
> 
> The UI customization is our first priority, which is currently in progress, after that we will add the catalog.

Yes, That was what I was thinking would make sense :)
Where should the operators add  the templates to?

sradco:
As discussed, I've moved the audit log viewer dashboard to the community-mixins repo:
https://github.com/perses/community-mixins/pull/267

This follows the agreed approach of having community-mixins be the home for these dashboards until the Perses catalog is available.

sradco:
Closing in favor of https://github.com/perses/community-mixins/pull/267

PR #1115, opened 2026-06-03
test: add DumpOnFailure debug infrastructure for all e2e tests

DavidRajnoha:
## Summary

- Introduces a composable `DebugFunc` type and `DumpOnFailure` helper in the e2e framework — tests register diagnostic functions that only fire when the test fails
- Adds `DebugNamespace(ns...)` which dumps deployments (with conditions), pods (with container statuses), and events for the given namespaces
- Refactors `dumpClusterHealthAnalyzerDebug` into a reusable `dumpUIPluginDebug` returning a `DebugFunc`
- Wires `DumpOnFailure` into all e2e test suites for consistent post-failure diagnostics

## How it works

```go
// Register namespace-level debug (runs only on failure, LIFO before CleanUp)
f.DumpOnFailure(t, f.DebugNamespace("my-namespace"))

// Register CR-specific debug
f.DumpOnFailure(t, dumpUIPluginDebug(plugin.Name))
```

Cleanups run in LIFO order, so `DumpOnFailure` should be called **before** `CleanUp` to ensure debug info is captured while resources still exist.

<details>
<summary>Example output (truncated) — namespace with 6 deployments, 7 pods, 137 events</summary>

```
=== RUN   TestObservabilityInstallerController/ObservabilityInstallerTracing
    framework.go: --- Dumping debug info for namespace openshift-cluster-observability-operator ---
    framework.go: === BEGIN DEBUG DUMP ===
    framework.go: Deployments in namespace openshift-cluster-observability-operator: 6
    framework.go:   Deployment: name=health-analyzer replicas=1 readyReplicas=1 availableReplicas=1
    framework.go:     condition: type=Available status=True reason=MinimumReplicasAvailable ...
    framework.go:   Deployment: name=obo-prometheus-operator replicas=1 readyReplicas=1 availableReplicas=1
    framework.go:     condition: type=Available status=True reason=MinimumReplicasAvailable ...
    framework.go:   ... (6 deployments total)
    framework.go: Pods in namespace openshift-cluster-observability-operator: 7
    framework.go:   Pod: name=health-analyzer-85bf6b7b69-ctcrp phase=Running
    framework.go:     container=health-analyzer ready=true restarts=0 state=Running
    framework.go:   Pod: name=observability-operator-fc695d7-jbxj5 phase=Running
    framework.go:     container=operator ready=true restarts=0 state=Running
    framework.go:   ... (7 pods total)
    framework.go: Events in namespace openshift-cluster-observability-operator: 137
[...]

PR #1144, opened 2026-06-30
chore: refactor e2e test diagnostic

simonpasquier:
Follow-up of https://github.com/rhobs/observability-operator/pull/1115

### DIFF
--- a/pkg/apis/uiplugin/v1alpha1/types.go
+++ b/pkg/apis/uiplugin/v1alpha1/types.go
@@ -43,15 +43,35 @@ type UIPluginType string
 const (
 	// TypeDashboards deploys the Dashboards Dynamic Plugin for OpenShift Console.
 	TypeDashboards UIPluginType = "Dashboards"
-	// DistributedTracing deploys the Distributed Tracing Dynamic Plugin for the OpenShift Console
+
+	// TypeDistributedTracing deploys the Distributed Tracing Dynamic Plugin for the OpenShift Console
 	TypeDistributedTracing UIPluginType = "DistributedTracing"
-	// TroubleshootingPanel deploys the Troubleshooting Panel Dynamic Plugin for the OpenShift Console
+
+	// TypeTroubleshootingPanel deploys the Troubleshooting Panel Dynamic Plugin for the OpenShift Console
 	TypeTroubleshootingPanel UIPluginType = "TroubleshootingPanel"
-	// Monitoring deploys the Monitoring Plugin for the OpenShift Console
+
+	// TypeMonitoring deploys the Monitoring Plugin for the OpenShift Console
 	TypeMonitoring UIPluginType = "Monitoring"
 
 	// TypeLogging deploys the Logging View Plugin for OpenShift Console.
 	TypeLogging UIPluginType = "Logging"
+)
+
+const (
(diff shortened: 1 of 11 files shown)


test/e2e/uiplugin_cluster_health_analyzer_test.go
```diff
@@ -32,13 +31,13 @@ func clusterHealthAnalyzer(t *testing.T) {
...
 
-	f.DumpOnFailure(t, f.DebugNamespace(uiPluginInstallNS))
+	f.DumpOnFailure(t, f.DebugNamespaces(uiPluginInstallNS))
 
 	plugin := resetMonitoringUIPlugin(t)
 	err = f.K8sClient.Create(t.Context(), plugin)
 	assert.NilError(t, err, "failed to create monitoring UIPlugin")
 
-	f.DumpOnFailure(t, dumpUIPluginDebug(plugin.Name))
+	f.DumpOnFailure(t, f.DebugUIPlugin(plugin.Name))
```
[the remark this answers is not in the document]
simonpasquier: there's no need to dump the plugin data if it wasn't created.

pkg/apis/uiplugin/v1alpha1/types.go
```diff
@@ -43,17 +43,37 @@ type UIPluginType string
...
+	// TypeMonitoring deploys the Monitoring Plugin for the OpenShift Console
 	TypeMonitoring UIPluginType = "Monitoring"
 
 	// TypeLogging deploys the Logging View Plugin for OpenShift Console.
 	TypeLogging UIPluginType = "Logging"
 )
 
+const (
+	// DashboardsPluginName is the required name for the Dashboards Plugin resource.
+	DashboardsPluginName = "dashboards"
```
danielmellado, line 62: nit: DashboardsPluginName uses a PluginName suffix while all the other constants use just Name (DistributedTracingName, TroubleshootingPanelName, MonitoringName, LoggingName). Could you rename this to DashboardsName for consistency?

test/e2e/framework/framework.go
```diff
@@ -280,20 +280,20 @@ func (f *Framework) DumpOnFailure(t *testing.T, fns ...DebugFunc) {
 		if !t.Failed() {
 			return
 		}
+		t.Logf("# Diagnostics for %s (FAILED)", t.Name())
```
danielmellado, line 283: nit: Since DumpOnFailure can be called multiple times within the same test (e.g. clusterHealthAnalyzer calls it three times line 34, line 40, and inside createRuleNamespace), this header will be printed once per cleanup callback. Wouldn't that be a bit noisy? Feel free to disregard this as it's a personal preference.

danielmellado (commented):
LGTM, just some minor cosmetic nits

pkg/apis/uiplugin/v1alpha1/types.go
simonpasquier: good catch, I'd probably go with the PluginName suffix for all as it's more explanatory.

test/e2e/framework/framework.go
simonpasquier: this is printed only when the test fails in which case I prefer to have more information than less.

danielmellado (approved):
/lgtm

PR #1148, opened 2026-07-03
chore: log or return error

simonpasquier:
This commit removes the "log the error and return it to the caller" patterns. The recommended Go approach is to either log the error or return it to the caller but not both.


slashpai (approved):
/lgtm

PR #1151, opened 2026-07-09
ACM-37236: feat: onboard Thanos Operator into COO

coleenquadros:
## Summary

[ACM-37236](https://redhat.atlassian.net/browse/ACM-37236)

Onboard the Thanos Operator into COO following the same pattern as the Perses operator onboarding.

- Add `deploy/thanos/` directory with 5 CRDs (`monitoring.thanos.io/v1alpha1`), operator deployment, RBAC, and user-facing editor/viewer roles
- Wire into OLM bundle via `deploy/olm/kustomization.yaml`
- Register Thanos CRDs as owned in the CSV
- Add Go dependency on `github.com/thanos-community/thanos-operator` and register types in scheme
- Add `generate-thanos-op-crds` Makefile target
- Add default `thanos-operator` image in `cmd/operator/main.go`

### CRDs added

| Kind | API Group | Scope |
|---|---|---|
| ThanosQuery | monitoring.thanos.io | Namespaced |
| ThanosReceive | monitoring.thanos.io | Namespaced |
| ThanosCompact | monitoring.thanos.io | Namespaced |
| ThanosStore | monitoring.thanos.io | Namespaced |
| ThanosRuler | monitoring.thanos.io | Namespaced |

API support level: `TechPreviewNoUpgrade`

### Pattern followed

Same as Perses operator onboarding (commit 5ba0330):
- Local CRD files with kustomize annotations
- Dedicated deployment, service account, cluster role
- User-facing editor/viewer ClusterRoles per CRD
- Types registered in `pkg/operator/scheme.go` under OpenShift feature gate

Jira: ACM-37236

## Test plan

- [x] `go build ./...` passes
- [x] `kustomize build deploy/olm/` renders all Thanos resources correctly
- [x] `go test ./...` unit tests pass (e2e skipped — requires live cluster)
- [ ] Deploy to test cluster and verify CRDs are created

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -59,6 +59,7 @@ var defaultImages = map[string]string{
 	"ui-monitoring-pf6":            "quay.io/openshift-observability-ui/monitoring-console-plugin:v0.5.4",
 	"ui-monitoring":                "quay.io/openshift-observability-ui/monitoring-console-plugin:v1.0.0",
 	"perses":                       "quay.io/openshift-observability-ui/perses:v0.54.0",
+	"thanos-operator":              "quay.io/thanos/thanos-operator:main-2026-07-02-e6d1790",
 }
 
 func imagesUsed() []string {
--- a/deploy/thanos/thanos-operator-cluster-role.yaml
+++ b/deploy/thanos/thanos-operator-cluster-role.yaml
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: thanos-operator
+  name: thanos-operator-manager-role
 rules:
 - apiGroups:
   - ""
--- a/deploy/thanos/thanos-operator-cluster-role.yaml
+++ b/deploy/thanos/thanos-operator-cluster-role.yaml
@@ -1,0 +1,125 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: thanos-operator
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - persistentvolumeclaims
+  - secrets
+  - serviceaccounts
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  - statefulsets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - prometheusrules
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - monitoring.thanos.io
+  resources:
+  - thanoscompacts
+  - thanosqueries
+  - thanosreceives
+  - thanosrulers
+  - thanosstores
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - monitoring.thanos.io
+  resources:
+  - thanoscompacts/finalizers
+  - thanosqueries/finalizers
+  - thanosreceives/finalizers
+  - thanosrulers/finalizers
+  - thanosstores/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - monitoring.thanos.io
+  resources:
+  - thanoscompacts/status
+  - thanosqueries/status
+  - thanosreceives/status
+  - thanosrulers/status
+  - thanosstores/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - policy
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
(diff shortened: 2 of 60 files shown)


go.mod
```diff
@@ -131,6 +133,7 @@ require (
 	github.com/spf13/cobra v1.10.2 // indirect
 	github.com/spf13/pflag v1.0.10 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
+	github.com/thanos-community/thanos-operator v0.0.0-20260702133656-54890f01eac9 // indirect
```
danielmellado: `github.com/thanos-community/thanos-operator` is marked `// indirect`, but it's imported directly in `pkg/operator/scheme.go` (line 15). A `go mod tidy` should move it to the direct dependencies block.

deploy/thanos/thanos-operator-cluster-role.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - rolebindings
+  - roles
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
danielmellado, line 125: The deployment passes `--leader-elect` (see `thanos-operator-deployment.yaml:29`), but this ClusterRole is missing permissions for `coordination.k8s.io/leases`. Since controller-runtime v0.19+ defaults to lease-based leader election, the operator pod may crash-loop with a `forbidden` error on startup.

deploy/thanos/thanos-operator-cluster-role.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - serviceaccounts
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
danielmellado, line 21: Controllers need to emit Kubernetes events for status updates. The Perses ClusterRole includes `events` in the core API group. This one does not and I assume it's not inteded. Please let me know otherwise!. If not, the operator will log warnings or fail to record events at runtime.

I'd add `events` to the core API group resources, or as a separate rule:

deploy/thanos/thanos-operator-deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      annotations:
+        kubectl.kubernetes.io/default-container: thanos-operator
+      labels:
+        app.kubernetes.io/component: controller
+        app.kubernetes.io/name: thanos-operator
+        app.kubernetes.io/part-of: observability-operator
+    spec:
+      containers:
+      - name: thanos-operator
+        image: quay.io/thanos/thanos-operator:main-2026-07-02-e6d1790
```
danielmellado, line 27: `main-2026-07-02-e6d1790` is a nightly build from `main`, not a release tag. This is fine for TechPreview, but we should track switching to a proper release tag before moving toward GA. Is there a released version of thanos-operator we can pin to already?

cmd/operator/main.go
```diff
@@ -59,6 +59,7 @@ var defaultImages = map[string]string{
 	"ui-monitoring-pf6":            "quay.io/openshift-observability-ui/monitoring-console-plugin:v0.5.4",
 	"ui-monitoring":                "quay.io/openshift-observability-ui/monitoring-console-plugin:v1.0.0",
 	"perses":                       "quay.io/openshift-observability-ui/perses:v0.54.0",
+	"thanos-operator":              "quay.io/thanos/thanos-operator:main-2026-07-02-e6d1790",
```
danielmellado, line 62: nit: Same dev tag note as on the deployment YAML.

danielmellado (changes requested):
Overall this LGTM, but I think that there may be a few issues. See inline comments. Thanks!

simonpasquier:
We have future work (https://redhat.atlassian.net/browse/COO-1969) to remove all second-level operators (e.g. Prometheus, Perses, ...) from the OLM CSV because it proved to be too inflexible. I suggest that we apply this decision to Thanos in advance.

deploy/thanos/thanos-operator-cluster-role.yaml
coleenquadros: Thanks for pointing it out. Totally missed the role and rolebinding for this from my POC [PR](https://github.com/coleenquadros/multicluster-observability-addon/blob/9fd1c3a20f59cd4c2f15abf6f591bbf78e9995b6/internal/addon/manifests/charts/mcoa/charts/metrics/templates/thanos-operator/operator/thanos-operator-role.yaml)

deploy/thanos/thanos-operator-deployment.yaml
coleenquadros: thanos-community/thanos-operator has no tags or GitHub releases yet. Rhobs next points to nightly too  I think @saswatamcode or @philipgough  would be able to confirm

jan--f: This is fine for upstream. Downstream we replace those image links anyway.

jan--f:
> We have future work (https://redhat.atlassian.net/browse/COO-1969) to remove all second-level operators (e.g. Prometheus, Perses, ...) from the OLM CSV because it proved to be too inflexible. I suggest that we apply this decision to Thanos in advance.

I thought about the same, but I perhaps its better to not delay/block the thanos addition on that? We don't have a fix for all issues the bundle removal will cause and I doubt adding yet another operator to migrate adds significant overhead.

That is unless Coleen is interested in the work needed for [COO-1969](https://redhat.atlassian.net/browse/COO-1969). I'd be happy to have this here, though I wouldn't want to block this on [COO-1969](https://redhat.atlassian.net/browse/COO-1969).

jan--f (changes requested):
Overall this lgtm, except for one nit:
Some RBAC releated resources specify `namespace: default`, but the operator deployment doesn't. Given that there might a leader election iiuc this could fail if someone deploys via the kustomization file. The operator deployment could end up in a different namespace then the SA and CRB allowing the leader election.
The easiest would be to set a namespace in `kustomization.yaml`
We do this for prometheus operator here: 
https://github.com/rhobs/observability-operator/blob/27dfbb99c794941ece618dfae51516daf89b1438/deploy/dependencies/kustomization.yaml#L42

PR #1155, opened 2026-07-16
COO-1973: fix: use patch instead of merge when registering plugins in the console

jgbernalp:
This PR:

- uses patch instead of merge to avoid polluting the console CR when registering plugins
- removes the merger from the reconciler as is not used anymore


simonpasquier (commented):
/lgtm

simonpasquier (commented):
/lgtm

PR #1173, opened 2026-07-29
fix(COO-1863): pass TLS profile args to korrel8r deployment

alanconway:
Add --tls-min-version and --tls-cipher-suites to the korrel8r deployment.


alanconway:
@zhuje needs an approving review as well as /lgtm?

zhuje (approved):

PR #1181, opened 2026-08-04
COO-1860: feat: add default global prometheus datasource

jgbernalp:
Add a default global datasource pointing to the in-cluster Thanos querier. This matches the current behavior of the legacy dashboards included with CMO.


zhuje (approved):
/lgtm

PR #1184, opened 2026-08-06
COO-2081: apply readOnly filesystem security attribute

tremes:


simonpasquier (commented):
can we add an e2e test ensuring that all pods in the COO namespace comply with the invariant? I'm not sure if we already have a test which enables all possible components though...

PeterYurkovich:
> can we add an e2e test ensuring that all pods in the COO namespace comply with the invariant? I'm not sure if we already have a test which enables all possible components though...

Perses writes to the filesystem so it's pod will need an exemption from that test

simonpasquier:
> Perses writes to the filesystem so it's pod will need an exemption from that test

I assume that the written data isn't persisted. Would it be possible to use an empty dir instead? (this it the recommended approach to get rid of read-write root fs).

PeterYurkovich:
The written data is persisted in a PVC on the stateful set

simonpasquier:
@PeterYurkovich so enabling `readOnlyRootFilesystem` is safe.

jan--f:
Should this also be added to the bundle managed deployments via kustomize? That can certainly be a follow-up though the Jira ticket indicates all containers.

simonpasquier:
IMHO the best approach for the longer term is to have 1 test enabling/deploying all the COO components which iterates over all the pods in the COO namespace. Doing so would ensure that 1) new components are also covered (provided that the test deploys them ofc) and 2) we can easily validate other pod invariants if we need to. We did something similar for CMO in https://github.com/openshift/cluster-monitoring-operator/pull/3039

jgbernalp (approved):
/lgtm

PR #1186, opened 2026-08-07
OU-1212: feat: add mcp-overview flag to monitoring plugin

jgbernalp:


pkg/controllers/uiplugin/monitoring.go
```diff
@@ -152,6 +152,8 @@ func createMonitoringPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace, name, im
 		return pluginInfo, fmt.Errorf("all uiplugin monitoring configurations are invalid or not supported in this cluster version")
 	}
 
+	features = append(features, "mcp-overview")
```
PeterYurkovich, line 155: This is fine for testing, but I think we need to follow up in 2 way:

1. Create the monitoring UIPlugin CR by default on installation with this flag enabled. This may need to perform a patch on the CR if one already exists to avoid conflicting with ACM creating the CR automatically as well, only patching if the field doesn't exist. If it exists and is set to disabled then we do nothing
2. Expose this flag like the others so that it can be turned off

PR #1187, opened 2026-08-07
fix: add update RBAC for NetworkPolicy resources

simonpasquier:
The Conforma OCP 5.0 compliance policy requires explicitly the `update` permission too.


danielmellado (approved):
Thanks Simon!

PR #1190, opened 2026-08-13
OU-1472: remove "incident" feature flag from monitoring backend

PeterYurkovich:
Since the `monitoring-console-plugin` deployment handles the `incidents` and `cluster-health-analyzer` feature flags as the same thing this change updates the observability operator to only send the `cluster-health-analyzer` flag to reduce the need to maintain the `incidents` flag on the `monitoring-console-plugin` backend. 

This does not change the CR for the UIPlugin, as our team is planning to create a v1 or v1alpha2 version of the CR in the somewhat near future and this merging within the CR should occur there so we don't need to have a second webhook to maintain.

PR may look unrelated to the jira ticket, but it a part of the step to remove unneeded feature flags to create a better mapping between the features and the files being tested for automation to pick up on. Related PR openshift/monitoring-plugin#1127

### DIFF
--- a/pkg/controllers/uiplugin/monitoring.go
+++ b/pkg/controllers/uiplugin/monitoring.go
@@ -164,11 +164,7 @@ func createMonitoringPluginInfo(plugin *uiv1alpha1.UIPlugin, namespace, name, im
 		features = append(features, "perses-dashboards")
 		pluginInfo.PersesImage = persesImage
 	}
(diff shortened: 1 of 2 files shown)


etmurasaki:
tested with/without incidents feature. working fine.

zhuje (approved):
/lgtm

PR #1191, opened 2026-08-13
feat(COO-1965): add offline generator that shares reconciler code

alanconway:
Kustomize-style offline manifest generator (pkg/generator) that
transforms ObservabilityInstaller and UIPlugin CRs plus support files
into the full resource set the operator would reconcile.

- Extract GenerateInstallerObjects from the ObservabilityInstaller
  reconciler so the generator and reconciler produce the same objects.
- Expand UIPlugin CRs emitted by GenerateInstallerObjects
  into their concrete operands via GenerateUIPluginObjects, matching what
  the UIPlugin controller reconciles.
- Add a FallbackReader that serves Secrets with stringData as data so
  object storage secrets resolve offline.
- Emit OLM Namespace/OperatorGroup for namespaces the operator assumes
  OLM creates, and add a --skip-operators flag to omit them.
- Add golden-file tests pinning the generated output against the
  reconciled object set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,10 @@ operator: generate build
 build:
 	go build -o ./tmp/operator ./cmd/operator/...
 
+.PHONY: generator
+generator:
+	go build -o ./tmp/generator ./cmd/generator/...
+
 .PHONY: operator-image
 operator-image: generate
 	$(CONTAINER_RUNTIME) build -f build/Dockerfile . -t $(OPERATOR_IMG)
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ CATALOG_TEMP := $(shell mktemp -d)
 ## Development
 
 .PHONY: all
-all: lint test-unit operator-image bundle-image
+all: generate lint test-unit operator-image bundle-image
 
 .PHONY: test-unit
 test-unit:
@@ -117,6 +117,10 @@ operator: generate build
 .PHONY: build
 build:
 	go build -o ./tmp/operator ./cmd/operator/...
+
+.PHONY: generator
+generator:
+	go build -o ./tmp/generator ./cmd/generator/...
 
 .PHONY: operator-image
 operator-image: generate
--- a/cmd/generator/main.go
+++ b/cmd/generator/main.go
@@ -1,0 +1,7 @@
+package main
+
+import "github.com/rhobs/observability-operator/pkg/generator"
+
+func main() {
+	generator.Main()
+}
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -20,12 +20,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"maps"
 	"os"
 	"slices"
 
 	configv1 "github.com/openshift/api/config/v1"
 	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
-	obopo "github.com/rhobs/obo-prometheus-operator/pkg/operator"
 	"go.uber.org/zap/zapcore"
 	k8sflag "k8s.io/component-base/cli/flag"
 	"k8s.io/utils/ptr"
@@ -33,60 +33,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	"github.com/rhobs/observability-operator/pkg/images"
 	"github.com/rhobs/observability-operator/pkg/operator"
 )
 
-// The default values we use. Prometheus and Alertmanager are handled by
-// prometheus-operator. For thanos we use the default version from
-// prometheus-operator.
-var defaultImages = map[string]string{
-	"prometheus":                   "",
-	"alertmanager":                 "",
-	"thanos":                       obopo.DefaultThanosImage,
-	"ui-troubleshooting-panel-pf6": "quay.io/openshift-observability-ui/troubleshooting-panel-console-plugin:v0.4.5",
-	"ui-troubleshooting-panel":     "quay.io/openshift-observability-ui/troubleshooting-panel-console-plugin:v1.0.0",
-	"ui-distributed-tracing-pf4":   "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v0.3.3",
-	"ui-distributed-tracing-pf5":   "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v0.4.3",
-	"ui-distributed-tracing-pf6":   "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v1.0.3",
-	"ui-distributed-tracing":       "quay.io/openshift-observability-ui/distributed-tracing-console-plugin:v1.1.0",
-	"ui-logging-pf4":               "quay.io/openshift-observability-ui/logging-view-plugin:v6.0.5",
-	"ui-logging-pf5":               "quay.io/openshift-observability-ui/logging-view-plugin:v6.1.6",
-	"ui-logging":                   "quay.io/openshift-observability-ui/logging-view-plugin:v6.2.1",
-	"korrel8r":                     "quay.io/korrel8r/korrel8r:0.11.1",
-	"health-analyzer":              "quay.io/openshiftanalytics/cluster-health-analyzer:v1.1.1",
-	"ui-monitoring-pf5":            "quay.io/openshift-observability-ui/monitoring-console-plugin:v0.4.5",
-	"ui-monitoring-pf6":            "quay.io/openshift-observability-ui/monitoring-console-plugin:v0.5.4",
-	"ui-monitoring":                "quay.io/openshift-observability-ui/monitoring-console-plugin:v1.1.0",
-	"perses":                       "quay.io/openshift-observability-ui/perses:v0.54.1",
-}
-
-func imagesUsed() []string {
-	i := 0
-	imgs := make([]string, len(defaultImages))
-	for k := range defaultImages {
-		imgs[i] = k
-		i++
+func validateImages(overrides *k8sflag.MapStringString) (map[string]string, error) {
+	if overrides.Empty() {
+		return images.Validate(nil)
 	}
-	slices.Sort(imgs)
-	return imgs
-}
-
-// validateImages merges the passed images with the defaults and checks if any
-// unknown image names are passed. If an unknown image is found, this raises an
-// error.
-func validateImages(images *k8sflag.MapStringString) (map[string]string, error) {
-	res := defaultImages
-	if images.Empty() {
-		return res, nil
-	}
-	imgs := *images.Map
-	for k, v := range imgs {
-		if _, ok := res[k]; !ok {
-			return nil, fmt.Errorf("image %v is unknown", k)
-		}
-		res[k] = v
-	}
-	return res, nil
+	return images.Validate(*overrides.Map)
 }
 
 func main() {
@@ -100,12 +55,12 @@ func main() {
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
-	images := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
+	imageOverrides := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
-	flag.Var(images, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", imagesUsed()))
+	flag.Var(imageOverrides, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", slices.Sorted(maps.Keys(images.DefaultImages))))
 	flag.BoolVar(&openShiftEnabled, "openshift.enabled", false, "Enable OpenShift specific features such as Console Plugins.")
 	flag.StringVar(&otelCSVName, "opentelemetry-csv", "", "OpenTelemetry Operator starting CSV name. This can be used to install a specific OpenTelemetry Operator version. Empty string means the latest version will be installed.")
 	flag.StringVar(&tempoCSVName, "tempo-csv", "", "Tempo Operator starting CSV name. This can be used to install a specific Tempo Operator version. Empty string means the latest version will be installed.")
@@ -122,11 +77,11 @@ func main() {
 	setupLog.Info("running with arguments",
 		"namespace", namespace,
 		"metrics-bind-address", metricsAddr,
-		"images", images,
+		"images", imageOverrides,
 		"openshift.enabled", openShiftEnabled,
 	)
 
-	imgMap, err := validateImages(images)
+	imgMap, err := validateImages(imageOverrides)
 	if err != nil {
 		setupLog.Error(err, "cannot create a new operator")
 		os.Exit(1)
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -21,11 +21,9 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"slices"
 
 	configv1 "github.com/openshift/api/config/v1"
 	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
-	obopo "github.com/rhobs/obo-prometheus-operator/pkg/operator"
 	"go.uber.org/zap/zapcore"
 	k8sflag "k8s.io/component-base/cli/flag"
 	"k8s.io/utils/ptr"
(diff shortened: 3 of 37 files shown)


pkg/generator/generator.go
```diff
@@ -0,0 +1,594 @@
...
+	})
+	exitErr(err, "error generating resources")
+
+	if stdout != "" {
+		fmt.Print(stdout)
+	}
+	if warnings != "" {
+		fmt.Fprintln(os.Stderr, warnings)
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
alanconway: Goal is to collect warnings without aborting, then indicate failure on exit.

cmd/operator/main.go
```diff
@@ -100,12 +53,12 @@ func main() {
 
 		setupLog = ctrl.Log.WithName("setup")
 	)
-	images := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
+	imageOverrides := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
-	flag.Var(images, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", imagesUsed()))
```
pavolloffay, line 108: is it worth making this breaking change?

hack/uninstall.sh
```diff
@@ -0,0 +1,315 @@
...
+    [[ "$answer" =~ ^[Yy]$ ]] || { log "Aborted."; exit 0; }
+fi
+
+# -------------------------------------------------------------------
+# Phase 1: Delete custom resources (operands first, then operator CRs)
+# -------------------------------------------------------------------
+log "=== Phase 1: Delete custom resources ==="
+
+# ObservabilityInstaller-managed operands
+for cr in opentelemetrycollectors.opentelemetry.io tempostacks.tempo.grafana.com; do
```
pavolloffay: It should be enough to delete the observability installer as it has owner references to all created objects. Otherwise this deletes other objects in the cluster (e.g. directly created by the user).

pkg/controllers/observability/generate.go
```diff
@@ -0,0 +1,86 @@
...
+	if tracing.Enabled {
+		return true
+	}
+	return tracing.GetOperators() != nil && tracing.GetOperators().Install != nil && *tracing.GetOperators().Install
+}
+
+// GenerateAllInstallerObjects returns the universe of operand objects across all
+// possible capabilities. The reconciler uses this to know which objects to clean
+// up when capabilities are disabled.
+func GenerateAllInstallerObjects(ctx context.Context, k8sClient client.Reader, k8sReader client.Reader, instance *obsv1alpha1.ObservabilityInstaller, opts Options) ([]client.Object, error) {
```
pavolloffay: why the reader is duplicated?

pkg/controllers/observability/generate.go
```diff
@@ -0,0 +1,86 @@
...
+	allInstance.Spec.Capabilities.Tracing.Enabled = true
+	return GenerateInstallerObjects(ctx, k8sClient, k8sReader, allInstance, opts, false, true)
+}
+
+// GenerateInstallerObjects builds the object set that reconciles an ObservabilityInstaller.
+//
+// It is called by the reconciler and the offline generator so both produce the same objects.
+// Subscriptions are included if the CR requests operator installation.
+// When installOperators is false, Subscriptions are omitted; when installResources is false, operand resources are omitted.
+func GenerateInstallerObjects(ctx context.Context, k8sClient client.Reader, k8sReader client.Reader, instance *obsv1alpha1.ObservabilityInstaller, opts Options, installOperators, installResources bool) ([]client.Object, error) {
```
pavolloffay: the same as before - the reader is twice. 

`installOperators, installResources bool` - these options are directly in the `instance *obsv1alpha1.ObservabilityInstaller` so it seems redundant

pkg/controllers/observability/generate_test.go
```diff
@@ -0,0 +1,38 @@
+package observability
```
pavolloffay, line 1: I would expect a test here to actually check which objects are returned

go.mod
```diff
@@ -43,6 +43,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.24.1
 )
 
+require sigs.k8s.io/yaml v1.6.0
```
pavolloffay, line 46: should be moved below

cmd/generator/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "github.com/rhobs/observability-operator/pkg/generator"
+
+func main() {
```
pavolloffay, line 5: I don't see how the generator is wired to the operator cmd. 

For distribution/productization it would be simpler if the generator was a subcomand so that we don't have to build/productize another image.

pkg/generator/generator_parity_test.go
```diff
@@ -0,0 +1,359 @@
...
+	scheme := operator.NewScheme(&operator.OperatorConfiguration{
+		FeatureGates: operator.FeatureGates{
+			OpenShift: operator.OpenShiftFeatureGates{
+				Enabled: true,
+				Version: testClusterVersion,
+			},
+		}})
+
+	yamlData, err := readInputs([]string{testdataDir})
+	assert.NilError(t, err)
```
pavolloffay: require might be better suited for this

pkg/generator/generator_parity_test.go
```diff
@@ -0,0 +1,359 @@
...
+// This test drives the
+// real reconcile path: observability.GenerateAllInstallerObjects feeds
+// reconciler.Updater / CreateUpdateReconciler, whose Reconcile methods add
+// common labels and apply (server-side apply) the resources; UIPlugin operands
+// are applied the same way the UIPlugin controller applies them (a
+// reconciler.Updater per operand). This catches drift between the operator and
+// generator that the shared functions cannot, e.g. in label handling.
+//
+// The two sides are not byte-identical by design:
+//   - The generator emits the COO Namespace and OperatorGroup, which OLM creates
```
pavolloffay: why the generator even creates COO namespace and operator group? I would expect the generator to do not create any namespace and operatorgroup with subscription if install operators is enabled in the installer CR

pkg/generator/generator.go
```diff
@@ -0,0 +1,485 @@
...
+	// Images overrides default images, keyed by image name.
+	Images map[string]string
+	// K8sClient, when set, is used to fetch version, TLS profile and secrets.
+	K8sClient client.Client
+	// TLSProfile overrides the TLS profile used for UIPlugin operands.
+	TLSProfile configv1.TLSProfileSpec
+	// SaveDir, when set, writes each intermediate resource to a file in the
+	// directory. OutputDir, when set, writes each output resource to a file in
+	// the directory instead of the returned manifest.
+	SaveDir   string
```
pavolloffay: I find this redundant for a CLI. Users can pipe the output to a file in shell.

pkg/generator/generator.go
```diff
@@ -0,0 +1,485 @@
...
+		}
+		for _, update := range cv.Status.History {
+			if update.State == configv1.CompletedUpdate {
+				clusterVersion = update.Version
+				break
+			}
+		}
+	}
+
+	scheme := operator.NewScheme(&operator.OperatorConfiguration{
```
pavolloffay, line 106: nit: maybe move to a var definition?

pkg/controllers/observability/tempo_components.go
```diff
@@ -122,7 +122,7 @@ type tempoSecrets struct {
 	objectStorageCAConfigMap *corev1.ConfigMap
 }
 
-func tempoStackSecrets(ctx context.Context, k8sClient client.Client, k8sReader client.Reader, instance obsv1alpha1.ObservabilityInstaller) (*tempoSecrets, error) {
```
jan--f, line 125: Is this change intentional?

jan--f (commented):
The technical approach seems sound. Sharing the asset generation is a good choice. Here and there some weird artifacts seem to have crept in (duplicated reader objects, changing a client object to a reader).

I'd like to have a more general discussion about how we expect this to be used. A PR is perhaps not ideal but let me lay out my questions:
Iiuc the goal is to allow users to transition from operator managed setups to user-owned. The generator code allows to synthesize the resources either against a cluster or fully offline. The later however requires additional arguments iiuc.

If a user runs this against a cluster, they already have the resources (though in the cluster). We can facilitate a user taking over those resources, perhaps with less complexity? The approach would be via owner references.

I don't quite follow the offline scenario requiring this (and following) refactors. How is the offline variant intended to be used and what inputs are required?

_edit_ Also the MonitoringStack is missing from this..?

cmd/generator/main.go
alanconway: generator and operator aren't connected, they just share code. I could make the generator a subcommand, I worried that it might confuse/distract from the role of the operator. Not hard to make it a subcommand, if you feel strongly we can discuss it.

pkg/generator/generator_parity_test.go
alanconway: They are required for the generator to create a functional deployment with operators. When the operator is installed, OLM creates the namespace and operatorgroup, but with the generator they don't exist.

pkg/generator/generator.go
alanconway: Agreed, I can drop it.

pkg/controllers/observability/tempo_components.go
alanconway: nope will fix

pkg/generator/generator.go
alanconway: why, what difference would that make?

hack/uninstall.sh
alanconway: The goal here is to handle reconciliation where some existing objects are  no longer needed, not just deleting the installer. Since the total set of resources that we may have created is now implicit in the config/ tree, we generate the total set, the desired set and delete the difference. These are only objects that the operator creates (or could create)

cmd/operator/main.go
alanconway: I moved the image constants to a shared package called "images" so can' t use that as a global variable name anymore. It's just a variable name change, not a "breaking" change.

pkg/generator/generator_parity_test.go
alanconway: You're right about the operatorgroup - the generator doesn't install the COO, so doesn't need to install its OG.
The namespace is required - with the operator it would be installed by OLM, but the generator has to create the namespace itself for plugin resources etc.
Will fix the OG

alanconway:
@jan--f 
> I don't quite follow the offline scenario requiring this (and following) refactors. How is the offline variant intended to be used and what inputs are required?

Mainly it's a development tool
- deploy obs. *without* operator, so dev can can patch/devspace resources without being overwritten on reconcile
- testing, developing new overlays: 
  - round trip to code change/check yaml is faster than real deploy
  - compare new output with previous output via simple diff
  -  or via cluster deployment: `go run ./cmd/generator/ -f ~/tmp/sample -cluster | oc diff -f -`

Maybe has production uses - theoretical so far
- constrained environments (edge) that want minimal operators
- "cookie cutter" multi-cluster deployments that can be statically defined
- interop with ACM and other multi-cluster tools that consume YAML

> _edit_ Also the MonitoringStack is missing from this..?
Intentional, want to work the kinks out of the concept first before applying it more generally.

alanconway:
@jan--f @pavolloffay ready for re-review

alanconway:
@jan--f @pavolloffay addressed all comments, need re-review and LGTM

hack/uninstall.sh
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash
+#
+# Uninstall observability-operator and all its managed resources from a cluster.
+# Works with both OpenShift (oc) and plain Kubernetes (kubectl).
+#
+# Usage: hack/uninstall.sh
```
pavolloffay: nit: this is not directly related to this PR objective. Perhaps can be added separately and hooked up to the makefile. 

When I was working on COO I found it complicated to get started and deploy it on OCP with all the required OCP configs.

pkg/generator/fallback_reader.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// FallbackReader implements client.Reader with a set of pre-loaded objects.
+// Get checks the pre-loaded set first, then falls back to the underlying Reader.
+type FallbackReader struct {
```
pavolloffay, line 14: Isn't there standard/upstream client which could be used in tests?

pavolloffay (approved):
Sorry for the delay. I was OOO last week. 

Overall LGTM - the installer CRD parts, just minor comments. Someone else from COO should also review/approve. 

I like the generator 👍🏼 We should think how we are going to distribute it. I would suggest making it a subcommand of the operator as I see it as the simplest way to distribute it without adding another build and productization.

pkg/generator/fallback_reader.go
alanconway: Not that I know of - this reader isn't part of a test, it's combining locally created objects with the view of the cluster. The generator is normally works with all local objects, but with the "-cluster" flag it can also check the cluster. This type provides the combined view. If there is such a client I'll use it - I had the same thought.

alanconway:
@pavolloffay I removed the uninstall script, can you re-LGTM?

pavolloffay (approved):
Adding back LGTM, regarding the ObservabilityInstaller parts. It would be also great to get review from other COO maintainers for the other parts.

Makefile
```diff
@@ -21,7 +21,7 @@ CATALOG_TEMP := $(shell mktemp -d)
 ## Development
 
 .PHONY: all
-all: lint test-unit operator-image bundle-image
+all: generate lint test-unit operator-image bundle-image
```
simonpasquier, line 24: not needed?

alanconway: Generating is a necessary step so I think should be part of an "all" target, otherwise forgetting to re-generate is a risk. We could improve the dependency tracking in the makefile to run generation as needed, but that's a bigger job.

simonpasquier: generate is already a pre-req of operator-image.

pkg/controllers/uiplugin/controller.go
```diff
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	obsv1alpha1 "github.com/rhobs/observability-operator/pkg/apis/observability/v1alpha1"
```
simonpasquier, line 31: it feels backwards that the uiplugin controller depends on the observabilityinstaller API.

I tried the generator locally and I noticed a Deployment for the distributed-tracing plugin, why is that? My expectation is that the generator produces only the first-level custom resources (e.g. UIPlugin).

jan--f:
I'm confused, I thought we were going to revisit this at a later time?

pkg/controllers/uiplugin/controller.go
alanconway: The goal of the generator is to produce a complete working set of operands that **does not require the COO** to reconcile them, so they can be deployed on clusters without the COO or its CRDs.
Internally the generator creates a temporary UIPlugin, then expands it into its operands and deletes the UIPlugin. 
I'll look into the backwards dependency, probably it's soemthing that can be moved to util...

alanconway:
> I'm confused, I thought we were going to revisit this at a later time?

I meant to revisit the re-factoring to use kustomize internally at a later time.

The generator has been re-written on the existing codebase, is complete bar review, and is a useful development tool if nothing else. It seems worth getting it in, but I defer to your judgement if you think its better to delay.

pkg/controllers/uiplugin/controller.go
alanconway: Good catch about the dependency, there was some AI slop there that I did not catch. Will fix and review more carefully.

Makefile
alanconway: Missed that, removed from all.

pkg/controllers/uiplugin/controller.go
alanconway: Fixed

simonpasquier:
> The goal of the generator is to produce a complete working set of operands that does not require the COO to reconcile them, so they can be deployed on clusters without the COO or its CRDs.
Internally the generator creates a temporary UIPlugin, then expands it into its operands and deletes the UIPlugin.

Sorry but I didn't understand the goal then. My initial assumption was that the generator would take an ObservabilityInstaller resource and generate the first-level custom resources from it (e.g. COO + Tempo + Loki resources to start with). The user could tweak each resource to their needs (for instance customizing pod scheduling) then apply the YAML manifests in their cluster. I don't see why we would want to remove COO from the picture.

alanconway:
> > The goal of the generator is to produce a complete working set of operands that does not require the COO to reconcile them, so they can be deployed on clusters without the COO or its CRDs.
> > Internally the generator creates a temporary UIPlugin, then expands it into its operands and deletes the UIPlugin.
> 
> Sorry but I didn't understand the goal then. My initial assumption was that the generator would take an ObservabilityInstaller resource and generate the first-level custom resources from it (e.g. COO + Tempo + Loki resources to start with). The user could tweak each resource to their needs (for instance customizing pod scheduling) then apply the YAML manifests in their cluster. I don't see why we would want to remove COO from the picture.

It makes more sense as part of the larger kustomization refactor - the generator exposes the kustomizations that are being used by the operator internally, which is useful for designing new ones. You can't see that in the current version.
It has value for development and testing - it's useful to be able to generate offline resources and compare them with other offline resources or "oc diff" to see what changes in different versions of the operator.
It's also handy in testing to deploy without the operator, and avoid the need to scale down to avoid interfering with devspace or other image modifications. 
It may have value in some resource constrained settings where you want to avoid an operator, or in inter-operating with other YAML based deployment tools that don't want to use the operator.

I extracted it to get an independent part of the kustomization work that could be reviewed and merged separately.

PR #1201, opened 2026-08-20
COO-2340: remove nonroot SCC to have PSA restricted

tremes:
This is an attempt to "increase" PSA level of the COO namespace to `restricted`. 

```
    pod-security.kubernetes.io/audit: restricted
    pod-security.kubernetes.io/warn: restricted
```


simonpasquier (commented):
/lgtm

PR #1210, opened 2026-08-31
OBSINTA-1588: feat/add liveness and readiness probes to health-analyzer deployment

rioloc:
This PR configures liveness and readiness probes in cluster-health-analyzer Deployment resource

To be merged after https://github.com/openshift/cluster-health-analyzer/pull/118
Assisted-By: Claude

### DIFF
--- a/pkg/controllers/uiplugin/health_analyzer.go
+++ b/pkg/controllers/uiplugin/health_analyzer.go
@@ -182,6 +182,30 @@ func newHealthAnalyzerDeployment(namespace string,
 									Name:          "metrics",
 								},
 							},
+							LivenessProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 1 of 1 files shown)


jgbernalp:
@rioloc we need to update the upstream image which includes https://github.com/openshift/cluster-health-analyzer/pull/118 in https://github.com/rhobs/observability-operator/blob/main/cmd/operator/main.go#L56

rioloc:
@jgbernalp liveness/readiness endpoint on cluster-health-analyzer server are already in the upstream image currently released (they were already served by https://github.com/openshift/cluster-health-analyzer/blob/c93d56f9396a27508ffc69b6943d222d884e192a/cmd/serve/apiserver.go#L28 ). I've already tested the changes in this PR and they work with current image.

Latest changes impacted only `pkg/mcp/*` which is not handled by COO, so no need to backport that for the scope of this PR.

jgbernalp (approved):
/lgtm

PR #1211, opened 2026-08-31
refactor: use kustomize overlays for ObservabilityInstaller and UIPlugin

alanconway:
Replace Go struct resources with kustomize overlay building in controllers.

See config/README.md and docs/design/kustomize-function-migration.md for more.

Add embedded kustomize component trees under config/ for
- ObservabilityInstaller (operator subscriptions and CRD instances)
- UIPlugin types (dashboards, distributed-tracing, logging, troubleshooting-panel, korrel8r).

Each capability is split into
- operator/ (namespaces, operatorgroups, subscriptions)
- resources/ (CRD instances, RBAC) directories.

Parity tests verify that:
- generator output matches "golden output" based on COO 1.5.1
- generator and operator output match each other.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


jan--f (commented):
I guess this is fine to merge into a feature branch, though I won't claim to have reviewed 5.4k added lines.
An LLM review found an issue that I think breaks the monitoring plugin. The Monitoring plugin is annotated with `service.beta.openshift.io` whereas all others use service.alpha.openshift.io`. The overlay replacement targets `service.alpha.openshift.io`.

Surely a small bug but this highlights a major drawback of this direction: we are moving to an asset generation model that isn't type checked at compile time and rely only on testing to ensure we produce the correct assets (which missed the bug outlined above).

So far I think this feature is aimed at developers, not users. We can already get the resources from a running cluster. I would recommend we reconsider this approach here.

alanconway:
> I guess this is fine to merge into a feature branch, though I won't claim to have reviewed 5.4k added lines.
I can split this into uiplugins first, then obs installer if that would help.

> An LLM review found an issue that I think breaks the monitoring plugin. The Monitoring plugin is annotated with `service.beta.openshift.io` whereas all others use service.alpha.openshift.io`. The overlay replacement targets `service.alpha.openshift.io`.
Will fix.

> Surely a small bug but this highlights a major drawback of this direction: we are moving to an asset generation model that isn't type checked at compile time and rely only on testing to ensure we produce the correct assets (which missed the bug outlined above).

Type checking doesn't prevent importing alpha vs beta packages by mistake. You can get type checking with `generator | kubectl apply -f - --dry-run` or similar, that can be added to the tests. I think the benefits of exposing the YAML  outweigh the type checking we get from Go structs.

> So far I think this feature is aimed at developers, not users. We can already get the resources from a running cluster. I would recommend we reconsider this approach here.

Yes, at developers who can access or provide YAML definitions easily, but can't easily modify the COO source code. The goal is to give us more flexibility to maintain a body of "sound configurations" in an accessible format - readable and writable.

Thoughts?

PR #1212, opened 2026-09-01
OU-1510: feat: upgrade perses operator to support read only filesystem

jgbernalp:
- Update the upstream images for perses
- Update the perses operator dependency to include the read only file system configuration


tremes:
Perhaps we could also remove https://github.com/rhobs/observability-operator/blob/main/test/e2e/all_components_test.go#L139

PR #1216, opened 2026-09-03
NO-JIRA: feat: default --openshift.enabled=true for openshift cluster

alanconway:
Auto detect openshift cluster and set flag default accordingly.
Flag still exists and takes priority if explicitly set.

/cc @simonpasquier


cmd/operator/main.go
```diff
@@ -100,13 +99,13 @@ func main() {
...
 	)
-	images := k8sflag.NewMapStringString(ptr.To(make(map[string]string)))
+	images := k8sflag.NewMapStringString(new(map[string]string))
 
 	flag.StringVar(&namespace, "namespace", "default", "The namespace in which the operator runs")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-bind-address", ":8081", "The address the health probe endpoint binds to.")
 	flag.Var(images, "images", fmt.Sprintf("Full images refs to use for containers managed by the operator. E.g thanos=quay.io/thanos/thanos:v0.33.0. Images used are %v", imagesUsed()))
-	flag.BoolVar(&openShiftEnabled, "openshift.enabled", false, "Enable OpenShift specific features such as Console Plugins.")
+	flag.BoolVar(&openShiftEnabled, "openshift.enabled", isOpenShift(), "Enable OpenShift specific features such as Console Plugins.")
```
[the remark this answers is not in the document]
alanconway: Deliberate. Setting before parse means that "operator -h" will correctly show what the default is. Any failures in auto-detect before parse are harmless, they just result in the default being false, which is correct.

cmd/operator/main.go
```diff
@@ -139,32 +140,26 @@ func main() {
...
+		openshiftVersion      string
+	)
+	if openShiftEnabled || !openShiftExplicit {
+		initialTLSProfileSpec, openshiftVersion, err = fetchOpenShiftData(ctx, setupLog)
+		if err != nil && openShiftExplicit {
+			setupLog.Error(err, "not an OpenShift cluster")
+			os.Exit(1) // Fail if enabled explicitly.
 		}
-		openshiftVersion = clusterVersion.Status.Desired.Version
+		openShiftEnabled = (err == nil)
```
jan--f: Not sure we want to treat any error from the openshift client as "not openshift". This could mean intermittent errors in the API Server can start the operator in non-OpenShift mode.

jan--f: This is falling back is also not logged. So running on openshift with auto mode with spotty api service will be rather tricky to debug.

alanconway: Good point. Fixed as follows:
```
	var (
		initialTLSProfileSpec configv1.TLSProfileSpec
		openshiftVersion      string
	)
	if openShiftEnabled || !openShiftExplicit {
		initialTLSProfileSpec, openshiftVersion, err = fetchOpenShiftData(ctx, setupLog)
		if err != nil {
			if apierrors.IsNotFound(err) {
				setupLog.Info("OpenShift API resources not found, running in non-OpenShift mode")
				openShiftEnabled = false
			} else {
				setupLog.Error(err, "OpenShift detection failed with unexpected error")
				os.Exit(1)
			}
		} else {
			openShiftEnabled = true
		}
	}
```

alanconway:
@jan--f fixed your comment, can you LGTM now?

cmd/operator/main.go
```diff
@@ -207,3 +209,22 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func fetchOpenShiftData(ctx context.Context, setupLog logr.Logger) (configv1.TLSProfileSpec, string, error) {
```
simonpasquier: It'd better to have an internal struct (`OpenShiftCluster` for instance) which stores TLSProfileSpec + version (multiple returned values are harder to grok and it'd be easier to extend).

cmd/operator/main.go
```diff
@@ -139,32 +141,32 @@ func main() {
...
-			os.Exit(1)
-		}
-
-		initialTLSProfileSpec, err = openshifttls.FetchAPIServerTLSProfile(ctx, directClient)
+	var (
+		initialTLSProfileSpec configv1.TLSProfileSpec
+		openshiftVersion      string
+	)
+	if openShiftEnabled || !openShiftExplicit {
+		initialTLSProfileSpec, openshiftVersion, err = fetchOpenShiftData(ctx, setupLog)
```
simonpasquier: can't we just always retrieve the OpenShift specific bits? If `--openshift.enabled` is true then any error is a failure otherwise "not found" means that we run in vanilla kubernetes.

alanconway: openshift enabled can be set explicitly false `--openshift.enabled=false` in which case we shouldn't try any openshift-specific operations.

cmd/operator/main.go
alanconway: Done.

cmd/operator/main.go
```diff
@@ -139,32 +143,32 @@ func main() {
...
+		case openShiftExplicit:
+			// --openshift.enabled=true was requested, honour it: any failure is fatal.
+			setupLog.Error(err, "failed to fetch OpenShift cluster data")
 			os.Exit(1)
-		}
-
-		initialTLSProfileSpec, err = openshifttls.FetchAPIServerTLSProfile(ctx, directClient)
-		if err != nil {
-			setupLog.Error(err, "failed to fetch TLS profile from cluster")
+		case apierrors.IsNotFound(err) || meta.IsNoMatchError(err) || runtime.IsNotRegisteredError(err):
```
simonpasquier, line 156: not sure why we need the 3 tests?

cmd/operator/main.go
simonpasquier: IMHO it would simplify the code to always try querying the OpenShift API but your call.

cmd/operator/main.go
alanconway: Distinguish between:
1. Explicilty requested openshift.enabled=true, fatal if we're not on openshift
2. Not explicitly requested, failure to find ocp resources - we are on a non-openshift cluster and it's OK.
3. Error is not "failed to find", cluster has a problem (regardless of whether its openshift or not) - fatal.
Last 2 cases distinguish between "failed to find openshift resources but cluster seems OK" and "cluster has a fatal problem, other than lack of openshift resources" Previous verions of this code didn't make that distinction and assumed any error meant "not on openshift".

cmd/operator/main.go
```diff
@@ -139,32 +143,32 @@ func main() {
...
 
-	var initialTLSProfileSpec configv1.TLSProfileSpec
-	var openshiftVersion string
-	if openShiftEnabled {
-		scheme := operator.NewOpenShiftScheme()
-		directClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{Scheme: scheme})
-		if err != nil {
-			setupLog.Error(err, "failed to create client for TLS profile fetch")
+	var data OpenShiftData
+	if openShiftEnabled || !openShiftExplicit {
```
jan--f: Wouldn't this try to fetch openshift data if I explicitly disable openshift?
`false || !false`

Would it be easier to follow the decision making if the decision would happen in two separate if statements? If `openShiftEnabled`, try to retrieve OpenshiftData. Then make a decision based in the retrievalresult and `openShiftExplicit`?

alanconway: No, explicit disable is openshiftEnabled=false, openshiftExplicit=true. 
We need to retrieve data if openshift is set explicilty (openshiftEnabled=true) or if the flag was not set explicitly (openshiftExplicit=false) Not sure how to make this simpler without repeating the fetch code.

cmd/operator/main.go
jan--f: Not sure I follow this paricular case. Afaiu `NotFound` means the openshift API exists, but the requested resource was not found. `NoMatch` means the API doesn't exist and we have a vanillar k8s.

If my understanding is correct, should we run with `openShiftEnabled = false` in this case?

cmd/operator/main.go
jan--f: Yeah I see. Perhaps trying to retrieve openshift data first and then making a decision based on the flags and such can make this clearer. 

But regardless, pulling this into a function and having unit tests for this would be helpful I think.

cmd/operator/main.go
simonpasquier: my question was about the specific case line, not the whole switch statement. IMHO the "not found" error should be considered a hard error (there should be no Openshift cluster without the resources). And only when getting "the resource doesn't exist in this cluster" error, it should fallback to vanilla Kubernetes behavior.

alanconway: Those errorrs are not hard errors if the cluster is not openshift.  The default behavior is to detect if the cluster is openshift and if it is use openshift features, but to **continue** without openshift features if its not.

alanconway:
I have made the logic more explicit - switch over set-true/set-false/not-set - and broken it out into functions with unit tests. It does read more easily I think - @simonpasquier @jan--f  ? Any chance of an LGTM?

PR #1225, opened 2026-09-11
chore: add make commit-lint target for local commit message validation

### LINKED ISSUE #305: Add make commit-lint target to help contributor validate commit message locally before submitting
Often PRs are blocked because the commit-lint action fails - e.g. https://github.com/rhobs/observability-operator/pull/304 fails
because of commit-lint https://github.com/rhobs/observability-operator/actions/runs/5377063687/jobs/9760871220 

It would reduce rework and toil if we provided a `make` target - `commit-lint` which allows contributors to validate commit messages locally. 

It does help if all they have to run is `make lint` to ensure all lints are caught before submission.

coffeegoesincodecomesout:
Closes #305

  ## Summary

  - Adds a `commit-lint` Make target so contributors can validate commit messages locally before pushing
  - Hooks it into `make lint` so it runs alongside the existing Go and shell linters
  - Documents the target in the commit message convention section of `docs/developer.md`
  
  ## Approach

  The comment on #305 suggests an approach that calls `commitlint` as a direct binary, which requires
  it to be installed first (e.g. `npm install -g @commitlint/cli`). This PR uses `npx` instead — a
  tool that ships with `npm`, which is already a documented prerequisite for this project. `npx`
  self-installs `@commitlint/cli` on first use with no extra setup required. 
  
  The target uses `--from $(git merge-base HEAD origin/main) --to HEAD`, which checks all commits in
  the branch since branching from `main`. This matches what `wagoid/commitlint-github-action` does in
  CI, so running `make commit-lint` locally gives you the same result you'd get from the PR check. 

  ## Output examples
  
  **Good commit** (`chore: add make commit-lint target...`) — exits 0, no output (silent pass).

  **Bad commit** (`this is a bad commit message`) — exits non-zero and prints:
  ⧗   --- input ---
  this is a bad commit message
  ✖   subject may not be empty [subject-empty]
  ✖   type may not be empty [type-empty]

  ✖   found 2 problems, 0 warnings


simonpasquier (commented):
/approve
/lgtm
/ok-to-test

PR #1229, opened 2026-09-14
docs: add user guide for authenticated access to Thanos Querier using…

### LINKED ISSUE #1228: Provide user doc example for authenticated access to metrics
Provide a documentation example of authenticated access to metrics using oauth-proxy to delegated subject access requests.

coffeegoesincodecomesout:
Authenticated access to metrics using oauth and subject access reviews 

Adds a new how-to guide (docs/user-guides/oauth-proxy.md) demonstrating a multi-tenant COO deployment where each namespace's ThanosQuerier is protected by an oauth-proxy sidecar with delegated subject access reviews.

Covers three projects with per-user access control, NetworkPolicy isolation, cross-namespace Thanos federation (project-c querying project-b), and bearer token authentication for service accounts (project-a only).

Closes #1228


coffeegoesincodecomesout:
Think this is now ready for a human review, I have addressed all of the suggestions and retested.

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+  namespace: project-a
+  labels:
+    discover: project-a # 👈 matched by ThanosQuerier selector below
+spec:
+  logLevel: debug
+  retention: 1d
+  resourceSelector:
+    matchLabels:
+      k8s-app: prometheus-coo-example-monitor
+  namespaceSelector:
```
simonpasquier, line 133: (nit) in this case we could omit the namespaceSelector and the MonitoringStack would only discover resources in its namespace.

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+
+The oauth-proxy sidecar delegates authentication to the OpenShift API server via subject access reviews. Users must authenticate with their OpenShift credentials before they can reach the Thanos Query UI.
+
+Access is controlled per namespace:
+
+* `user1` can query metrics in `project-a` only.
+* `user2` can query metrics in `project-b` only.
+* `user3` can query metrics from both `project-b` and `project-c` via the `ThanosQuerier` in `project-c`.
+
+The `ThanosQuerier` in `project-c` uses a `namespaceSelector` to federate metrics from both `project-b` and `project-c`. A `NetworkPolicy` in `project-b` permits the Thanos sidecar in `project-c` to scrape it.
```
simonpasquier: "scrape" might not the right term, "communicate with" maybe? Also I assume that the communication is between the Thanos querier in project-c and the Thanos sidecars in project-b.

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+    app.kubernetes.io/managed-by: observability-operator
+    app.kubernetes.io/part-of: ThanosQuerier
+  name: thanos-querier-example-coo-thanos
+  namespace: project-a
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: thanos-querier-example-coo-thanos
+  template:
+    metadata:
```
simonpasquier: do we need to reapply the labels?

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+  ports:
+  - name: proxy
+    port: 8888
+    protocol: TCP
+    targetPort: oauth-proxy # 👈 route external traffic to the oauth-proxy container port
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
```
simonpasquier: do we need to reapply the labels?

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+    app.kubernetes.io/instance: thanos-querier-example-coo-thanos
+    app.kubernetes.io/managed-by: observability-operator
+    app.kubernetes.io/part-of: ThanosQuerier
+  name: thanos-querier-example-coo-thanos
+  namespace: project-a
+spec:
+  selector:
+    matchLabels:
```
simonpasquier: same here

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+
+#### Grant user3 access
+
+```sh
+kubectl -n project-c create rolebinding view-user3 \
+  --clusterrole=view \
+  --user=user3
+```
+
+## Validation
```
simonpasquier: (nit) maybe the validation steps should be moved to their respective namespace? it would also provide a nice ramp-up from deploying a single-namespace thanos querier to a multi-namespace setup.

simonpasquier (commented):
Very good job!

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+* `user3` can authenticate to the `project-c` route and will see metrics from both `project-b` and `project-c`.
+
+### Bearer token access (project-a only)
+
+Generate a short-lived token for the `robot-user` ServiceAccount and query the Thanos API directly:
+
+```sh
+TOKEN=$(kubectl -n project-a create token robot-user)
+ROUTE=$(kubectl get route -n project-a thanos-querier-authenticated -o jsonpath='{.spec.host}')
+curl -H "Authorization: Bearer ${TOKEN}" "https://${ROUTE}/api/v1/query?query=up"
```
simonpasquier: I needed `-k`.

docs/user-guides/oauth-proxy.md
```diff
@@ -0,0 +1,544 @@
...
+      expr: version{job="prometheus-coo-example-app"} > 0
+      labels:
+        severity: warning
+```
+
+**NOTE:** The test application is included in `00-project-a.yaml` applied in the previous step. Equivalent applications for `project-b` and `project-c` are included in `01-project-b.yaml` and `02-project-c.yaml` respectively.
+
+#### Enable bearer token authentication
+
+To allow ServiceAccounts to authenticate with bearer tokens, grant the `thanos-querier` ServiceAccount permission to create `TokenReview` resources:
```
simonpasquier: I would have thought that COO manages the servicea ccount field in the deployment but it isn't for now and I think it's a bug. I'd be better to patch the SA managed by COO than creating a new one if possible.

coffeegoesincodecomesout:
@simonpasquier thank you for the comments and suggestions, I am much happier with the flow now. 
One thing that feels slightly off is that the "Configure OAuth Proxy" section is the only section that does not have a "or run" option. The secret creation cant be done in a manifest.. it would need a .sh file.  I could add the other two commands in that section to manifests and leave just one stand alone command for the secret - wdyt?

simonpasquier:
> I could add the other two commands in that section to manifests and leave just one stand alone command for the secret - wdyt?

It works for me.

